PR #33, opened 2015-12-23
Add idea to gitignore

poulp:


zoidyzoidzoid:
This should be in your [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore), so we don't have to support ignoring all possible IDE's project files.

poulp:
ok :)

PR #130, opened 2016-01-23
New option: _always_build.

ajdavis:
Ensure some artifacts are built even if nothing depends on them. For example, on my site, there is an Atom feed consumed by Planet Python but nothing on my site links to it. Or, if you have HTML redirect pages from old locations on your site, you want those built even though nothing refers to them on your own site.

### DIFF
--- a/lektor/datamodel.py
+++ b/lektor/datamodel.py
@@ -214,9 +214,9 @@ class DataModel(object):
 
     def __init__(self, env, id, name_i18n, label_i18n=None,
                  filename=None, hidden=None, protected=None,
-                 child_config=None, attachment_config=None,
-                 pagination_config=None, fields=None,
-                 primary_field=None, parent=None):
+                 always_build=None, child_config=None, 
+                 attachment_config=None, pagination_config=None, 
+                 fields=None, primary_field=None, parent=None):
         self.env = env
         self.filename = filename
         self.id = id
@@ -228,6 +228,9 @@ class DataModel(object):
         if protected is None:
             protected = False
         self.protected = protected
+        if always_build is None:
+            always_build = False
+        self.always_build = always_build
         if child_config is None:
             child_config = ChildConfig()
         self.child_config = child_config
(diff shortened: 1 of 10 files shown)


ajdavis:
If you like this idea I'll submit documentation too.

mitsuhiko:
So generally +1 on the idea, I just also want to quickly throw this into the ring: #37

Maybe we can build it somehow that the same key can be used for this. The main idea was also that plugins can customize this.

ajdavis:
I _think_ this is totally distinct from #37. In #37 we rebuild an artifact
even if Lektor thinks it's up to date. In this PR, we do the _first_ build
of an artifact, even if Lektor thinks it shouldn't be built at all. Once an
always_build artifact is built, Lektor decides whether to rebuild using the
usual logic.

I should come up with a different name than always_build. How about is_root?

They are related and should be configured in a way that's consistent for
the user, but I think the implementation of the two features is separate.

mitsuhiko:
Ah I see. I wonder why this happens though. Should this not be found just by iterating over the children? I assume this might happen because a pagination filter now restricts what we iterate over?

ajdavis:
Yes that's right. In my case, the page at https://emptysqua.re/blog/ has pagination enabled, so the build program iterates over child pages that match the pagination filter. That excludes some things like my Atom feeds, which wouldn't be built at all unless I find some way to force it. HTML redirect pages would have the same problem.

mitsuhiko:
I think we should solve this somehow differently. Things not included by pagination should still be included for the builder automatically.

ajdavis:
Ah I see. I should mark such sources "undiscoverable". I'll leave a more general solution to you, I didn't realize I could fix this with the existing Lektor code.

mitsuhiko:
@ajdavis i think there should be a bug opened that if something is missed by the pagination that they are not picked up by the build process if that's the case.

ajdavis:
OK, I've opened #132.

PR #133, opened 2016-01-25
Build all children and attachments despite pagination.

### LINKED ISSUE #132: Children filtered by pagination query aren't built
In `PageBuildProgram.iter_child_sources`, if a page's children are paginated, Lektor adds attachments and undiscoverable children as dependencies of the page's child page number 1. This causes two bugs: discoverable children that are not included in the pagination query are not built, and attachments aren't built if there is no page 1.

First, if I have a `page.ini` like:

``` ini
[pagination]
enabled = yes
items = this.children.filter(F._model == 'foo')
```

...then children with datamodels besides "foo" are not built, unless they become dependencies some other way.

Second, if a page has zero pagination items, then there is no page 1 and attachments aren't built either.

ajdavis:
Fixes #132.

### DIFF
--- a/lektor/build_programs.py
+++ b/lektor/build_programs.py
@@ -187,17 +187,10 @@ class PageBuildProgram(BuildProgram):
             yield Page(self.source.pad, self.source._data,
                        page_num=page_num)
 
-    def _iter_undiscoverable_children(self):
-        return self.source.children \
-            .include_undiscoverable(True) \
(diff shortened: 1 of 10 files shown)


ajdavis:
New patch based on [our Gitter chat](https://gitter.im/lektor/lektor?at=56a7c7638fbaf4220af97743).

PR #158, opened 2016-02-26
Web Admin User Authentication

ryneeverett:
Minimal user model and authentication so lektor server can be deployed
to the web. New `/users` views:
- `/login`
- `/logout`
- `/set-password` A page for initializing or resetting a password.
- `/set-password-link` A view providing a /set-password link.
- `/is-admin` Is the client an admin?
- `/` An admin-only page for managing users with the following API's:
  - `/delete` Delete user.
  - `/reset` Reset password.
- `/add-user` View for adding a user
  - `/add` API for adding a user

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ python:
   - "2.7"
 
 install:
-  - pip install --editable .
+  - pip install --editable ".[webadmin]"
 
 script: make test
 
--- a/lektor/admin/models.py
+++ b/lektor/admin/models.py
@@ -1,0 +1,43 @@
+from flask import current_app
+from flask.ext.login import UserMixin, make_secure_token
+from flask_sqlalchemy import SQLAlchemy
+from werkzeug.security import generate_password_hash, check_password_hash
+
+db = SQLAlchemy(current_app)
+
+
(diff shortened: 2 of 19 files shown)


ryneeverett:
I think I'm going to end up going with a proxy authentication setup for my own purposes, but I'm still interested in seeing some progress -- if only conceptualization -- within Lektor. I'd probably rather see "hosted Lektor" take a similar proxy architecture than the "throw it all in the Flask app" approach I took here, but there are a lot of different packaging approaches one could take to that.

I'd have no objection to this being closed but it would be nice if it were closed with some comment on what a better approach would look like.

singingwolfboy:
Hey @ryneeverett, thanks for the pull request! I'm going through some old pull requests to try to clean them out and get this project properly maintained again. Sorry for the long delay in getting to this!

This is a cool idea, but I don't think it fits with the core idea behind Lektor. Lektor is made to generate a static website, not to run `lektor server` as a dynamic website. (You can write a Flask for that!) I agree that "hosted Lektor" is a really cool idea, and something I'd like to see, but I think that belongs as a separate application built on top of Lektor, rather than built-in to this codebase. Does that make sense?

PR #212, opened 2016-05-10
lektor server --no-prune option

ajdavis:
I build to a directory that also has non-Lektor files in it, so I always build with --no-prune. In order to preview a page without deleting all my non-Lektor files from the build directory, I need a `lektor server --no-prune` option.


ajdavis:
Would you like to merge this? I really rely on it. =)

ajdavis:
Thanks!

PR #220, opened 2016-06-03
Catalan translation

XaviTorello:
Created **Catalan translation**

ISO 639-1 Code 'ca'

lektor/translations/ca.json


singingwolfboy:
I don't speak Catalan, but I'm going to assume that this translation is correct, and merge it. Thank you! ✨

PR #231, opened 2016-06-14
Serve external fonts locally

### LINKED ISSUE #230: Serve external fonts locally
Currently, Lektor's admin interface makes a call to Google's font API for styling. For some end users this is a concern, as it requires an internet connection for proper styling and puts them at risk of being tracked. A simple fix would be to just serve the fonts locally from the server instead of pulling them in from Google.

Would it be okay to just implement this and submit a PR along these lines? I am just wondering whether this was a conscious decision to reduce page load or merely a more convenient solution to get the font in.

cc @mrcosta

milibopp:
Resolves #230.


lektor/admin/static/fonts/LICENSE
```diff
@@ -0,0 +1,203 @@
...
+      License. However, in accepting such obligations, You may act only
+      on Your own behalf and on Your sole responsibility, not on behalf
+      of any other Contributor, and only if You agree to indemnify,
+      defend, and hold each Contributor harmless for any liability
+      incurred by, or claims asserted against, such Contributor by reason
+      of your accepting any such warranty or additional liability.
+
+   END OF TERMS AND CONDITIONS
+
+   APPENDIX: How to apply the Apache License to your work.
```
singingwolfboy: Seems like all we need is this appendix section. We can replace `[yyyy]` with `2013` and `[name of copyright owner]` with `Christian Robertson`, since the person who designed the font is presumably the copyright owner.

milibopp: Thanks for the hint. Guess I just blindly copied the file ^^

milibopp:
@singingwolfboy addressed the licensing issue. should be ready to merge now.

singingwolfboy:
This definitely seems like a good change to make. Thank you!

PR #232, opened 2016-06-14
Ignore dev-example suggested in README

milibopp:
As the dev-example folder is suggested in the README for local testing it is nice to ignore this for convenience and to prevent developers from accidentally committing their example project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ build
 dist
 venv
 lektor/admin/static/gen
+/dev-example
 
 node_modules
 gui/build


singingwolfboy:
This is not a bad idea, but I'm not sure that `dev-example` is the best name to use. It's not very clear. Perhaps `example-project` would be better? Do you want to update the README and the `.gitignore` file?

@mitsuhiko, any thoughts on this?

milibopp:
Of course I am happy to update the README, once the bikeshedding is done.

singingwolfboy:
Fair point. I've merged #303, which resolves this pull request. Thanks for the idea, @aepsil0n!

PR #235, opened 2016-06-25
Add italian translation

emilyalbini:
Thank you for this great piece of software!


singingwolfboy:
I don't speak Italian, but I'm going to assume that this translation is correct, and merge it. Thank you! ✨

PR #246, opened 2016-07-27
Added pagination properties for first and last

chipperdoodles:
I added pagination properties for first page and last page, these can useful for First and Last links for pagination such as

`<a href="{{ pagination.last_page|url }}"> Last page </a>`

This is my very first pull request and I'm a beginner! but I believe these could be useful shortcuts for templates. I would like to attempt making the same effect for siblings and children pages but that takes a bit more poking around and understanding.


chipperdoodles:
Ah sorry, you know how i said i was a begginer? I accidently deleted the branch my pull request was from

PR #247, opened 2016-07-27
added first_page and last_page pagination

chipperdoodles:
Accidently deleted my original branch (relearning git)

Reopening PR for first_page and last_page pagination properties.

first_num and last_num pull an int value of first page and last page respectively
first_page and last_page pull the record of those pages resepectivly and can be used in template such as

 `<a href="{{ this.pagination.last_page|url }}"> Last Page </a>`

Now i'm a beginner and it could use some refined checks for edge cases but i know it at least provides a functional base..

I would also like propose adding these to the siblings option in a form if we can.

### DIFF
--- a/lektor/pagination.py
+++ b/lektor/pagination.py
@@ -67,6 +67,34 @@ class Pagination(object):
         return self.config.get_record_for_page(self.current,
                                                self.page + 1)
 
+    @property
+    def first_num(self):
+        """Number of the first page"""
+        num = list(range_type(1, self.pages + 1))
+        return min(num)
+
+    @property
+    def last_num(self):
+        """Number of the last page"""
+        num = list(range_type(1, self.pages + 1))
+        return max(num)
+
+    @property
+    def first_page(self):
+        """Returns the record for first page of pagination"""
+        if self.pages == 0:
+            return None
+        else:
+            return self.config.get_record_for_page(self.current, self.first_num)
+
+    @property
+    def last_page(self):
+        """Reutrns the record for last page of pagination"""
+        if self.pages == 0:
+            return None
+        else:
+            return self.config.get_record_for_page(self.current, self.last_num)
+
     def for_page(self, page):
         """Returns the pagination for a specific page."""
         if 1 <= page <= self.pages:


singingwolfboy:
Hey @chipperdoodles, thanks for the pull request! However, I'm not sure if these properties are useful. `first_num` will always be `1` and `last_num` will always be `self.pages`, won't it? Similarly, you can just do `self.for_page(1)` and `self.for_page(self.pages)` to get `first_page` and `last_page`, right?

Maybe there's some use for these properties that I'm not seeing. Can you give me an example of where you might use them? (Automated tests are great for doing that!)

And also, congratulations on making your first pull request! That's really exciting! ✨

chipperdoodles:
@singingwolfboy Ah, I see what you mean, since pagination is int based those are definitely easier ways of grabbing the values and we wouldn't need those two properties. I think I did it by min() and max() originally thinking it would be useful if the values were ever less than 1 or maybe a float of some kind, But in the case of pagination I suppose they never will be.

also thank you very much, I'm very excited myself!

and just to check, first_page and last_page are still welcome properties, the problem is just first_num and last_num being redundant ?

chipperdoodles:
Oh sorry i just re-read, the last_page and first_page properties are mainly meant to make writing a template more smooth aka being able to call links in a template in a clean manner like this:
`<a href="{{ this.pagination.last_page|url }}"> Last Page </a>`

singingwolfboy:
Hey @chipperdoodles, sorry for the long delay in responding -- this project hasn't been maintained very well, but I'm trying to change that now. I think we can get this merged, but before that happens, the code needs to be cleaned up a bit, and we need automated tests to cover this feature. If you don't know how to do that, maybe we could work together on that at some point?

Alternatively, if you don't care about this feature anymore, you can close the pull request. (It's been so long that I certainly wouldn't blame you if that were the case!)

chipperdoodles:
@singingwolfboy No worries! I hope it does pick up, I like this project and wish i could actually do more to help. I am very green so i don't know how to make automated tests yet. I can clean up the code so the only two properties are first page and last page for the sake of templating. I do care about this but only want it accepted if it seems like a relevant or reasonable addition!

singingwolfboy:
@chipperdoodles I'm going to close this pull request for now, since it doesn't seem like we can get it to a mergeable state any time soon. I'd be happy to teach you more about automated tests if you need, or here's [pretty good introduction to automated testing](https://eev.ee/blog/2016/08/22/testing-for-people-who-hate-testing/) that might be useful.

chipperdoodles:
Sounds good, i'll give it a read and a shot and reopen a better one when i do! Thanks for your feedback :D

PR #252, opened 2016-08-04
Fix updating attachment_types config from ini file

philippbosch:
I saw there is some code that checks the project config for additional attachment types and adds them to the ones from the default config. So I added this to my `.lektorproject` file:

``` ini
[attachment_types]
.mp4 = video
```

Unfortunately it didn't work but gave me this error:

```
…
File "/usr/local/lib/lektor/lib/python2.7/site-packages/lektor/environment.py", line 98, in <genexpr>
    for k, v in inifile.section_as_dict('attachment_types'))
ValueError: too many values to unpack
```

This commit fixes this by using `iteritems()` for the section dict.

### DIFF
--- a/lektor/environment.py
+++ b/lektor/environment.py
@@ -93,12 +93,9 @@ def update_config_from_ini(config, inifile):
     set_simple(target='LESSC_EXECUTABLE',
                source_path='env.lessc_executable')
 
-    config['ATTACHMENT_TYPES'].update(
-        (k.encode('ascii', 'replace'), v.encode('ascii', 'replace'))
-        for k, v in iteritems(inifile.section_as_dict('attachment_types')))
-
-    config['PROJECT'].update(inifile.section_as_dict('project'))
-    config['PACKAGES'].update(inifile.section_as_dict('packages'))
+    for section_name in ('ATTACHMENT_TYPES', 'PROJECT', 'PACKAGES'):
+        section_config = inifile.section_as_dict(section_name.lower())
+        config[section_name].update(section_config)
 
     for sect in inifile.sections():
         if sect.startswith('servers.'):
--- a/lektor/environment.py
+++ b/lektor/environment.py
@@ -95,7 +95,7 @@ def update_config_from_ini(config, inifile):
 
     config['ATTACHMENT_TYPES'].update(
         (k.encode('ascii', 'replace'), v.encode('ascii', 'replace'))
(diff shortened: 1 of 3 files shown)


singingwolfboy:
@philippbosch: This is great, thank you! Could you add an automated test to go with this bugfix, so that the bug stays fixed? If you don't know how to do that, I might be able to help, just let me know.

philippbosch:
@singingwolfboy: I just added a test for this. Two questions:
1. I didn't find an existing `test_*.py` file that would be a good fit for this test so I created a new one (`test_config.py`). Please let me know if you think it should go somewhere else.
2. The test currently fails on Python 3 because apparently [the key and value I put in the test config](https://github.com/philippbosch/lektor/blob/7a3f33ac057bcfac3c6f7212f3466b26f13ff75a/tests/demo-project/Website.lektorproject#L24) end up as _bytes_ instead of _str_ in the config values dict. I think that has to do with the way these are treated after reading from the .ini file (see [here](https://github.com/lektor/lektor/blob/master/lektor/environment.py#L98)). `'foo'.encode('ascii', 'replace')` returns a _str_ (`'foo'`) in Python 2 but a _bytes_ object (`b'foo'`) in Python 3. I'm stuck here – any advice would be appreciated …

singingwolfboy:
@philippbosch: this is perfect! I think creating a new `test_config.py` file is the best option for where to put the test, and the test itself is very simple and clear.

As for the test failing on Python 3, I have no idea why those `.encode()` calls are in there at all -- I suspect they _shouldn't_ be there. I'd suggest taking them out, which would result in three lines that look almost identical. That can be refactored into something that looks like this:

``` python
for section_name in ('ATTACHMENT_TYPES', 'PROJECT', 'PACKAGES'):
    section_config = inifile.section_as_dict(section_name.lower())
    config[section_name].update(section_config)
```

Do you want to add a commit to this pull request that makes that change? @mitsuhiko, can you take a look at this and let me know if there's something I'm missing? If there's a reason why those `.encode()` calls _do_ need to be there, then this gets more complicated.

philippbosch:
@singingwolfboy Thanks for your detailed feedback. I just pushed two more commits including the suggested refactoring. Tests are passing now for Python 2 and 3. Waiting for @mitsuhiko's feedback now.

singingwolfboy:
Doesn't look like @mitsuhiko is going to get around to reviewing this, so I'm just going to merge it. Thanks for the help, @philippbosch!

PR #253, opened 2016-08-04
Bumped dependencies

go-bears:


gui/package.json
```diff
@@ -12,20 +12,25 @@
...
+    "babel": "^6.1.0",
+    "babel-core": "^6.1.0",
+    "babel-loader": "^6.1.0",
+    "babel-plugin-transform-decorators-legacy": "^1.3.4",
+    "babel-plugin-transform-object-rest-spread": "^6.1.0",
+    "babel-plugin-transform-runtime": "^6.1.0",
+    "babel-preset-es2015": "^6.1.0",
+    "babel-preset-react": "^6.1.0",
+    "babel-preset-stage-0": "^6.1.0",
+    "babel-register": "^6.1.0",
```
singingwolfboy, line 24: Seems like you've added a few dependencies that weren't there before. Can you explain why?

go-bears: I believe these were dependencies needed for setting up Jasmine testing for Travis CI. would that be the right summary @arbonap @minifast?

singingwolfboy: I guess my question is, were these dependencies needed before, and weren't properly marked in the `package.json` file? Or did upgrading the existing dependencies require also pulling in these new dependencies?

go-bears: @arbonap @minifast these dependencies weren't needed before Jasmine unit testing was introduced, yes?  To the best of my knowledge, these dependencies were added to "appease Travis" and to get Jasmine working properly.

singingwolfboy:
I'm going to say that this pull request is ready to be merged. Thanks for doing this!

PR #254, opened 2016-08-05
Transform admin jsx from Es5 to es6

go-bears:


item4:
@mitsuhiko I think you might choice convention of JS

singingwolfboy:
Thanks for the pull request, @go-bears! I'm sorry it took so long to get it reviewed, but it looks great to me. ✨ 🚀 ✨

PR #260, opened 2016-08-21
Shift cache to local appdata on Windows

chris-morgan:
Lektor has been storing its cache in the roaming app data directory (`%APPDATA%`, typically `C:\Users\username\AppData\Roaming` but quite possibly on a remote machine, either genuinely living on another machine or uploaded when you log out and synchronised again when you log in).

For Vista and later, this is not appropriate; the local app data directory (`%LOCALAPPDATA%`, typically `C:\Users\username\AppData\Local` and defined as not being on a remote machine) is where it should be instead, because it is machine-specific (for example, two machines may have different versions of Python, or different architectures).

APPDATA remains as a fallback before ~, because XP won’t have LOCALAPPDATA set.

### DIFF
--- a/lektor/utils.py
+++ b/lektor/utils.py
@@ -625,9 +625,11 @@ def get_app_dir():
 
 def get_cache_dir():
     if is_windows:
-        folder = os.environ.get('APPDATA')
+        folder = os.environ.get('LOCALAPPDATA')
         if folder is None:
-            folder = os.path.expanduser('~')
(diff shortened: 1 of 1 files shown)


singingwolfboy:
Hey @chris-morgan, sorry for the long delay in getting to your pull request. This looks totally reasonable to me, so I'm going to merge it. Thanks so much for your help! ✨ 🚀 ✨

PR #262, opened 2016-08-22
Use dest_path if file is moved. Fixes #261.

### LINKED ISSUE #261: Lektor will not kick off build process on Linux (Debian) when a file is changed.
When running Lektor on Debian, the build process will not start after a content file is modified. This behavior is not seen when running on OSX.

**OSX Behavior**
1. Edit a file in the Admin UI
2. The `Watcher` class instance will see the following events:
   1. `FileDeletedEvent` for the stale version of the `contents.lr` file.
   2. `FileCreatedEvent` for the new version of the `contents.lr` file.

When creating the file descriptor for the `.__atomic-write` temporary file, a `FileCreatedEvent` is never seen.

**Linux Behavior**
1. Edit a file in the Admin UI
2. The `Watcher` class instance will see the following events:
   1. `FileCreatedEvent` for the `.__atomic-write` file.
   2. `FileModifiedEvent` for the `.__atomic-write` file.
   3. `FileMovedEvent` for the `.__atomic-write` file.
      - The `src_path` for the event points to the `.__atomic-write` file.
      - The 'dest_path`for the event points to the`contents.lr` file.

---

Only the `src_path` from Watchdog is used when generating the `item` value to then be passed the the event queue / callback.

If the event is an instance of `FileMovedEvent`, then the `dest_path` should be used when generating items, so valid filesystem events aren't ignored by the `is_interesting` predicates that filter out superfluous events.

coreygrunewald:
Use the `dest_path` value from `FileMovedEvent` class instances so changes will not be ignored.

Fixes #261.

### DIFF
--- a/lektor/watcher.py
+++ b/lektor/watcher.py
@@ -2,7 +2,7 @@ import os
 import time
 
 from watchdog.observers import Observer
-from watchdog.events import FileSystemEventHandler, DirModifiedEvent
+from watchdog.events import FileSystemEventHandler, DirModifiedEvent, FileMovedEvent
 
 from lektor._compat import queue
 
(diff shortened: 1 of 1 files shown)


singingwolfboy:
This seems fine to me. Thanks for the bump, @coreygrunewald!

PR #271, opened 2016-08-29
refine integration tests and setup

go-bears:
*implemented pytest's mock
*configured travis build to pass python 3.5 quirks
*flagged test_publication with `skipif` so rest of tests can pass


.travis.yml
```diff
@@ -7,15 +7,33 @@ cache:
...
+  - sed -i 's/import cmdline/from . import cmdline/' /home/travis/virtualenv/python3.5.2/lib/python3.5/site-packages/pytest_shutil/workspace.py || true
+  - sed -i 's/RqlRuntimeError, err/RqlRuntimeError as err/' /home/travis/virtualenv/python3.5.2/lib/python3.5/site-packages/pytest_server_fixtures/rethink.py || true
+before_script:
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
+
+
+before_script:
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
```
singingwolfboy: We should only have one `before_script` section, not two

.travis.yml
```diff
@@ -7,15 +7,33 @@ cache:
...
+  firefox: "47.0.1"
+
 before_install:
   - curl -s -o $HOME/.nvm/nvm.sh https://raw.githubusercontent.com/creationix/nvm/v0.31.4/nvm.sh
   - source $HOME/.nvm/nvm.sh
   - nvm install stable
   - node --version
+  - pip install -U pip 
+  - pip install pytest==2.9.2
+  - pip install pytest-mock
```
singingwolfboy: If we need `pytest-mock`, then it should be in the `tests_require` list in the `setup.py` file.

singingwolfboy:
Can you delete the `jasmine` and `lektor@0.0.0` files from this pull request? Also squash commits together, so there aren't so many fixup commits in the history? If you're not sure how to squash commits, [here's a guide on how to rebase your pull request](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request), and it has a section in there about squashing commits together.

tests/conftest.py
```diff
@@ -197,3 +197,8 @@ def cleanup():
     email = "lektortest@example.com"
     subprocess.check_call(['git', 'config', 'user.email', email])
     return email
+
+
+@pytest.fixture(scope='session')
+def splinter_screenshot_dir(request):
+    return os.path.join(os.path.abspath(request.config.option.splinter_screenshot_dir), 'tmp')
```
singingwolfboy: Can you explain what this splinter screenshot directory is for? Are you expecting that the developer running the tests will look in this directory and view the screenshots after the tests run? If that's the case, you should use a more descriptive name than "tmp" for this directory, and you should add some documentation to the `README` file about viewing screenshots after running tests.

If you don't expect people to view screenshots after every run, maybe this fixture isn't necessary? Or maybe you can use the [`tmpdir` fixture](http://docs.pytest.org/en/latest/tmpdir.html) to easily get a temporary directory? That way, you won't have to add this directory to the `.gitignore` file.

tests/test_integration.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+import mock
+import sys
+
+from lektor.admin.webui import WebUI
+# from lektor.publisher import RsyncPublisher
```
singingwolfboy: Please don't commit commented-out code to the repository.

tests/test_integration.py
```diff
@@ -0,0 +1,31 @@
...
+
+@pytest.yield_fixture(scope='function')
+def app(mocker, scratch_env, simple_http_server):
+    mocker.patch("lektor.publisher.RsyncPublisher.publish")
+    yield WebUI(scratch_env, output_path=simple_http_server.document_root)
+
+
+@pytest.mark.skipif(sys.version_info > (3,0),
+                    reason="python 3 currently not compatible pytest plugin")
```
singingwolfboy, line 14: I think you're missing a word in this reason.

tests/test_preview.py
```diff
@@ -0,0 +1,22 @@
+import pytest
+from lektor.admin.webui import WebUI
+
+@pytest.yield_fixture(scope='function')
+def app(request, env, pad, simple_http_server):
+    yield WebUI(env, output_path=simple_http_server.document_root)
+
+def test_preview(live_server, browser):
+    browser.visit('%s/dir_with_index_html' % live_server.url())
```
singingwolfboy: Rather than calling `live_server.url()` all the time, can you call it once, store the result into a variable, and use that variable throughout the rest of the test?

I also find it more readable to use [format strings](https://docs.python.org/3.5/library/string.html#formatspec) rather than the sprintf format. For example, you could do something like:

``` python
server_url = live_server.url()
browser.visit("{url}/dir_with_index_html".format(url=server_url))
assert browser.url == "{url}/dir_with_index_html/".format(url=server_url)
```

go-bears:
@singingwolfboy  thanks for the feedback! @arbonap and I  be working on some of the fixes :) 
The tmp directory would store screenshots if a browser test fails, and splinter takes a snapshot of the browser state upon the error. Since we would hope that all the tests would pass, a developer wouldn't have to check the tmp directory all the time, so the tmpdir fixture seems appropriate. We'll see if we can get that working with the switch too.

go-bears:
@singingwolfboy  @arbonap and I (with the help of a mentor) think we squashed the commit logs? We had a lot of merge conflicts to work through. (@luckyrome I'm not exactly sure if we did it right in `git reflog` i see the squashes, but in `git --oneline` I still all the little commits we made during debugging. where i thought these would've been merged into a single commit.

go-bears:
@arbonap @luckyrome think the build is fixed again! @seemaullal helped us with the commit squashing, but we may have accidentally skipped a critical commit for the .travis.yml while resolving merge conflicts .  But rebuilt the travis config from the previous working travis builds. @singingwolfboy a githubber showed us that the merge button has an option to squash commits and then merge in a drop down menu. but we have our commit log down way from 44 commits :)

luckyrome:
@go-bears for future reference, git reflog always shows you your git history (like switching branches etc etc), regardless of what you've committed (that way you can revert to a pre-crisis state if you accidentally delete something). 

What you want to actually check is `git log` which will show you actual commits. When you squash, whatever diff in your branch exists will be applied to the current branch, so unless you were missing a commit in your branch, you shouldn't be missing anything in the squash unless the auto-merge didn't happen correctly.

go-bears:
Thanks @luckyrome for the advice!

singingwolfboy:
@go-bears @arbonap Thanks for all the work you've done here! Are you still looking to get this merged? If so, you'll need to rebase on top of `master` and resolve conflicts.

Also, is pytest-shutil compatible with Python 3 yet? I remember seeing some activity there, but I don't know if they have a release out on PyPI yet that actually works properly...

go-bears:
@singingwolfboy  that would be great to get these merged! I think Pytest shutils has been updated in Pypi according the changelog https://pypi.python.org/pypi/pytest-shutil

The last time we tried to merge and rebase these files was a pretty serious disaster for us. Is rebasing on top of master different than trying to merge and rebase on our branch?

tariquesani:
Any progress on this?

tariquesani (changes requested):
Resolve conflicts with the current master

yagebu:
I'm closing this stale PR, feel free to reopen if you want to continue this.

PR #274, opened 2016-09-04
Make Created At correspond to actual creation date and add Modified At.

alysbrooks:
I noticed that the created_at image item didn't seem to correspond to the actual date the photograph was taken, so I modified the property to use the correct EXIF field. I also added a modified_at item.

### DIFF
--- a/lektor/imagetools.py
+++ b/lektor/imagetools.py
@@ -165,6 +165,15 @@ class EXIFInfo(object):
     @property
     def created_at(self):
         try:
+            print(self._mapping)
+            return datetime.strptime(self._mapping['EXIF DateTimeOriginal'].printable,
+                                     '%Y:%m:%d %H:%M:%S')
+        except (KeyError, ValueError):
+            return None
+
+    @property
+    def modified_at(self):
+        try:
             return datetime.strptime(self._mapping['Image DateTime'].printable,
                                      '%Y:%m:%d %H:%M:%S')
         except (KeyError, ValueError):
--- a/tests/test_images.py
+++ b/tests/test_images.py
@@ -22,6 +22,7 @@ def test_exif(pad):
     assert image.exif.camera_model == 'iPhone 6'
     assert image.exif.copyright is None
     assert image.exif.created_at == datetime(2015, 12, 6, 11, 37, 34)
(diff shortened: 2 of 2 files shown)


lektor/imagetools.py
```diff
@@ -165,6 +165,15 @@ def iso(self):
     @property
     def created_at(self):
         try:
+            print(self._mapping)
```
singingwolfboy, line 168: This pull request can't be merged with debugging statements like this still in the code.

tests/test_images.py
```diff
@@ -22,6 +22,7 @@ def test_exif(pad):
     assert image.exif.camera_model == 'iPhone 6'
     assert image.exif.copyright is None
     assert image.exif.created_at == datetime(2015, 12, 6, 11, 37, 34)
+    assert image.exif.modified_at == datetime(2015, 12, 6, 11, 37, 34)
```
singingwolfboy, line 25: Thanks for the test! 💖

singingwolfboy (changes requested):
Can you point me to some documentation about EXIF information? I don't know how to determine if your pull request is correct or not.

alysbrooks:
According to its readme, EXIFRead uses the official tag names. Referring to the [2.3 EXIF spec](http://www.cipa.jp/std/documents/e/DC-008-2012_E.pdf) entries for the tags, 
- DateTimeOriginal is "The date and time when the original image data was generated. " (pg. 48 or 53 in the file)
- DateTime is "The date and time of image creation. In this standard it is the date and time the file was changed. " (pg. 38 or 44 in the file) That's more ambiguous than I realized. It looks like the original TIFF format used the same [tag](http://www.awaresystems.be/imaging/tiff/tifftags/datetime.html) for the image creation date, but EXIF now uses it for the modification date.

Lektor doesn't have special handling for TIFF, so I think we're okay to use DateTime as date modified. I could look into whether it makes sense to do something more complicated, e.g., using DateTime as date created if DateTimeOriginal is missing. It looks like that's what #276 does.

tariquesani:
Created at has been fixed via https://github.com/lektor/lektor/pull/276 closing this for now. Can be reopened if conflicts  are resolved and changes requested done.

PR #276, opened 2016-09-07
Retrieve creation date from EXIF data more correctly

skorokithakis:
Currently, Lektor just retrieves the `Image DateTime` for `created_at`. That's incorrect, since, in my tests, that's usually when the image was last edited. This change tries four tags consecutively, and returns the one that's available first. The order it tries the tags more accurately reflects the creation date of the image.


skorokithakis:
Aaand looks like this is a duplicate of #274. A combination of the two would possibly be best, since `modified_at` is also desirable.

singingwolfboy:
Can you tell me how you determined those four tags? I don't know much about EXIF metadata -- is there a standard somewhere that defines what tag means what?

skorokithakis:
I downloaded `exiftool` and found the descriptions online (here's a [good example](http://www.awaresystems.be/imaging/tiff/tifftags/privateifd/exif/datetimeoriginal.html)). I also compared the tags my DSLR wrote and the ones Lightroom updated, to come up with the most representative order for the tags.

tariquesani:
If this PR gets an automated test it would be good to merge this. @skorokithakis would you be interested in writing one? I know it has been a while now...

skorokithakis:
Sure, I'll give it a shot tomorrow.

skorokithakis:
I'm not sure quite how I can test this, but I have modified the test image to have different EXIF dates per field and ensured that the right field is the one that gets picked up first.

skorokithakis:
Yeah, screw everything about git. I'm not interested in spending the twenty hours this merge would require to clean up, I'm afraid.

tariquesani:
@singingwolfboy This one seems good to be merged. Should I?

tests/test_images.py
```diff
@@ -72,4 +72,4 @@ def test_thumbnail_quality(builder):
     # See if the image file with said quality suffix exists
     assert os.path.isfile(image_file)
     # And the filesize is less than 9000 bytes
```
singingwolfboy, line 74: Can you make the value in the comment match the value in the code?

singingwolfboy (approved):
Just one little nitpick. Otherwise, I think this is ready to merge!

tests/test_images.py
tariquesani: I will merge this and then patch the comment separately.

Issue #174, opened 2016-03-17
Ordering of subpages in admin


mdashx:
Subpages should be ordered according to the model definition, but they appear to be always ordered alphabetically.

PR #280, opened 2016-09-20
Return record children sorted by configured order

ulope:
This updates the `/recordinfo` admin API endpoint to return the list of child pages sorted by the configured order (#174).

It seems to me that there should be a more efficient way than calling `pad.get()` on every child in the sorting lambda but I haven't found it. So if there is a way to optimise this I'd be happy for any pointers.

### DIFF
--- a/lektor/admin/modules/api.py
+++ b/lektor/admin/modules/api.py
@@ -46,8 +46,9 @@ def get_path_info():
 
 @bp.route('/recordinfo')
 def get_record_info():
-    db = g.admin_context.pad.db
-    tree_item = g.admin_context.tree.get(request.args['path'])
+    pad = g.admin_context.pad
+    request_path = request.args['path']
+    tree_item = g.admin_context.tree.get(request_path)
     children = []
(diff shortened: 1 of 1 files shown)


singingwolfboy:
Hey @ulope, thanks for the pull request! You're right, there's probably a more efficient way of doing this, but I don't know how to do it either. Maybe we can optimize it after the fact -- if it works, that's a good start. :smile:

However, this _does_ need an automated test to make sure that the code you've written does in fact work properly. It would be really annoying to merge unoptimized code only to find out later on that it got broken by another pull request, and nobody noticed!

ulope:
Sorry for the late reply. 
It looks like there are no API tests as of yet. Should we add them to `test_webui.py` or rather add a separate `test_api.py` (for example)?

tariquesani:
@ulope If you are still interested adding  `test_api.py`  would be the way to go...

tariquesani (changes requested):
Needs integration tests

PR #284, opened 2016-09-22
Add CLI commit message option

go-bears:
Add `--message` flag to `lektor deploy` CLI menu.
Resolves Issue #197.

### DIFF
--- a/lektor/cli.py
+++ b/lektor/cli.py
@@ -251,8 +251,10 @@ def clean_cmd(ctx, output_path, verbosity):
 @click.option('--key', envvar='LEKTOR_DEPLOY_KEY',
               help='The contents of a key file directly a string that should '
               'be used for authentication of the deployment.')
+@click.option('--message', default=None,
+               help='Add custom commit message upon deployment')
 @pass_context
-def deploy_cmd(ctx, server, output_path, **credentials):
+def deploy_cmd(ctx, server, output_path, message, **credentials):
     """This command deploys the entire contents of the build folder
     (`--output-path`) onto a configured remote server.  The name of the
     server must fit the name from a target in the project configuration.
@@ -289,7 +291,8 @@ def deploy_cmd(ctx, server, output_path, **credentials):
 
     try:
         event_iter = publish(env, server_info.target, output_path,
-                             credentials=credentials, server_info=server_info)
+                             credentials=credentials, server_info=server_info, 
+                             message=message)
     except PublishError as e:
         raise click.UsageError('Server "%s" is not configured for a valid '
                                'publishing method: %s' % (server, e))
--- a/lektor/cli.py
+++ b/lektor/cli.py
@@ -257,7 +257,7 @@ def clean_cmd(ctx, output_path, verbosity):
 def deploy_cmd(ctx, server, output_path, message, **credentials):
     """This command deploys the entire contents of the build folder
     (`--output-path`) onto a configured remote server.  The name of the
-    server must fit the name from a target in the project configuration.
+    server must fit the n ame from a target in the project configuration.
     If no server is supplied then the default server from the config is
     used.
 
@@ -270,6 +270,7 @@ def deploy_cmd(ctx, server, output_path, message, **credentials):
     For more information see the deployment chapter in the documentation.
     """
     from lektor.publisher import publish, PublishError
+
 
     if output_path is None:
         output_path = ctx.get_default_output_path()
@@ -287,15 +288,15 @@ def deploy_cmd(ctx, server, output_path, message, **credentials):
         server_info = config.get_server(server)
         if server_info is None:
             raise click.BadParameter('Server "%s" does not exist.' % server,
-                                     param_hint='server')
-
+                                     param_hint='server')    
     try:
         event_iter = publish(env, server_info.target, output_path,
                              credentials=credentials, server_info=server_info, 
                              message=message)
     except PublishError as e:
         raise click.UsageError('Server "%s" is not configured for a valid '
-                               'publishing method: %s' % (server, e))
+                                'publishing method: %s' % (server, e))
+
 
     click.echo('Deploying to %s' % server_info.name)
     click.echo('  Build cache: %s' % output_path)
--- a/lektor/cli.py
+++ b/lektor/cli.py
@@ -257,7 +257,7 @@ def clean_cmd(ctx, output_path, verbosity):
 def deploy_cmd(ctx, server, output_path, message, **credentials):
     """This command deploys the entire contents of the build folder
     (`--output-path`) onto a configured remote server.  The name of the
-    server must fit the n ame from a target in the project configuration.
+    server must fit the name from a target in the project configuration.
     If no server is supplied then the default server from the config is
     used.
 
@@ -288,7 +288,7 @@ def deploy_cmd(ctx, server, output_path, message, **credentials):
         server_info = config.get_server(server)
         if server_info is None:
             raise click.BadParameter('Server "%s" does not exist.' % server,
-                                     param_hint='server')    
+                                     param_hint='server')  
     try:
         event_iter = publish(env, server_info.target, output_path,
                              credentials=credentials, server_info=server_info, 
--- a/lektor/publisher.py
+++ b/lektor/publisher.py
@@ -219,6 +219,9 @@ class RsyncPublisher(Publisher):
         if username:
             target.append(username + '@')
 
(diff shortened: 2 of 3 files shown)


lektor/publisher.py
```diff
@@ -636,9 +638,15 @@ def git(args, **kwargs):
 
             self.link_artifacts(path)
             self.write_cname(path, target_url)
+
+            default_message = 'Synchronized build'
+            message = extra.get('message')
```
singingwolfboy: You can replace these four lines with:

``` python
message = extra.get('message', 'Synchronized build')
```

singingwolfboy:
Is it possible to write an automated test for this?

go-bears:
We can try. @arbonap and I tried and requested how best [to try and get feedback](https://github.com/lektor/lektor/issues/197) .  the message gets added with the `deploy` command, so we couldn't quite figure out how to mock that up for a test without actually deploying a lektor build. We tried following the patterns that you added to `test_deploy.py`, but we can try again. Would do-ing a draft of a test, and making a PR on the draft the best way to go about this?

lektor/publisher.py
go-bears: great! thanks for feedback.

lektor/publisher.py
```diff
@@ -636,9 +641,15 @@ def git(args, **kwargs):
...
+            if message is None:
+                message = default_message
+
```
go-bears, line 644: I'm reviewing our notes on this, and we found get() method didn't work if there is already a key, here `message` and it was set to `None`.  @arbonap, @gecrooks, and I had tried doing this with the single line method that you also suggested, but found this implementation to be more sound in cases were a message wasn't being passed in.

singingwolfboy: Oh, to cover cases like that, you can do this instead:

``` python
message = extra.get('message') or 'Synchronized build'
```

vassudanagunta:
I disagree that this should close #197 if it is a CLI-only solution. Lektor is all about making it easy for non-technos to publish via the GUI, so a publish comment should be enterable via the GUI.

go-bears:
@arbonap and I have been working on Lektor as a student project with @mitsuhiko and @singingwolfboy . It was suggested to us that Lektor's UI would likely be undergoing significant refactoring.  would it be a good idea to add the GUI-side commit message as a milestone for that project since the flow of the UI will be rethought?

go-bears:
@singingwolfboy @arbonap and I wrote a unit test for the cli git commit feature here #296.  It passes travis and seems to test the functionality of passing in a commit message into publisher.publish() . Thanks for your guidance whilst we wrote the integration and jasmine tests!

vassudanagunta:
@go-bears I don't know. Lektor is all about the UI. Seems odd to "fix" this in the CLI but not the UI. I imagine (I emphasize "imagine") that it is a relatively simple thing to add a comment field to the publish UI, and that such an addition is so simple it won't matter if it were done before or after a major refactor. If I were already familiar with the Lektor codebase and not a Python beginner, I'd just fix it already rather than comment here :)  But perhaps far easier to say than do :) JMHO

lektor/cli.py
```diff
@@ -251,11 +251,13 @@ def clean_cmd(ctx, output_path, verbosity):
...
               'be used for authentication of the deployment.')
+@click.option('--message', default=None,
+               help='Add custom commit message upon deployment')
 @pass_context
-def deploy_cmd(ctx, server, output_path, **credentials):
+def deploy_cmd(ctx, server, output_path, message, **credentials):
     """This command deploys the entire contents of the build folder
     (`--output-path`) onto a configured remote server.  The name of the
-    server must fit the name from a target in the project configuration.
+    server must fit the n ame from a target in the project configuration.
```
singingwolfboy: Oops, I don't think you meant to add this space here :)

lektor/cli.py
```diff
@@ -251,11 +251,13 @@ def clean_cmd(ctx, output_path, verbosity):
 @click.option('--key', envvar='LEKTOR_DEPLOY_KEY',
               help='The contents of a key file directly a string that should '
               'be used for authentication of the deployment.')
+@click.option('--message', default=None,
+               help='Add custom commit message upon deployment')
```
singingwolfboy, line 255: This help message talks specifically about a commit message, but it's conceivable that you could use this flag to provide a message for something other than a commit. Can you make this a bit more generic? It would be up to the Publisher object how to use this value, if at all.

lektor/cli.py
```diff
@@ -285,14 +288,15 @@ def deploy_cmd(ctx, server, output_path, **credentials):
         server_info = config.get_server(server)
         if server_info is None:
             raise click.BadParameter('Server "%s" does not exist.' % server,
-                                     param_hint='server')
-
+                                     param_hint='server')
```
singingwolfboy: I don't think you need this trailing whitespace :) You might want to configure your text editor to automatically remove it.

lektor/publisher.py
```diff
@@ -609,6 +609,8 @@ def write_cname(self, path, target_url):
     def publish(self, target_url, credentials=None, **extra):
         if not locate_executable('git'):
             self.fail('git executable not found; cannot deploy.')
+        if target_url.host is None:
+            self.fail('host name missing')
```
singingwolfboy, line 613: This seems unrelated to the rest of this pull request. If it's important, can you pull it out into a separate pull request?

singingwolfboy (changes requested):
Overall, this looks great, but there are a few changes that need to happen before it can be merged. We're almost there, though!

singingwolfboy:
@vassudanagunta I agree that it should be possible to enter a commit message via the GUI. However, small pull requests are a lot easier to review than big pull requests, and so it's almost always a good idea to add features incrementally. Before we can enter commit messages via the GUI, the machinery has to exist for Lektor to be able to handle custom commit messages at all, and that's what this pull request introduces. Maybe a follow-up pull request can bring this capability to the GUI!

go-bears:
@singingwolfboy we cleaned up the small edits on you noted. Thanks! Hopefully the cli message commits are ready for merging now?

lektor/cli.py
```diff
@@ -285,14 +288,15 @@ def deploy_cmd(ctx, server, output_path, **credentials):
...
         if server_info is None:
             raise click.BadParameter('Server "%s" does not exist.' % server,
-                                     param_hint='server')
-
+                                     param_hint='server')  
     try:
         event_iter = publish(env, server_info.target, output_path,
-                             credentials=credentials, server_info=server_info)
+                             credentials=credentials, server_info=server_info, 
+                             message=message)
```
singingwolfboy, line 295: You still need a test to cover this change. When the `deploy_cmd` is called with a message passed in on the command line, it needs to pass that message to the publisher. Here's how you might write a test like that:

``` python
from click.testing import CliRunner

def test_deploy_message_proxies_to_publisher(mocker):
    url = url_parse("rsync://myserver.com/path/to/website")
    mock_publisher = mocker.Mock(spec=Publisher)
    mocker.patch.dict(builtin_publishers, {'rsync': mock_publisher})
    runner = CliRunner()

    runner.invoke(deploy_cmd, ['--message', 'a custom message'])

    mock_publish_method = mock_publisher.return_value.publish
    mock_publish_method.assert_called_with(url, None, message='a custom message')
```

Let me break this down. First we set up a `url` variable, to be used later -- it's defined in the `tests/demo-project/Website.lektorproject` file in the repo. Then we'll make a fake publisher class, which responds to all the same method calls as a real publisher class but does nothing. We use `mocker.patch.dict` to slide this fake publisher into Lektor's dictionary of builtin publishers, to take the place of the `RsyncPublisher` class. This way, it'll get used for the `url` that Lektor is using! Then we set up a Click `CliRunner` [as described in the Click documentation](http://click.pocoo.org/5/testing/#basic-testing), and use that to make the actual call that is being tested. That call should cause the publisher's `publish` method to be called, but since we've swapped out the real publisher for a fake one, it doesn't actually publish anything. Then we can assert that the fake publisher was actually called with that `message` that we passed on the command line. 

Does that make sense?

singingwolfboy (changes requested):

tariquesani:
@go-bears Interested in following this up?

go-bears:
yes, I think we can get this done.

tariquesani:
Closing for now. Can be reopened if there is any activity.

nixjdm:
@go-bears Any chance you could take another look at this and finish it up? If you still have the work and finish it, you can submit another pull request for it.

Issue #286, opened 2016-09-23
malformed urls for publisher targets raise traceback instead of failing gracefully


go-bears:
if target server urls are malformed in the lektorproject file  a traceback call displays instead of failing gracefully with an error

PR #287, opened 2016-09-23
refactor url error handling for publishing targets

go-bears:
This PR resolves issue #286, It generalizes the 

```
raise PublishError('Server "%s" is not configured for a valid '
            'publishing method: %s is an unknown scheme.' % (url.host, url.scheme))
```

in cli.py to apply to all of the various publisher targets

### DIFF
--- a/lektor/cli.py
+++ b/lektor/cli.py
@@ -291,8 +291,7 @@ def deploy_cmd(ctx, server, output_path, **credentials):
         event_iter = publish(env, server_info.target, output_path,
                              credentials=credentials, server_info=server_info)
     except PublishError as e:
-        raise click.UsageError('Server "%s" is not configured for a valid '
-                               'publishing method: %s' % (server, e))
+        raise click.UsageError('%s' % e)
 
     click.echo('Deploying to %s' % server_info.name)
     click.echo('  Build cache: %s' % output_path)
--- a/lektor/publisher.py
+++ b/lektor/publisher.py
@@ -658,7 +658,7 @@ def publish(env, target, output_path, credentials=None, **extra):
     url = urls.url_parse(text_type(target))
     publisher = env.publishers.get(url.scheme)
     if url.host is None:
-        raise PublishError('missing host name')
+        raise PublishError(('Publishing target does not have a host name: %s' % target))
     if publisher is None:
         raise PublishError('Server "%s" is not configured for a valid '
             'publishing method: %s is an unknown scheme.' % (url.host, url.scheme))
--- a/lektor/publisher.py
+++ b/lektor/publisher.py
@@ -657,6 +657,9 @@ builtin_publishers = {
 def publish(env, target, output_path, credentials=None, **extra):
     url = urls.url_parse(text_type(target))
     publisher = env.publishers.get(url.scheme)
+    if url.host is None:
+        raise PublishError('missing host name')
     if publisher is None:
-        raise PublishError('"%s" is an unknown scheme.' % url.scheme)
+        raise PublishError('Server "%s" is not configured for a valid '
+            'publishing method: %s is an unknown scheme.' % (url.host, url.scheme))
     return publisher(env, output_path).publish(url, credentials, **extra)
--- a/tests/test_deploy.py
+++ b/tests/test_deploy.py
@@ -2,7 +2,6 @@ import textwrap
 from lektor.publisher import GithubPagesPublisher, RsyncPublisher, publish
 from werkzeug.urls import url_parse
 import pytest
-import re
 
 
 def test_get_server(env):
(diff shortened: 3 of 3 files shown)


lektor/publisher.py
```diff
@@ -657,6 +657,9 @@ def git(args, **kwargs):
 def publish(env, target, output_path, credentials=None, **extra):
     url = urls.url_parse(text_type(target))
     publisher = env.publishers.get(url.scheme)
+    if url.host is None:
+        raise PublishError('missing host name')
```
singingwolfboy: Can you provide a little more context in this error message? When writing error messages, consider what information you'd need to know in order to fix the problem. If I saw "missing host name" in a traceback, my first question would be, "_What_ is missing a host name?"

I would recommend something like this:

``` python
raise PublishError("Publishing target does not have a host name: %s", target)
```

singingwolfboy (changes requested):

singingwolfboy:
Thanks for the pull request! Can we get an automated test for this, as well?

go-bears:
Thanks for your time to review @singingwolfboy! We wrote tests for the error handling, but the tests are failing for Python 3.5. According the message in Travis, `.message` is no longer a valid attribute  for pytest.raises? would this be a case to use `skipif`?

go-bears:
I wrote the test to raise the exception wrong in the previous commit. However, I think I have it right now, and it passes the python 2.7 and 3.5 check. Thanks for the second pair of eyes on this!

tests/test_deploy.py
```diff
@@ -89,3 +91,25 @@ def test_rsync_command_username_in_url(tmpdir, mocker, env):
...
+
+
+def test_publish_target_url_error_handling(tmpdir, mocker, env):
+    output_path = tmpdir.mkdir("output")
+    target_url = None
+    publisher = GithubPagesPublisher(env, str(output_path))
+    
+    with pytest.raises(Exception) as exception_info:
+        mocked_publish = mocker.patch("lektor.publisher.publish")   
+        raise Exception('PublishError: Publishing target does not have a host name: ')
```
singingwolfboy, line 103: Unfortunately, this test isn't actually testing the code you've modified. The `pytest.raises()` statement asserts that an exception is raised inside the indented block, and that does indeed happen -- but it happens in the test itself, rather than in the Lektor code that is being tested. I would recommend writing this section more like this:

``` python
def test_publish_no_host(tmpdir, env):
    output_path = tmpdir.mkdir("output")
    target_url = "http://"

    with pytest.raises(PublishError) as excinfo:
        publish(env, target_url, output_path)

    expected_message = (
        "Publishing target does not have a host name: http://"
    )
    assert str(excinfo.value) == expected_message
```

Do you see how I'm calling the function that was modified, using arguments that should trigger the behavior that you've coded up, and verified that the behavior is what you want? This way, if someone accidentally modifies this behavior in the future, then the test will fail, which will alert them that something is wrong.

tests/test_deploy.py
```diff
@@ -89,3 +91,25 @@ def test_rsync_command_username_in_url(tmpdir, mocker, env):
...
+    assert 'Publishing target does not have a host name: ' in str(exception_info._excinfo[1]) 
+
+
+def test_publisher_error_handling(tmpdir, mocker, env):
+    output_path = tmpdir.mkdir("output")
+    target_url = url_parse("ghpages+https://pybee/pybee.github.io?cname=pybee.org")
+    publisher = None
+    mocked_publish = mocker.patch("lektor.publisher.publish")
+    with pytest.raises(Exception) as exception_info:   
+        raise Exception('Server "%s" is not configured for a valid publishing method')
```
singingwolfboy, line 113: Same problem here: the test isn't actually testing the code you've modified. For this, I would suggest something like this:

``` python
def test_publish_bad_scheme(tmpdir, env):
    output_path = tmpdir.mkdir("output")
    target_url = "badscheme://host/path"

    with pytest.raises(PublishError) as excinfo:
        publish(env, target_url, output_path)

    expected_message = (
        "Server "host" is not configured for a valid publishing method: "
        "badscheme is an unknown scheme."
    )
    assert str(excinfo.value) == expected_message
```

lektor/cli.py
```diff
@@ -291,8 +291,7 @@ def deploy_cmd(ctx, server, output_path, **credentials):
         event_iter = publish(env, server_info.target, output_path,
                              credentials=credentials, server_info=server_info)
     except PublishError as e:
-        raise click.UsageError('Server "%s" is not configured for a valid '
-                               'publishing method: %s' % (server, e))
+        raise click.UsageError('%s' % e)
```
singingwolfboy, line 294: It looks like you're just trying to turn the PublishError into a string. This seems like a clearer way to do that:

``` python
raise click.UsageError(str(e))
```

singingwolfboy (changes requested):

go-bears:
appreciate the time you're taking to help us learn how to write tests appropriately, @singingwolfboy ! It was hard to figure out to mock out the calls that the function made. Thanks for helping us do the incremental improvements

nixjdm:
@go-bears want to take another shot at finishing these tests up :)

yagebu:
I'm closing this very stale and unfinished PR

PR #293, opened 2016-10-06
Unless make build-js is done JS tests fail

tariquesani:
Unless make build-js is done JS tests fail, corrected the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,5 +43,6 @@ If you want to run the test suite instead:
 $ virtualenv venv
 $ . venv/bin/activate
 $ pip install --editable .[test]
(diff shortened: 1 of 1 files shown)


singingwolfboy:
This is not the right way to solve this problem. If running the tests requires building the Javascript, then the Makefile should build the Javascript automatically when you run `make test`. Can you mark it as a dependency of the `test-js` block in the Makefile?

tariquesani:
This appears to be something different than just running make build-js. Editing Makefile does not really solve it. 

Closing this pull request and creating an issue ticket instead.

PR #296, opened 2016-10-08
unit test for PR #284

go-bears:
added working unit test for git commit feature on command line

### DIFF
--- a/tests/test_deploy.py
+++ b/tests/test_deploy.py
@@ -102,4 +102,14 @@ def test_cli_commit_msg(tmpdir, mocker, env):
     publisher.publish.extra = {'message': expected}
     
     assert mocked_deploy.extra.get('message') == expected
+
+def test_cli_commit_msg_empty(tmpdir, mocker, env):
+    output_path = tmpdir.mkdir("output")
+    publisher = GithubPagesPublisher(env, str(output_path))
+    repo_path = tmpdir.mkdir("repo")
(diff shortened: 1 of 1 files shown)


go-bears:
@luckyrome @tjarratt 
@arbonap and I actually wrote a test on our own on our feature!

item4:
Can you merge #284 and #296?

go-bears:
not sure how to merge two different pull requests after I've made them? i think the decision to merge is up to @singingwolfboy and @mitsuhiko ?

singingwolfboy:
@item4: only people in the Lektor organization on GitHub have permission to merge pull requests, so @go-bears can't do so. I'm glad you're excited about this feature, though!

@go-bears @arbonap: Thanks for the work you've done here, but unfortunately, these tests aren't actually testing the code you wrote. They're on a different pull request, so the code you wrote in #284 doesn't exist in this pull request! I think the tests are passing because you mocked out the functions you're trying to test.

In order to add tests for your feature, you want to do so on the same branch that you used for your original pull request. Think about how you would manually test the feature -- what content would you write, what commands would you run, and how would you look at the result to decide if it's working or not? Then automate that in code. I wonder if it's possible to do remote paircoding on this...

singingwolfboy:
I'm actually going to close this pull request, because as I said, the tests for a pull request belong in _that_ pull request, not in a separate one. You don't need this one!

PR #302, opened 2016-10-21
Fixed typo's in Dutch translations

redrohX:
I noticed a couple of typo's in the Dutch translation while checking out Lektor.


singingwolfboy:
Thank you! I don't speak Dutch, so I can't actually review this, but I trust that your fixes are correct.

PR #306, opened 2016-10-27
Ignore the cache directory when watching for changes

emilyalbini:
In my setup I have the lektor cache directory in a subfolder of the project (changed with the `$XDG_CACHE_HOME` environment variable here in linux), but since `lektor server` writes in that directory after each build, when I start the development server lektor starts spamming empty builds, without stopping.

This patch tweaks the watcher to ignore events from the cache directory Lektor uses.

### DIFF
--- a/tests/test_watcher.py
+++ b/tests/test_watcher.py
@@ -1,0 +1,23 @@
+import py.path
+import functools
+
+from lektor import utils
+from lektor import watcher
+
+
+def test_is_interesting(env):
+    cache_dir = py.path.local(utils.get_cache_dir())
+    build_dir = py.path.local("build")
+
+    w = watcher.Watcher(env, str(build_dir))
+
+    # This partial makes the testing code shorter
+    is_interesting = functools.partial(w.is_interesting, 0, "generic")
+
+    assert is_interesting("a.file")
+    assert not is_interesting(".file")
+    assert not is_interesting(str(cache_dir / "another.file"))
+    assert not is_interesting(str(build_dir / "output.file"))
+
+    w.output_path = None
+    assert is_interesting(str(build_dir / "output.file"))
--- a/lektor/watcher.py
+++ b/lektor/watcher.py
@@ -5,6 +5,7 @@ from watchdog.observers import Observer
 from watchdog.events import FileSystemEventHandler, DirModifiedEvent, FileMovedEvent
 
 from lektor._compat import queue
+from lektor.utils import get_cache_dir
 
 # Alias this as this can be called during interpreter shutdown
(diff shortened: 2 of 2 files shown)


singingwolfboy:
This looks good, but can you add an automated test, please?

emilyalbini:
Tests added!

tests/test_watcher.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def test_is_interesting(env):
+    cache_dir = os.path.abspath(utils.get_cache_dir())
+    build_dir = os.path.abspath("build")
+
+    w = watcher.Watcher(env, build_dir)
+
+    # This makes the testing code shorter
+    partial = functools.partial(w.is_interesting, 0, "generic")
```
singingwolfboy: Can you rename `partial` to `is_interesting`? That makes the code clearer.

tests/test_watcher.py
```diff
@@ -0,0 +1,23 @@
+import os.path
+import functools
+
+from lektor import utils
+from lektor import watcher
+
+
+def test_is_interesting(env):
+    cache_dir = os.path.abspath(utils.get_cache_dir())
```
singingwolfboy: Rather than calling `utils.get_cache_dir()`, I recommend using the [`tmpdir`](http://docs.pytest.org/en/latest/tmpdir.html) fixture from `pytest` to get a temporary directory. That way, you get to use the [py.path API](http://py.readthedocs.io/en/latest/path.html), to do nice things like this:

``` python
def test_is_interesting(env, tmpdir):
    cache_dir = tmpdir / "cache"  # construct a path without the verbose `os.path.abspath` function
    cache_dir.ensure(dir=True)  # make sure it exists on the filesystem as a directory
    build_dir = tmpdir / "build"
    build_dir.ensure(dir=True)
    # ...
    assert not is_interesting(cache_dir / "another.file")  # much shorter
```

singingwolfboy (changes requested):

tests/test_watcher.py
emilyalbini: Wouldn't this create a directory on the filesystem? Even if I use `os.path.abspath`, I don't write anything during the test.

tests/test_watcher.py
emilyalbini: Will do!

emilyalbini:
Fixed and squashed the commits (sorry for multiple pushes, I had a few problems with my git config). About `py.path`: I switched to it, but without using `tmpdir`, because this way no filesystem access is done.

singingwolfboy (approved):

PR #321, opened 2016-11-12
Scale rotated images to correct width

### LINKED ISSUE #316: Rotated thumbnails do not report correct dimensions
I use Lektor with [PhotoSwipe](http://photoswipe.com/), which expects displayed image to be rotated correctly (that is, it does not read the EXIF header to achieve correct orientation).  I don't want to modify my original uploaded images if I can avoid it, but I noticed that Lektor's thumbnail feature generates images with the correct orientation thanks to imagemagick's `auto-orient` parameter ([code](https://github.com/lektor/lektor/blob/422af5b7e2732b88a4116ee510939b3bb2309d87/lektor/imagetools.py#L373)).

However, the `width` and `height` of that thumbnail are reported as if the image had not been rotated. A 2000x1500 (w x h) image with 90deg rotation information in the EXIF header that is scaled using `thumbnail(1000)` will become a 750x1000 image, but the data returned by `width` and `height` will be 1000x750.

Therefore,
1) an image that is expected to be 1000px wide will actually be 750px wide, and
2) there is no way to detect that.

In a perfect world,
- Lektor would expose a method to read the rotation information from the EXIF header,
- scale the thumbnail to the specified width
- return the correct dimensions of the generated image via `height` and `width`.

tgpfeiffer:
Fixes #316 and the three wishlist items therein.


lektor/imagetools.py
```diff
@@ -213,6 +213,11 @@ def documentname(self):
     def description(self):
         return self._get_string('Image ImageDescription')        
 
+    @property
+    def rotated(self):
```
singingwolfboy: Since this is a boolean, can you change the name to `is_rotated`?

lektor/imagetools.py
```diff
@@ -213,6 +213,11 @@ def documentname(self):
...
+        # cd. <http://www.impulseadventure.com/photo/images/orient_flag.gif>
```
singingwolfboy: Can you add a textual description as a docstring, instead of linking to an image? This image may not always exist at this URL, and a docstring is clearer anyway.

lektor/imagetools.py
```diff
@@ -213,6 +213,11 @@ def documentname(self):
...
+        return self._get_int('Image Orientation') in [5, 6, 7, 8]
```
singingwolfboy: Nitpick: you can make the integers a set instead of a list.

``` python
return self._get_int('Image Orientation') in {5, 6, 7, 8}
```

lektor/imagetools.py
```diff
@@ -344,7 +349,16 @@ def get_quality(source_filename):
...
-def computed_height(source_image, width, actual_width, actual_height):
+def computed_height(source_image, format, width, actual_width, actual_height):
+    # If the file is a JPEG file and the EXIF header shows that the image
+    # is rotated, imagemagick will auto-orient the file. That is, a 400x200
+    # file where the target width is 100px will *not* be converted to a
+    # 100x50 image, but to 100x200.
+    if format == 'jpeg':
+        exif = read_exif(open(source_image, 'rb'))
+        if exif.rotated:
+            return int(float(actual_width) *
```
singingwolfboy: Instead of re-implementing the logic on this line that _also_ exists on the next line, can you just swap the values?

``` python
if exif.rotated:
    actual_width, actual_height = actual_height, actual_width
return int(float(actual_height) * (float(width) / float(actual_width)))
```

lektor/imagetools.py
```diff
@@ -365,15 +379,15 @@ def process_image(ctx, source_image, dst_filename, width, height=None,
...
-    cmdline = [im, source_image]
+    cmdline = [im, source_image, '-auto-orient']
     if crop:
         cmdline += ['-resize', resize_key + '^',
                     '-gravity', 'Center',
                     '-extent', resize_key]
     else:
         cmdline += ['-resize', resize_key]
 
-    cmdline += ['-auto-orient', '-quality', str(quality), dst_filename]
```
singingwolfboy, line 376: Any particular reason why you moved the `-auto-orient` parameter earlier? It's fine either way, I'm just curious if there's some rationale that I'm missing.

singingwolfboy (changes requested):
This is a great start, but there are a few changes that need to be addressed. We also need an automated test to cover this change.

lektor/imagetools.py
tgpfeiffer: Thanks for your feedback, I will fix the points you mentioned. Just for this one: ImageMagick is very picky about the order of arguments:

```
$ convert dsc_tp_8801.jpg -resize 300x400 -auto-orient test.jpg && identify test.jpg
test.jpg JPEG 225x300 225x300+0+0 8-bit sRGB 52.4KB 0.000u 0:00.009
$ convert dsc_tp_8801.jpg -auto-orient -resize 300x400 test.jpg && identify test.jpg
test.jpg JPEG 300x400 300x400+0+0 8-bit sRGB 72.7KB 0.000u 0:00.000
```

tgpfeiffer:
I fixed all the issues you pointed out and added a test. I also updated the PR to the docs on the website.

singingwolfboy (approved):

singingwolfboy:
This is great, thank you! ✨❤️

PR #332, opened 2016-12-08
Restore the about page in the example project

jesuslosada:
It looks like the about page in the example project was accidentally removed in commit 7cab136c3cbd36c0413db5e83f2c9e355dac1ff7. This PR restores the missing page.


singingwolfboy:
Thanks for catching this! 👍

PR #333, opened 2016-12-08
Fix the blog index in the example project

jesuslosada:
I've updated the example project based on the current quickstart template. This PR fixes some pagination issues in the blog index.


singingwolfboy:
This looks great!

jesuslosada:
Thanks for the fast response!

PR #342, opened 2017-01-07
Make GH Pages branch detection case-insensitive

uranusjr:
GitHub usernames (and org names) are case-insensitive, and for a user named Foo the user page's URL would be normalized to <foo.github.io>. This patch mimics this behavior by converting both target_url's host and path to lowercase before comparing.


singingwolfboy:
Thanks!

PR #350, opened 2017-02-02
Modified date field type from text to date.

aavrug:


tariquesani:
What will this change do?

aavrug:
Previously the date fields were displaying just as text field, after this change the date picker popup will display for selection of dates.

tariquesani:
@singingwolfboy  I think we should merge this PR with master. It is a very small change enables the react Datepicker and does not break anything as far as I can tell.

singingwolfboy:
@tariquesani Seems reasonable to me. You want to do so?

tariquesani:
Sure, going ahead.

aavrug:
@tariquesani Thanks for merging.

Issue #352, opened 2017-02-06
Tests failing for master branch


tariquesani:
Have attached the complete output but typically getting errors similar to 

`import file mismatch:
imported module 'data.clientmodule_test' has this __file__ attribute:
  /home/tariquesani/Projects/lektor/venv/lib/python2.7/site-packages/astroid/tests/testdata/python2/data/clientmodule_test.py
which is not the same as the test file we want to collect:
  /home/tariquesani/Projects/lektor/venv/lib/python2.7/site-packages/astroid/tests/testdata/python3/data/clientmodule_test.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
 ERROR collecting venv/lib/python2.7/site-packages/astroid/tests/testdata/python3/data/suppliermodule_test.py 
`

[testreport.txt](https://github.com/lektor/lektor/files/753733/testreport.txt)

coreygrunewald:
This is the same thing I saw locally, however, the CI build for my PR (https://travis-ci.org/lektor/lektor/jobs/198731896) doesn't see the same issue.

singingwolfboy:
This looks like a problem with [`asteroid`](https://github.com/PyCQA/astroid), which is a part of [`pylint`](https://pylint.org/). You can run the tests without pylint, if you want: just delete the [`pytest.ini`](https://github.com/lektor/lektor/blob/master/pytest.ini) file and run `make test` again.

tariquesani:
May be I am doing something wrong but deleting  [`pytest.ini`](https://github.com/lektor/lektor/blob/master/pytest.ini) did not change anything. I tried after removing the  __pycache__  and .pyc files.

singingwolfboy:
@tariquesani hmm, that's odd. Are you able to reproduce this with a new virtualenv? (If you're not familiar with virtualenv, it's [an installable tool for Python 2](https://pypi.python.org/pypi/virtualenv), which has been [pulled into the Python standard library in Python 3](https://docs.python.org/3/library/venv.html).) Try creating a new virtualenv, installing Lektor's dependencies into that virtualenv, and then running the tests. If you can still reproduce the issue, please attach a traceback of all the commands you ran, from creating the virtualenv to getting the error.

tariquesani:
OK, let me try with the new virtualenv. Thanks for being so helpful.

tariquesani:
Not much difference. I am on a default install of Ubuntu 16.04. Here is what I did 

```
[...]

PR #354, opened 2017-02-09
Implement standard for JS linting

singingwolfboy:
We should have a linting system for the Javascript in this project, and [Standard](http://standardjs.com/) is a good choice. I'm open to other options, though.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ coverage-python: test-python
 
 test-js: build-js
 	@echo "---> running javascript tests"
+	@cd lektor/admin; npm run lint
 	@cd lektor/admin; npm test
 
 coverage-js: test-js
--- a/gui/static/js/app.jsx
+++ b/gui/static/js/app.jsx
@@ -106,7 +106,7 @@ class App extends Component {
     let projectPath = this.state.projectData
       ? this.state.projectData.project_path : null
 
-    function couldNotOpen(message) {
+    const couldNotOpen = (message) => {
       dialog.showErrorBox(i18n.trans('FAILED_TO_OPEN_CONTENT_FILE'),
                           message.toString())
     }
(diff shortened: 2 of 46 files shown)


singingwolfboy:
@coreygrunewald, could you review this please?

coreygrunewald:
I'll take a closer look this weekend, but in general, I would recommend creating a git pre-push script that runs the JS tests and linting, and then a Make task that symlinks the file into `.git/hooks/pre-push`. That way if people are inclined to symlink the hook, they will be notified of a failure before even pushing to `origin`.

singingwolfboy:
@coreygrunewald: still hoping you can review this, if possible. The failed GitHub checks are due to code coverage issues, and for this pull request, I think it's totally reasonable to merge in spite of that, since there's no change to the functionality.

tariquesani:
@singingwolfboy Can you give a hint on how to solve those failing tests? I just asked someone to write a test for their PR only to realise that they are broken...

singingwolfboy:
@tariquesani There are two different checks happening here: whether the automated tests pass, and how well they cover the codebase. The second check is the problem here -- it's triggered when a pull request changes a lot of code that is not covered by automated tests. The fix is to write new automated tests that cover the lines that have been changed.

I don't think the tests are broken, it's just that there is a lot of untested code. We should incrementally write new tests to cover the untested code, and theoretically I should do that in this pull request. But this pull request is already large enough as it is, and as I said, there are no changes to the functionality, so it should be a safe change. That's why I'm advocating to merge this even when Codecov is reporting a failure. Does that make sense?

tariquesani:
Understood, I got confused by what I have notice in https://github.com/lektor/lektor/issues/352

singingwolfboy:
I'd really like to get this merged so that other PRs can build on top of it. It it needs further fixes, we can get them in future PRs.

coreygrunewald:
Got wrapped up with work, sorry to leave this one hanging. In general though, LGTM.

PR #416, opened 2017-08-01
Enable publishing from different filesystem

jgoodson:
os.link() fails if files are on different filesystems. This change enables fallback to shutil.copy() in this case.


nixjdm:
I added a comment on the OSError line. It looks like that whole method is doesn't have any tests written for it yet, which is unfortunately common. I'll take the bug fix as is. You can't easily add this to an existing test that doesn't exist, but if you want to make a new test for this (and likely other parts of this method) I'll welcome it gladly.

nixjdm:
That Appveyor tests that fail look unrelated, and will get there own tickets.

PR #427, opened 2017-08-26
Fixing arrow function errors ref #390

nixjdm:
Removing problematic arrow functions added in commit 33972ed (see https://github.com/lektor/lektor/issues/390#issuecomment-325061735). Also added in a few missing parens around arrow function params of arity 1. They aren't strictly required, but that is the norm in this project, and a common rule.

Many arrow functions already existed in the project and were untouched by 33972ed, and are untouched here. It turns out most arrow functions in 33972ed are also fine, just that most that used `this` inside of them were bad. Those are reverted here. I also manually tested this with the lektor-website repo, and while there are still unrelated warnings in the console, I didn't see any related exceptions. I also grepped through the app.js and vendor.js and did not see any suspicious looking `undefined`s anywhere.

Edit after failed tests / fix: The reverted functions didn't conform to the new linting rules. Now they do.

Ref #390

### DIFF
--- a/lektor/admin/static/js/bootstrap-extras.jsx
+++ b/lektor/admin/static/js/bootstrap-extras.jsx
@@ -1,7 +1,7 @@
 import $ from 'jquery'
 
-$(document).ready(function() {
-  $('[data-toggle=offcanvas]').click(function() {
+$(document).ready(function () {
+  $('[data-toggle=offcanvas]').click(function () {
     const target = $($(this).attr('data-target') || '.block-offcanvas')
     const isActive = target.is('.active')
     target.toggleClass('active', !isActive)
--- a/lektor/admin/static/js/bootstrap-extras.jsx
+++ b/lektor/admin/static/js/bootstrap-extras.jsx
@@ -1,7 +1,7 @@
 import $ from 'jquery'
 
-$(document).ready(() => {
-  $('[data-toggle=offcanvas]').click(() => {
+$(document).ready(function() {
+  $('[data-toggle=offcanvas]').click(function() {
     const target = $($(this).attr('data-target') || '.block-offcanvas')
     const isActive = target.is('.active')
     target.toggleClass('active', !isActive)
--- a/lektor/admin/static/js/widgets/flowWidget.jsx
+++ b/lektor/admin/static/js/widgets/flowWidget.jsx
@@ -57,7 +57,7 @@ const parseFlowFormat = (value) => {
 
 const serializeFlowFormat = (blocks) => {
   let rv = []
-  blocks.forEach(block => {
+  blocks.forEach((block) => {
     const [blockName, lines] = block
     rv.push('#### ' + blockName + ' ####\n')
     lines.forEach((line) => {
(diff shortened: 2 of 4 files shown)


nixjdm:
I see standard wants a little more whitespace. All amend and update this very soon, and also make a ticket for running the linter on commit :)

nixjdm:
@coreygrunewald you reviewed the original, perhaps you're interested in this too?

vanadium23:
@nixjdm imho, it's bad revert. `this` doesn't change if you use short had as arrow function & standard function declaration. Better to dig in, why errors are here. Can you show example of console error or how to reproduce it?
Also, I don't see any mistakes in reverted code.

nixjdm:
@vanadium23 Thanks for taking a look.

To reproduce the original issue:

1. checkout and run lektor on master
1. Open up the site [lektor-website](https://github.com/lektor/lektor-website)
1. Go to /admin, then (with console open), go to "Edit" the homepage.

You should see the admin fields not be present and the console throw the error
```
flowWidget.jsx?e153:211 Uncaught (in promise) TypeError: Cannot read property 'props' of undefined
    at Object.renderFormField (eval at <anonymous> (app.js:658), <anonymous>:262:30)
    at Array.map (<anonymous>)
    at eval (eval at <anonymous> (app.js:628), <anonymous>:247:11)
    at Array.forEach (<anonymous>)
    at Object.renderFieldRows (eval at <anonymous> (app.js:628), <anonymous>:239:8)
    at eval (eval at <anonymous> (app.js:658), <anonymous>:287:28)
    at Array.map (<anonymous>)
    at Object.renderBlocks (eval at <anonymous> (app.js:658), <anonymous>:281:29)
    at Object.render (eval at <anonymous> (app.js:658), <anonymous>:382:12)
    at eval (eval at <anonymous> (vendor.js:889), <anonymous>:795:21)
```
This particular error is caused by the `this` in renderFormField in lektor/admin/static/js/widgets/flowWidget.jsx being undefined. I'm not super experienced with React, but if I understand what I've read about this sort of thing, this is because [arrow functions do not bind `this`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#No_binding_of_this), and a [`this` used inside an arrow function has the same value of it's enclosing context](https://stackoverflow.com/questions/28371982/what-does-this-refer-to-in-arrow-functions-in-es6?rq=1
), which in this case means undefined, since `this` is undefined in the enclosing context. 

If you `make build-js` on master, and then grep through lektor/admin/static/gen/app.js you'll find other instances of problematic `undefined`s that seem to be caused by the same thing. Every arrow function I turned back into an regular function has the effect of removing at least one `undefined` in the compiled app.js file.

Does that make sense? Again, I'm relatively new to React, so if my thinking is wrong please tell me. Thanks

vanadium23:
@nixjdm re-check your code and you're right. All function, that calls `.bind(this)` need to be converted back from arrow function to declarative one.
Simple example to show, why this error happens:
```js
const consoleThis = () => {console.log(this)};
consoleThis();  // window
consoleThis.bind({a: 'a'})();  // window
```

LGTM

nixjdm:
Excellent. Well with that I really don't want to leave this particular fix just hanging out here. Thanks again for giving it a look @vanadium23. Merging.

Issue #428, opened 2017-08-26
Add Git script to run linter before pushing


nixjdm:
@coreygrunewald had a great suggestion [here](https://github.com/lektor/lektor/pull/354#issuecomment-278771962) to add a git pre-push script. This issue is for making that script that runs the js linter before a push such that it displays any error output and ends the push if there is a problem. Instructions should also be added to the README for how to symlink this to `.git/hooks/pre-push` in the instructions section for doing development on Lektor.

vanadium23:
@nixjdm I'd like to make pr, can you assign issue to me?

jab:
In case you haven't seen, http://pre-commit.com is a great way to automate adding and running best-of-breed git pre-commit hooks, both locally and as part of the CI build.

vanadium23:
@jab yeah, agree (:

nixjdm:
@vanadium23 Github won't let me assign things to people not in the organization (a policy I question), but I can definitely informally give this to you! Please go right ahead.

I'm open to this being either pre-commit or pre-push. Doesn't make a big difference to me. Perhaps devs would find pre-commit less annoying than pre-push.

@vanadium23 Just the linter is fine for this issue, but if you have ideas for similar handy hooks, I'm all ears. For instance a trailing-whitespace commit block is pretty standard and could be added.

PR #429, opened 2017-08-31
Add simple bash script for pre commit js linting

vanadium23:
Also updated Makefile with task for install hooks
Ref #428.

This is a very simple and naive implementation of lint before commit, but while scripting I've done some research of tools:
1. for [pre-commit](http://pre-commit.com/) from yelp we need to create plugins for standard linter.
2. there are various soltuions based for npm, like this [one](https://github.com/okonet/lint-staged#working-from-a-subdirectory), but after local testing they are not working well in sub-directories (although with gitDir proposed in docs).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,3 +38,6 @@ coverage: coverage-python coverage-js
 
 osx-dmg:
 	$(MAKE) -C gui osx-dmg
+
+install-git-hooks:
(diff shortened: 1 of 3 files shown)


nixjdm:
This is fine for now. I'm open to changing it to using pre-commit later if that becomes easier with Standard JS. This does have the benefit of being pretty straight forward and transparent, which I like, even if it's little clunky. Thanks!

PR #437, opened 2017-09-07
[Tests] we need to use system file encoding, if on windows

### LINKED ISSUE #403: Test failing on windows (see appveyor) test_unicode_project_folder
```
_________________________ test_unicode_project_folder _________________________
tests\test_unicode.py:21: in test_unicode_project_folder
    prog, _ = builder.build(pad.root)
lektor\builder.py:1098: in build
    prog.build()
lektor\build_programs.py:101: in build
    _build(artifact, self.build_artifact)
lektor\build_programs.py:91: in _build
    ctx = gen.build_artifact(artifact, build_func)
lektor\builder.py:1058: in build_artifact
    build_func(artifact)
c:\python27\lib\contextlib.py:24: in __exit__
    self.gen.next()
lektor\builder.py:803: in update
    self.finish_update(ctx)
lektor\builder.py:863: in finish_update
    ctx.referenced_virtual_dependencies.values())
lektor\builder.py:738: in _memorize_dependencies
    self._auto_deferred_update_operation(operation)
lektor\builder.py:783: in _auto_deferred_update_operation
    f(con)
lektor\builder.py:673: in operation
    source = self.build_state.to_source_filename(source)
lektor\builder.py:146: in to_source_filename
    return self.path_cache.to_source_filename(filename)
lektor\builder.py:912: in to_source_filename
    (filename, self.env.root_path))
E   ValueError: The given value (u'C:\\projects\\lektor\\tests\\\xc3\xbcnic\xc3\xb6de-project\\templates\\page.html') is not below the source folder (u'C:\\projects\\lektor\\tests\\\xfcnic\xf6de-project')
```

vanadium23:
Problem was, because we decode latin-1 string with utf-8.

Fixes #403

### DIFF
--- a/lektor/environment.py
+++ b/lektor/environment.py
@@ -13,14 +13,14 @@ from inifile import IniFile
 from werkzeug.urls import url_parse
 from werkzeug.utils import cached_property
 
-from lektor._compat import iteritems, string_types
+from lektor._compat import iteritems, string_types, PY2
 from lektor.context import (config_proxy, get_asset_url, get_ctx, get_locale,
     site_proxy, url_to)
 from lektor.i18n import get_i18n_block
 from lektor.markdown import Markdown
 from lektor.pluginsystem import PluginController
 from lektor.utils import (bool_from_string, format_lat_long, secure_url,
-    tojson_filter)
+    tojson_filter, is_windows)
 
 
 # Special value that identifies a target to the primary alt
@@ -215,7 +215,13 @@ class CustomJinjaEnvironment(jinja2.Environment):
         try:
             rv = jinja2.Environment._load_template(self, name, globals)
             if ctx is not None:
-                ctx.record_dependency(rv.filename)
+                filename = rv.filename
+                if PY2 and is_windows:
+                    try:
+                        filename = filename.decode('utf-8')
+                    except UnicodeDecodeError:
+                        pass
+                ctx.record_dependency(filename)
             return rv
         except jinja2.TemplateSyntaxError as e:
             if ctx is not None:


vanadium23:
@nixjdm I've found root cause, that os.path returns path in latin-1 on win32, but don't know how to fix it properly. Can you help me?

nixjdm:
@vanadium23 You're talking about that there is still this error, right?

```
_________________________ test_unicode_project_folder _________________________
tests\test_unicode.py:20: in test_unicode_project_folder
    pad, builder = get_unicode_builder(tmpdir)
tests\test_unicode.py:13: in get_unicode_builder
    env = Environment(proj)
lektor\environment.py:402: in __init__
    self.root_path = os.path.abspath(project.tree)
E   AttributeError: 'NoneType' object has no attribute 'tree'
```

Let me try to think through this with you. I think this is happening because https://github.com/lektor/lektor/blob/master/tests/test_unicode.py#L11 is passing in a unicode path, but the `sys.getfilesystemencoding()` is returning `latin1`. What does https://github.com/lektor/lektor/blob/master/lektor/project.py#L44 set the path to, when the argument is unicode and the `os` is using `latin1`? I'm wondering if you're getting down to the caught `OSError` and returning `None`.

I don't have a solution for you at the moment. I hope that helps.

vanadium23:
@nixjdm finally solved it, this is more jinja2 problem and currently I've added workaround to fix it.

nixjdm:
It took me a while but I think I basically understand this now. I do have a question though. Is this really just a problem on PY2? Wouldn't the default encoding be latin1 on PY3 as well, and we'd run into the same problem? Maybe I should just merge this and see. If PY3 fails, appveyor will catch it and we can remove the PY2 bit here. What do you think @vanadium23 ?

vanadium23:
@nixjdm I've next PR #442 and in it's build you see different errors for PY3 on windows. Also I don't reproduce this error with python3 on local machine (:

nixjdm:
Sounds good to me. Good work @vanadium23 !

Issue #431, opened 2017-09-05
Fix React warnings


vanadium23:
We couldn't upgrade react in future, if we would not fix warnings from it.

```
Accessing PropTypes via the main React package is deprecated, and will be removed in  React v16.0. Use the latest available v15.* prop-types package from npm instead. For info on usage, compatibility, migration and more, see https://fb.me/prop-types-docs
printWarning
Accessing createClass via the main React package is deprecated, and will be removed in React v16.0. Use a plain JavaScript class instead. If you're not yet ready to migrate, create-react-class v15.* is available on npm as a temporary, drop-in replacement. For more info see https://fb.me/react-create-class
printWarning
```

runfalk:
This is not only related to Lektor itself, but also to its dependencies such as `react-router`

nixjdm:
Fixed in #444

runfalk:
This isn't really fixed, as long as `react-router` is not upgraded. There may be more dependencies that need upgrading to get rid of this warning too.

nixjdm:
Agreed. I closed this too early.

PR #444, opened 2017-09-12
JS warning fixes

runfalk:
I did some things to fix trivial JS warnings. This properly fixes the test suite warnings (`make test-js` and fixes everything Lektor is responsible when it comes to PropTypes usage (#431).

Lektor does however depend on `react-router` 2.8.1 which does use `React.PropTypes`. There may be other dependencies that use `PropTypes` in a deprecated way too.

### DIFF
--- a/lektor/admin/package.json
+++ b/lektor/admin/package.json
@@ -31,7 +31,6 @@
     "prop-types": "^15.5.10",
     "querystring": "^0.2.0",
     "react": "^15.3.0",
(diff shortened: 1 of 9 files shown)


vanadium23:
Great job! Looks good to me :+1:

vanadium23:
@nixjdm can we merge this?

nixjdm:
Yep, lgtm. Thanks for getting rid of those pesky warnings @runfalk :)

PR #450, opened 2017-09-14
[Fix] change url for appveyor build badge

vanadium23:
Since it's not unknown (:


nixjdm:
I changed the travis one too to point to the master branch. I like your change for appveyor better than mine. This is the first time appveyor is marked as passing!

Issue #449, opened 2017-09-13
Update Node packages


nixjdm:
Time for another round of updating our packages as much as we can without breaking anything. This should regularly be done to keep compatibility with our dependencies. If we discover an update would break something, that can be fixed here or we can put that particular package update and fix together in a separate ticket. Ample testing and care should be taken with this issue. I'd like to have this pulled into master not long after 3.1 is released so we have plenty of time to test it on master before reaching 3.2.

Suggested by: https://github.com/lektor/lektor/issues/430#issuecomment-328800515

on-hold because of #457

rlaverde:
Using a tool for monitor dependencies will prevent them for getting outdated.

I don't know too much about the js/node world, but after some google searching I found https://greenkeeper.io/ and https://david-dm.org/

nixjdm:
Since we've had further discussion on #457 / #458, I'm removing the on-hold label.

runfalk:
This has been done for now I think.

Issue #420, opened 2017-08-10
Modernize Desktop application build process


nixjdm:
Releasing 3.0 / 3.0.1, and now working on a pre release of 3.1, I've come to the opinion that it really needs an update. Most of that code hasn't been touched in 2 years, and it's showing. It's a struggle to get it to work. I think it needs to be updated. So far, I've found that:

1. It's generally out of date and prone to breaking by just sitting there. I had to make a few changes to produce the DMGs for the previous releases, and I'm running into new errors now.
1. It is undocumented.
1. It can't use Python 3. At the very least a dependency [node-gyp](https://github.com/nodejs/node-gyp) is not compatible with Python 3, and I definitely won't be surprised if a lot of other things break too. Since we're supporting Python 3 now, the dmg should use it if possible.
1. The whole gui is based on a very outdated [electron-prebuilt](https://github.com/electron-userland/electron-prebuilt). This cannot be trivially updated to use the modern electron.
1. I'd like to give some thought to make binaries for other platforms. If we can do this it would be best done using as many cross-platform tools as possible. Making inroads to this may mean some big changes to the gui sub-project
1. Currently the main project and the gui sub-project have separate dependency lists, namely, their package.json files for node. That's asking for trouble.
1. There is no automated testing of the build process or of the build products. Granted, setting this up is non-trivial, but it would be very useful.

nixjdm:
I just made a label for other issues related to the Desktop App. https://github.com/lektor/lektor/issues?q=is%3Aissue+is%3Aopen+label%3Adesktop-app

yagebu:
Hey, I'm a happy user of Lektor, thanks for your work on the project! I'm a maintainer of [Fava](https://github.com/beancount/fava), also a Python project that has an Electron GUI. Our situation is slightly different as the GUI wraps the web interface and doesn't just act as a "supervisor" of the process. There are still enough similarities so that our approach might give some ideas that could work for Lektor too. The build process goes as follows:

1. We use Pyinstaller to compile Fava to a single executable (currently on Python 3.5).
1. Use electron-builder to create an installer for the Electron GUI (`dmg` for MacOS, `AppImage` for Linux). We don't do a Windows build currently, mainly because our main dependency Beancount doesn't compile easily on Windows - otherwise that shouldn't be too hard to set up. When run on Travis, the builds get uploaded to Github as a pre-release.

[...]

PR #451, opened 2017-09-15
Webpack upgrade

runfalk:
This became a three part story. Firstly there was one more instance where `React.PropTypes` was used (it's still used in some dependencies).

The second commit is a complete update of Webpack to version 3.6.0. Since I can't test the electron build process yet I haven't done anything with that Webpack installation. Do we even want to use the current setup there? I took time to update Webpack, Babel and some of the loaders to get everything working. This fixes some parts of #449. One change I did was switching the source maps to proper ones instead of the `cheap-module-eval-source-map`. At least my Firefox versions (57 nightly and current stable) were unable to use those source maps. This adds about 10%-20% overhead to the JS build process. The maximum time on my (built 2010) computer is however still less than 10 seconds.

The third commit is a fix for broken usage of `this` in `multiWidgets.jsx`. I think this issue may be in master too, but I fixed it during my tests of the Webpack update to make sure nothing else was broken.

### DIFF
--- a/lektor/admin/static/js/widgets/multiWidgets.jsx
+++ b/lektor/admin/static/js/widgets/multiWidgets.jsx
@@ -55,7 +55,7 @@ const CheckboxesInputWidget = React.createClass({
     let {className, value, placeholder, type, ...otherProps} = this.props  // eslint-disable-line no-unused-vars
     className = (className || '') + ' checkbox'
 
-    const choices = this.props.type.choices.map(function (item) {
+    const choices = this.props.type.choices.map((item) => {
       return (
         <div className={className} key={item[0]}>
           <label>
--- a/lektor/admin/package-lock.json
+++ b/lektor/admin/package-lock.json
@@ -1,0 +1,7186 @@
+{
+  "name": "lektor",
+  "version": "0.0.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abab": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/abab/-/abab-1.0.3.tgz",
+      "integrity": "sha1-uB3l9ydOxOdW15fNg08wNkJyTl0=",
+      "dev": true
+    },
+    "acorn": {
+      "version": "2.7.0",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-2.7.0.tgz",
+      "integrity": "sha1-q259nYhqrKiwhbwzEreaGYQz8Oc=",
+      "dev": true
+    },
+    "acorn-dynamic-import": {
+      "version": "2.0.2",
+      "resolved": "https://registry.npmjs.org/acorn-dynamic-import/-/acorn-dynamic-import-2.0.2.tgz",
+      "integrity": "sha1-x1K9IQvvZ5UBtsbLf8hPj0cVjMQ=",
+      "dev": true,
+      "requires": {
+        "acorn": "4.0.13"
+      },
+      "dependencies": {
+        "acorn": {
+          "version": "4.0.13",
+          "resolved": "https://registry.npmjs.org/acorn/-/acorn-4.0.13.tgz",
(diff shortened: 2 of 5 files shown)


runfalk:
I'm not sure why Travis fails on Node v8. That's what I've tested with locally. Could it be related to cached package versions? What's the stance on committing the `package-lock.json` to avoid works on my machine issues?

It seems that it ran on Node v8 on Appveyor with no issues.

lektor/admin/static/js/widgets/multiWidgets.jsx
```diff
@@ -55,7 +55,7 @@ const CheckboxesInputWidget = React.createClass({
     let {className, value, placeholder, type, ...otherProps} = this.props  // eslint-disable-line no-unused-vars
     className = (className || '') + ' checkbox'
 
-    const choices = this.props.type.choices.map(function (item) {
```
nixjdm, line 58: This actually needs to stay using a normal function. Just changing it to an arrow function will result in the `this` inside the function being undefined. See https://github.com/lektor/lektor/pull/427 and the conversation on that pull request.

nixjdm (dismissed):

nixjdm:
Thanks @runfalk! It looks pretty good, but I want to take a little time to review it carefully and probably test it out manually before merging, so please be a little patient. I don't have a well developed opinion on the package-lock file, so I'll do some reading. The only bad part about it that I'm aware of is excess commits / noise it might cause, which probably isn't a huge deal. React isn't my forte. On that note, @vanadium23 or @singingwolfboy I'd appreciate another set of eyes as well.

lektor/admin/static/js/widgets/multiWidgets.jsx
runfalk: Before I changed this to an arrow function I got errors about `this` being undefined. After changing it to an arrow function, my model worked as intended. Are you certain this particular case needs to be a normal anonymous function?

This was tested in firefox stable and nightly.

runfalk:
If necessary I can put the commits into separate pull requests, but since both commit 1 and 3 are so small I figured I'd just piggy back them into this one.

lektor/admin/static/js/widgets/multiWidgets.jsx
runfalk: I did some more reading on [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) and the important thing here is that they'll not bind `this`, meaning they are free to use the outer scope's `this`. Since both `this.isActive` and `this.onChange` is part of the outer scope this should/must be an arrow function.

nixjdm: It seems I 'fixed' a false positive in #427. This function can be an arrow function (I'll remove the change request). As an arrow function it works fine for me, which you're right should be expected since there is a functional this in the outer scope. Though, I'm not getting an additional 'undefined' in app.js as a normal function (as it was even in the initial commit), and I would expect that to still work too. I'm curious about that. Where are you seeing the error?

nixjdm:
@runfalk You don't have to split the commits up. Thanks for the offer though.

lektor/admin/static/js/widgets/multiWidgets.jsx
runfalk: I haven't isolated it exactly, but it might be related to a list of checkboxes. For this particular case I've made a plugin that reads categories from a databag and displays the contents as checkboxes, but I assume any collection of checkboxes will trigger it.

nixjdm:
This fixes #454. I'm almost done reviewing. Hopefully we can pull this soon.

nixjdm:
@runfalk Thank you for your work and your patience! I had to go slow to keep looking things up and reading, but it makes sense and looks like some good cleanup, and it works. lgtm!

nixjdm:
@runfalk And by the way, don't worry about the JS environment of the desktop app right now. It's got it's own headaches https://github.com/lektor/lektor/issues/420

runfalk:
Thank you for merging this!

PR #468, opened 2017-10-20
Add additional markdown plugin hook

eigengrau:
In order to instantiate additional mistune lexers, plugins require
access to the Renderer instance. This commit adds an additional plugin
hook after instantiating the Renderer, but prior to creating the
Markdown processor.

The new hook is of use to plugins like lektor-shortcodes, which seek to
add new lexing rules to the Markdown syntax. Currently, such plugins
have to resort to a workaround, requiring users to pipe Lektor fields
through Jinja filters within HTML templates in order to process the
augmented syntax. By adding an additional plugin hook, Markdown syntax
additions can be processed without changes to Jinja templates.

### DIFF
--- a/lektor/markdown.py
+++ b/lektor/markdown.py
@@ -59,6 +59,7 @@ def make_markdown(env):
     cfg = MarkdownConfig()
     env.plugin_controller.emit('markdown-config', config=cfg)
     renderer = cfg.make_renderer()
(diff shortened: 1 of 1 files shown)


nixjdm:
Thanks @eigengrau. And fyi if you'd like to update any docs about this and related markdown hooks, please take a look at https://github.com/lektor/lektor-website/issues/163

PR #470, opened 2017-10-24
PR: Use primary alternative in markdown image links

### LINKED ISSUE #313: Attached image and alternatives

```
my_page
  contents.lr
  contents+es.lr
  my_image.png
```

    Alt text

```py
```

rlaverde:
Hi, this PR tries to fix  #313 

Personally I don't like the import in the function, but putting it at the top of the file cause a circular import, this could be fixed moving the constants out on `environment.py` to another file (maybe `config.py`), is that ok?

https://github.com/lektor/lektor/blob/master/lektor/environment.py#L27-L75

This is also missing a test

EDIT: Resolving the url doesn't seem like the right solution, I'll further investigate this issue

### DIFF
--- a/lektor/markdown.py
+++ b/lektor/markdown.py
@@ -27,10 +27,12 @@ class ImprovedRenderer(mistune.Renderer):
         return '<a href="%s" title="%s">%s</a>' % (link, title, text)
 
     def image(self, src, title, text):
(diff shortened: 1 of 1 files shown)


ikus060:
Bump. Any reason why this patch is not merged into Lektor ?

PR #475, opened 2017-11-02
Disable broken mistune html parsing

daradib:
Cherry-pick 2ce0b54 from master to work around lepture/mistune#137


nixjdm:
Sorry for the late handling, but his is no longer needed with 3.1 released. Thanks

PR #501, opened 2018-02-21
fix jpeg dimensions detection. closes #500

### LINKED ISSUE #500: thumbnails think they are the wrong size
Trying to do something along the lines of:
```
  {% set thumb = image.thumbnail(480) %}
  <a href="..." data-size="{{ thumb.width }}x{{ thumb.height }}">...</a>
```
Results in a wrong perceived thumb height (0 or 1) for many jpeg images. I tracked the cause to `get_image_info()` which considers all the markers in the range `ffc0` - `ffcf` to be SOF markers, which is not true.

Here is an example of such an image: https://pasteboard.co/H8HvXRn.jpg

xlotlu:
I got the SOF marker values from http://lad.dsc.ufcg.edu.br/multimidia/jpegmarker.pdf
I also found a ruby implementation using the same values:  https://github.com/sstephenson/dimensions/blob/master/lib/dimensions/jpeg_scanner.rb

### DIFF
--- a/lektor/imagetools.py
+++ b/lektor/imagetools.py
@@ -282,11 +282,23 @@ def get_image_info(fp):
     elif fmt == 'gif':
         width, height = struct.unpack('<HH', head[6:10])
     elif fmt == 'jpeg':
+        # note: some of the markers in the range
+        # ffc0 - ffcf are not SOF markers
+        SOF_MARKERS = (
+            # nondifferential Hufmann-coding frames
+            0xc0, 0xc1, 0xc2, 0xc3,
+            # differential Hufmann-coding frames
+            0xc5, 0xc6, 0xc7,
+            # nondifferential arithmetic-coding frames
+            0xc9, 0xca, 0xcb,
+            # differential arithmetic-coding frames
+            0xcd, 0xce, 0xcf,
+        )
         try:
             fp.seek(0)
             size = 2
             ftype = 0
-            while not 0xc0 <= ftype <= 0xcf:
+            while ftype not in SOF_MARKERS:
                 fp.seek(size, 1)
                 byte = fp.read(1)
                 while ord(byte) == 0xff:


nixjdm:
I've tried to compare your test file with the one at https://github.com/lektor/lektor/blob/master/tests/demo-project/content/test.jpg in https://www.onlinehexeditor.com/ (using the 'load' .jpg button on the bottom).

They both have examples of 0xc4 and 0xc0, and the test jpg in lektor's code also has a 0xcf.

Since both have a 0xc4, which is affect by this code change, perhaps that is a necessary but not sufficient condition for this error. Since that doesn't completely describe why the error occurred, this may completely solve the bug or not, I'm not sure. We should add a test for a file that passes only with this code change.

With an appropriate image, that ought to be as easy / tedious as copying existing code in test_images.py, and adding to demo-project/templates/page.html and changing some of the image details to match.

xlotlu:
See how in your test image the old code's first match is 0xc0 at position 0x18ec. The non-SOF 0xc4 is just a few bytes later at 0x18ff. It didn't trip on it simply because they're in the "lucky order".

With my image i get:
```
pos 0x2d92 , marker 0xc4 (bad) 
pos 0x2f36 , marker 0xc0 (good)  
```

xlotlu:
This bugged me, so i tested some more. The full list of markers according to the alorightm is:
```
test.jpg:
0x18ec 0xc0
0x18ff 0xc4
0x1920 0xc4
0x19d7 0xc4
0x19f8 0xc4
0x5a28 0xc6
0x1235c 0xce
```
```
H8HvXRn.jpg:
0x2d92 0xc4
0x2f36 0xc0
0x2d759 0xc5
0x96fed 0xc1
0xcf3cc 0xcb
0x1e37f8 0xc2
```

The thing is, only the c0s and c4s are real markers, because the rest aren't prepended with ff. The way the code is written now it just skips over any number of 0xff found until it reaches a marker, but it doesn't make sure the previous byte was indeed a 0xff. This is probably a bug waiting to happen, because the jpeg metadata could be anywhere in the file.

It's a bit late (in the night) to try to fix this now, and starting tomorrow I'll be offline for a couple of weeks. @nixjdm my suggestion is to merge this commit for now, and create a new issue to fix the algorithm. I'll happily take care of it when I'll be back.

Issue #509, opened 2018-03-24
Fix SOF marker identification for images


nixjdm:
See https://github.com/lektor/lektor/pull/501#issuecomment-373563060 by @xlotlu 

There's more work that should be done to ensure images are processed correctly. Additional / better testing  should be done as well.

PR #514, opened 2018-04-05
improve jpg info detection. addresses #509

xlotlu:
This is a bit more complicated than initially expected. The specs are at  https://www.w3.org/Graphics/JPEG/itu-t81.pdf -- Figure B.2 (page 34  / 38) shows the whole structure. There is also a "hierarchical mode" (figure B.13), but the code hits that as well.

It goes like this:
- there are standalone markers, with no extra-information;
- all the other markers carry some form of "payload". they specify the payload's length, so one can skip ahead without reading byte-by-byte;
- a "frame" (the image data), has a frame header introduced by the SOF marker, but it can be prepended with other info (e.g. a Huffman table, introduced by a DHT marker (this was breaking the old code));
- the frame header contains info about the width and height (number of lines). except the lines can be 0;
- a frame consists of a number of "scans". if the lines in the frame header was 0, the lines will be provided later in a special segment;
- this segment is the DNL segment, and it must occur after the first scan;
- a scan itself has a header (with a marker), and can also be prepended with other info;
- a scan is made up of one or more "entropy-coded segments", that is, binary image data, separated by standalone markers;
- as far as i can tell, one cannot know in advance the length of an ECS or of a scan, so, if the frame header didn't provide the lines number, we must plough ahead through all the binary data until we hit another (the DNL) marker.

Please do review the code, but don't merge yet. I would like to create a few test images:
- one with SOF first, like the current one,
- one with DHT followed by SOF,
- one with the crazy frame header with lines = 0, followed by the DNL.

(I'm not sure yet how to do this. I guess I'll need to do some manual hex editing. oh well. :) )

One last thing: the code could be improved to make sure the markers are in the proper order, and to bail out early if the image is non-conformant. For now I have decided against this however, because:
- I'm really not sure how conformant jpeg encoders are,
- I'm really not sure how lenient jpeg decoders generally are (I've read somewhere that they have to be, because for example both the EXIF header and the JFIF header want to be first after the start-of-image marker, but in practice cameras create both headers, so one of them will be "against the spec"),
- it would complicate the code,
- I won't be 100% sure I got it right,
- I'm lazy.


nixjdm:
I've read through this all, and I *think* I get it. I can follow all of the steps, I think, and I haven't spotted any issues yet. The spec is a little convoluted. That's not your fault though. It's a bit hard to wrap my head around.

I haven't read though the relevant section in the pdf yet. Assuming this actually matches the pdf, this looks like great improvement. I might have the time to attempt that over the weekend.

nixjdm:
@xlotlu

My only question so far is about 0xc8 for JPG. Is that in the right section? Table B.1 lists it in "Start Of Frame markers, non-differential, arithmetic coding", which I would think would make us want to treat it as an SOF, and thus put it in `_JPEG_SOF_MARKERS` instead of `_JPEG_OTHER_MARKERS`. Why is this not the case?

Filling in something for that last `#raise Exception("malformed file")` seems like a good idea :)

I *have* looked at that https://www.w3.org/Graphics/JPEG/itu-t81.pdf, but I can't claim a great understanding. I can't spot any other problems at the moment though. If you provide some passing tests I'll be pretty convinced it works and happily merge. I really appreciate the liberal comments for this section.

xlotlu:
> My only question so far is about 0xc8 for JPG. Is that in the right section? Table B.1 lists it in "Start Of Frame markers, non-differential, arithmetic coding", which I would think would make us want to treat it as an SOF, and thus put it in _JPEG_SOF_MARKERS instead of _JPEG_OTHER_MARKERS. Why is this not the case?

I don't know what the rationale of that marker is, or why it's included in that section, but it's not a SOF marker: its symbol is "JPG", and its description "Reserved for JPEG extensions". Also, every implementation I've seen skips it, see for example [PIL/JpegImagePlugin.py](https://github.com/python-pillow/Pillow/blob/d173e8179824e7f24ad6aa482a7986a7f7f0272a/src/PIL/JpegImagePlugin.py#L234).

> Filling in something for that last #raise Exception("malformed file") seems like a good idea :)

I'm not really sure what should happen there. I'm open to suggestions.

xlotlu:
@nixjdm I gave up handling DNL because nobody really cares about it. See [libjpeg-turbo/jdmarker.c#L267](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/1095b822b07a484c70fd40078f3e39485585a80d/jdmarker.c#L267) and [the rationale](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/da2a27ef056a0179cbd80f9146e58b89403d9933/structure.txt#L855). The code and the blurb are inherited from [upstream](http://www.ijg.org/files/), which imagemagick also uses.

I also got rid of all the extra-header handling, so this is pretty much what was in the beginning, but without the bug, and with early bailout. I could have tested only the expected markers to be encountered, but... I calls this good enough.

The only question left is whether you want those exceptions there or a quiet exit. The initial code was a totally quiet try: ... except: return.

nixjdm:
@xlotlu once again thanks for the links and explanation. That definitely helped get broader context to review this.

I think this PR should stay the way it is in regards to the exceptions. I prefer the idea that if the image is bad, we be pretty clear about that. Throwing an exception and failing building that page I think is much more helpful than silently failing.

Thank you for this! I know it was some effort!

xlotlu:
:) @nixjdm and thank you for your work.

> Throwing an exception and failing building that page I think is much more helpful than silently failing.

I agree,  it's just that now it's inconsistent with the gif/png above. :) (and maybe there should be a nicer, custom exception? design decision.)

P.S.: this stuff so belongs in a stdlib package to supersede imghdr.

nixjdm:
There is a custom `LektorException` that is used a couple places. For example https://github.com/lektor/lektor/blob/05a541120e6f6add0d5d9d30d0585166f838594c/lektor/build_programs.py#L178. Throwing an exception for bad pngs and gifs would be better that what's going on now.

I totally agree, this should be in an external package. If you have the time and gumption for a side project and do that, I'll gladly import it!

PR #515, opened 2018-04-05
Protect against building in existing nonempty dir

### LINKED ISSUE #513: Add a protection around 'build' to make it less likely to delete unrelated data
I suggest adding logic on the existence on the `.lektor` dir in the build dir. Then, whenever `prune` is run (as in right after `build`), prune will make a quick check for the existence of this dir, and if it does not exist and if the dir isn't empty, ask for user confirmation to proceed in deleting untracked files. This could be overridden with a flag preemptively.

The help for `build` *does* warn that `prune` will be ran next, but I don't feel that it does enough to make it hard to accidentally delete things it shouldn't. I remember once I heard of someone building to their Desktop, which shouldn't be done, and that caused some headache to say the least.  Without thinking too hard first, I was too loose with my fingers and just did it (to a place that didn't matter, but still). It's a common enough mistake and potentially really destructive, so we should add in a check to prevent it.

nixjdm:
Fixes #513 

I decided to put the warning before the build even happens, as it can be potentially problematic by itself. I also think it's so unlikely to be the case someone really wants to do this that I didn't feel the need to add in an ability to bypass the check with a flag. It would mainly be clutter in the cli.

The prompt is tripped when `.lektor` needs to be made (i.e. this dir hasn't been used before as a build dir) and is nonempty.

With tests.

### DIFF
--- a/lektor/builder.py
+++ b/lektor/builder.py
@@ -6,6 +6,7 @@ import shutil
 import sqlite3
 import hashlib
 import tempfile
+import click
 
 from contextlib import contextmanager
 from itertools import chain
@@ -964,6 +965,15 @@ class Builder(object):
 
         try:
             os.makedirs(self.meta_path)
+            if os.listdir(self.destination_path) != ['.lektor']:
+                if not click.confirm(click.style(
+                        "The build dir %s hasn't been used before, and other "
+                        "files or folders already exist there. If you prune "
+                        "(which normally follows the build step), "
+                        "they will be deleted. Proceed with building?"
+                        % self.destination_path, fg='yellow')):
+                    os.rmdir(self.meta_path)
+                    raise click.Abort()
         except OSError:
             pass
 
--- a/lektor/builder.py
+++ b/lektor/builder.py
@@ -6,12 +6,12 @@ import shutil
 import sqlite3
 import hashlib
 import tempfile
-import click
 
 from contextlib import contextmanager
 from itertools import chain
 from collections import deque, namedtuple
 
+import click
 from werkzeug.posixemulation import rename
 
 from lektor._compat import PY2, iteritems, text_type
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -5,13 +5,19 @@ import os
 from lektor.cli import cli
 
 def test_build_abort_in_existing_nonempty_dir(project_cli_runner):
-    result = project_cli_runner.invoke(cli, ["build", "-O", os.getcwd()], input='n\n')
+    os.mkdir('build_dir')
+    with open('build_dir/test', 'w'):
+        pass
+    result = project_cli_runner.invoke(cli, ["build", "-O", "build_dir"], input='n\n')
     assert "Aborted!" in result.output
     assert result.exit_code == 1
 
 
 def test_build_continue_in_existing_nonempty_dir(project_cli_runner):
-    result = project_cli_runner.invoke(cli, ["build", "-O", os.getcwd()], input='y\n')
+    os.mkdir('build_dir')
+    with open('build_dir/test', 'w'):
+        pass
+    result = project_cli_runner.invoke(cli, ["build", "-O", "build_dir"], input='y\n')
     assert "Finished prune" in result.output
     assert result.exit_code == 0
 
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,7 +1,18 @@
 import re
 import warnings
+import os
+
 from lektor.cli import cli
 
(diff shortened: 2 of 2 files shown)


nixjdm:
I'm not sure what's going on with the [Windows Appveyor test failures](https://ci.appveyor.com/project/Lektor/lektor/build/1.0.150). They seem to be failing in the same spot, but only about 1/4 of the time. Our testing is repetitive enough for that to be caught. I don't know why they're failing. Do I need to specify a different path for the constructed `build_dir`? Any Windows users able to test this themselves? Another clue perhaps is that these three all seem to fail together. I don't think any of the test failures on Appveyor is a subset of these three. It's either all three or no failures.

The first failure is an exit status of 1, but coming after `assert "Finished prune" in result.output` passes, which is confusing. Similar story for the second failure.

For the third, [`builder.build_all()`](https://github.com/lektor/lektor/blob/master/tests/test_images.py#L68) seems to work most of the time, and hasn't failed on Travis. Could this be a hangup on the confirmation? Maybe it took too long?

nixjdm:
Thanks @Andrew-Shay!

After activating your virtual env, and in the lektor repo dir (and on the lektor-513 branch),
```
pip install -e ".[test]"
```
and then manually running a small variant of [this line in the Makefile](https://github.com/lektor/lektor/blob/master/Makefile#L22), run all python tests with:
```
py.test . --tb=short -vs --cov=lektor
```
or substitute the `.` for a single test module (like test_cli.py) to run just that. I typically add the `s` flag to turn on print statements in the tests for easier debugging. FWIW, appveyor runs the tests [slightly differently](https://ci.appveyor.com/project/Lektor/lektor/build/1.0.150/job/f29s81f90hs2xljc#L1277), but I think this will work for you.

If you figure it out, feel free to submit a PR against the lektor-513 branch :) Let me know.

Andrew-Shay:
It is probably just imagemagick not being on appveyor?
Before installing imagemagick and checking "add to system path" in installer:
`Lektor: 187 total, 4 failed, 1 ignored, 182 passed`
I had the same failed tests as appveyor

After:
`Lektor: 187 total, 1 failed, 186 passed`

```
=================================== ERRORS ====================================
_________________ ERROR at setup of test_default_author_email _________________
tests\conftest.py:169: in git_user_email
    stderr=subprocess.PIPE).communicate()[0].strip()
c:\program files\python36\Lib\subprocess.py:709: in __init__
    restore_signals, start_new_session)
c:\program files\python36\Lib\subprocess.py:997: in _execute_child
    startupinfo)
I:\Program Files\JetBrains\PyCharm Community Edition 2017.3.2\helpers\pydev\_pydev_bundle\pydev_monkey.py:420: in new_CreateProcess
    return getattr(_subprocess, original_name)(app_name, patch_arg_str_win(cmd_line), *args)
E   FileNotFoundError: [WinError 2] The system cannot find the file specified
==================== 186 passed, 1 error in 189.37 seconds ====================
```

Andrew-Shay:
Also, git isn't configured where I'm running the tests, so the failure makes sense.

nixjdm:
> Before installing imagemagick and checking "add to system path" in installer:
Lektor: 187 total, 4 failed, 1 ignored, 182 passed
I had the same failed tests as appveyor

Huh, well, that's interesting. I wonder why that's a problem all of a sudden. I think you might be right though, I didn't see it in the log before, but check out [this line](https://ci.appveyor.com/project/Lektor/lektor/build/1.0.146/job/ftcipn8w2fsbt96u#L50). Definitely seems like that's a problem! There's even a [long timeout for the imagemagick install](https://github.com/lektor/lektor/blob/05a541120e6f6add0d5d9d30d0585166f838594c/appveyor.yml#L34) - like known to be flakey. That may be a part of the reason the appveyor builds are slow.

So now, what do we do about it? Attempt to download it, test for it, and try again if it failed? Maybe we can cache it too.

Thanks for digging into it @Andrew-Shay!

Andrew-Shay:
Ah good catch. That choco install is probably the problem
I've never used appveyor, can the imagemagick binary be manually added to the environment to ensure it always exists?  
Or have it pull the portable version and extract it?
https://www.imagemagick.org/script/download.php  
`ImageMagick-7.0.7-28-portable-Q16-x64.zip`

nixjdm:
Thanks for your help @Andrew-Shay. In #518 I added caching for imagemagick, which won't *guarantee* it the download succeeds, but it ought to make it a lot more likely. I'm hoping that'll be the end of that!

With that, the tests are passing, and I don't see anything in the way of this, so I'm pulling it.

PR #521, opened 2018-04-15
Detect format when computing height if none has been passed (fixes #520)

### LINKED ISSUE #520: Restore compatibility with lektor-thumbnail-generator
The plugin https://github.com/skorokithakis/lektor-thumbnail-generator currently doesn't work with 3.1 / master, because of [this commit](https://github.com/lektor/lektor/commit/b867261f27426670a250ea3b5e77b80152f9f967#diff-6189d81443cffb507c815d0c2ec6e783) that introduced another param, `format`,  on `computed_height`.

This addition isn't needed. We could extract the format in other ways from existing code in imagetools.py. For instance, possibly from `get_image_info`. If we do that then we also will restore compatibility with the thumbnail generator plugin.

skorokithakis:
I have reorganized the arguments of the `computed_height` function to make it backwards-compatible and I have added detecting the format if no format has been passed. This fixes compatibility with `lektor-thumbnail-generator`.

### DIFF
--- a/lektor/imagetools.py
+++ b/lektor/imagetools.py
@@ -363,11 +363,17 @@ def get_quality(source_filename):
     return 85
 
 
-def computed_height(source_image, format, width, actual_width, actual_height):
+def computed_height(source_image, width, actual_width, actual_height, format=None):
     # If the file is a JPEG file and the EXIF header shows that the image
     # is rotated, imagemagick will auto-orient the file. That is, a 400x200
     # file where the target width is 100px will *not* be converted to a
     # 100x50 image, but to 100x200.
+
+    # Detect the format, as none has been passed.
+    if format is None:
+        with open(source_image, 'rb') as f:
+            format = get_image_info(f)[0]
+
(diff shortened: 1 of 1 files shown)


skorokithakis:
I forgot to cc @nixjdm.

nixjdm:
Hey, that was fast! Thanks!

I have a little mixed feelings about leaving the `format` param in at all, since I think it probably shouldn't have been added in the first place. @skorokithakis it looks like your plugin is the only one on https://www.getlektor.com/docs/plugins/list/ that actually does something explicit with images, so there probably isn't much risk in just removing it.

Is there a use case you can think of for it? It seems completely unnecessary to me. If it is, and nothing's using it, I think we should just take it back out.

skorokithakis:
It seems unnecessary to me too, except for the minor optimization of not having to read the image twice. Then again, that's a pretty fast operation.

nixjdm:
Thanks guys!

PR #524, opened 2018-04-20
run with "python -m lektor"

### LINKED ISSUE #523: run with "python -m lektor"
I'd like to run Lektor from within PyCharm in order to take advantage of its debugger during plugin development. PyCharm has the ability to run Python modules, `python -m lektor`, but Lektor isn't set up for this, so I have to point at the script generated by the setuptools entry point instead, `/home/david/.virtualenvs/lektor/bin/lektor`.

davidism:
closes #523 

Adds a `__main__.py` file so that `python -m lektor` works. Instead of aliasing `main = cli`, main is now a function that passes the correct args and name to the cli when running as a module. This is all taken directly from Flask.

I made this against the 3.x-maintenance branch, but I'm not sure about that since it doesn't seem to be synced with the 3.1 release. I can resubmit this against master if you're not using the maintenance branches.

### DIFF
--- a/lektor/__main__.py
+++ b/lektor/__main__.py
@@ -1,0 +1,3 @@
+if __name__ == '__main__':
+    from .cli import main
+    main(as_module=True)
--- a/lektor/cli.py
+++ b/lektor/cli.py
@@ -616,4 +616,16 @@ from .devcli import cli as devcli  # pylint: disable=wrong-import-position
 cli.add_command(devcli, 'dev')
 
 
-main = cli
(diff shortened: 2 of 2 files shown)


nixjdm:
PR looks good to me. We are not using the 3.x-maintenance branch right now, so yes, please switch it to be against master an it'll get right into the next release.

Thanks!

davidism:
This should merge into master now. Waiting for AppVeyor to finish.

PR #528, opened 2018-04-28
Added tests for the `lektor dev new-plugin` command.

nixjdm:
Here are some new tests for the CLI command `lektor dev new-plugin`.

These tests pass locally, but I've keep running into the same errors on Appveyor with Python 2, and Travis with Python 3. Appveyor with Python 3 passes, as does Travis with Python 2. I don't know what the cause is of the errors on those platforms or how to prevent them yet.

Easily see the difference: [Travis](https://travis-ci.org/lektor/lektor/builds/373597166), and [Appveyor](https://ci.appveyor.com/project/Lektor/lektor/build/1.0.238).

I have tried to reproduce the errors locally with Python2 and Python3, and with live debugging on both Appveyor (rdp + powershell) and Travis (ssh), and haven't succeeded.

For some reason, the part of each test that runs 
```python
result = project_cli_runner.invoke(cli, ["dev", "new-plugin"],
                                   input='Plugin Name\n\nAuthor Name\nauthor@email.com\ny\n'
)
```
does not execute as expected, giving `result.output` equal to
```
Step 1:

```
instead of the larger, complete output, which varies a bit per test, but in general looks like
```
Step 1:
| This is the human readable name for this plugin
> Plugin Name: Plugin Name

Step 2:
| The place where you want to initialize the plugin
> Plugin Path [/tmp/tmp6xn_rxhh/packages/plugin-name]: 

Step 3:
| Your name as it will be embedded in the plugin metadata.
> Author Name [joe,,,]: Author Name

Step 4:
| Your e-mail address for the plugin info.
> Author E-Mail [nixjdm@terminallabs.com]: author@email.com

Create Plugin? [Y/n] y
```

I'll keep poking around, but I don't want to pull these tests if they only work locally. Any help is appreciated. I'm running out of ideas atm.

--
The simplest test that ought to fail would be:
```python
import os

from lektor.cli import cli

def test_new_plugin(project_cli_runner):
    result = project_cli_runner.invoke(cli, ["dev", "new-plugin"],
                                       input='Plugin Name\n\nAuthor Name\nauthor@email.com\ny\n'
    )
    assert "Create Plugin?" in result.output
    assert result.exit_code == 0
```

-----
**Update:**

Thanks to @anlutro's help on freenode, we I now know what's causing this to some extant. [Lektor was subtracting 2 from the terminal width](https://github.com/lektor/lektor/blob/1e4b8b2b8ed8f1c5a436ad0b243f6eaae7ec2c75/lektor/quickstart.py#L50), and Click was erring on a width of -2 since the detected width was 0. This was again the case on Py3 Travis and Py2 Appveyor, and not Py2 Travis and Py3 Appveyor, for reasons I have not dug into. Where the error exists though, that's what's going on. I'll update the PR later to fix that as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.egg-info
 *.pex
 .cache
+#*
 build
 dist
 venv
@@ -14,6 +15,7 @@ coverage/
 coverage.xml
 coverage.lcov
 .nyc_output
+.pytest_cache
 
 node_modules
 gui/build
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ pex:
 
 test-python:
 	@echo "---> running python tests"
-	py.test . --tb=short -v --cov=lektor
+	py.test . --tb=long -svv --cov=lektor
 
 coverage-python: test-python
 	coverage xml
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -68,3 +68,6 @@ test_script:
   - npm list
   - npm run lint
   - npm test
+
+on_failure:
+  - ps: $blockRdp = $true; iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -60,7 +60,7 @@ test_script:
   # Run Python tests
   - "%PYTHON%\\python.exe --version"
   - "%PYTHON%\\python.exe -m pip list"
-  - "%PYTHON%\\Scripts\\pytest.exe . --tb=short -v --cov=lektor"
+  - "%PYTHON%\\Scripts\\pytest.exe . --tb=long -svv --cov=lektor"
   # Run JS tests
   - cd lektor\admin
   - node --version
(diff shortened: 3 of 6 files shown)


nixjdm:
@lektor/developers After much hullabaloo figuring out the term_width issue, this PR is ready to go. Tests are currently passing on Travis and Appveyor. This also fixes #529.

The problem was caused by a detected terminal width of `0` in certain environments, and subtracting `2` from it in [our code](https://github.com/lektor/lektor/blob/a1139cd32e52d496c9a86d4183b38847b42a56f9/lektor/quickstart.py#L50). The `-2` has the effect of making sure there are at least 2 empty spaces on the end of each line. The text wrapping for prompts needs a terminal width greater than `0`.

There were a couple potential ways to fix the term_width issue. I choose to remove the extra padding on prompts, and to set a minimum width of `1`. I'm not strongly tied to this approach, but I felt like it was not consistent with other echos throughout Lektor that do not do this, and it changes the behavior of output to something not normal, which helped cause this problem. An alternative approach is to keep the padding and change the minimum effective terminal width to `3`.

PR #533, opened 2018-05-07
New plugin updates

nixjdm:
Corresponding to the changes in https://github.com/lektor/lektor-website/pull/194, this updates the default setup.py we give people when they create a new plugin. It will default to having a valid long_description and other useful fields.


goanpeca (approved):
:+1:

PR #549, opened 2018-06-11
fix test image rotation

xlotlu:
the original test image had an exif rotation, but the newly-created progressive one didn't inherit the exif data.

this fixes that by rotating the image.


xlotlu:
oops, forgot about that test. closing this, will submit the updated jpg along with the big changes.

Issue #526, opened 2018-04-26
Add tests for url processing


nixjdm:
https://github.com/lektor/lektor/blob/master/tests/test_urls.py could use better test coverage. `url_to` accepts `path, alt=None, absolute=None, external=None, base_url=None`, and only some of that is tested. See https://github.com/lektor/lektor/blob/05a541120e6f6add0d5d9d30d0585166f838594c/lektor/sourceobj.py#L88-L89

nixjdm:
Fixed in #552

PR #552, opened 2018-06-13
test added under tests/test_urls.py about SourceObject.url_to()

mickaelReuze:
Issue #526 

Hello,

adding tests for all `url_to()` parameters.

Can you confirm that `expected` parameter is always good.
I have a doubt for some tests with `absolute = True`, as `external` and `base_url` are not took into account. (see below)
```
@pytest.mark.parametrize("path, alt, absolute, external, base_url, expected", [
("/projects/slave/", "de", True, True, None,  '/de/projects/sklave/'),
("/projects/slave/", "de", True, True, '/content/',  '/de/projects/sklave/'),
("/projects/slave/", "de", None, True, '/content/',  '/projects/slave1/de/projects/sklave/'),
("/projects/slave/", None, True, True, None,  '/projects/slave/'),
("/projects/slave/", None, True, True, '/content/',  '/projects/slave/'),
("/projects/slave/", None, True, None, '/content/',  '/projects/slave/')
])
```
It's clearly written under `url_to()` to return `path` before `absolute` or `base_url` are processed: 
        ```if absolute:
            return path```

Also I have to set `pad.db.config.base_url` when `external = True`, is that an expected behavior ?
```
    if external :
        pad.db.config.base_url = "/projects/slave1/"
```
Regards.

### DIFF
--- a/tests/test_urls.py
+++ b/tests/test_urls.py
@@ -27,26 +27,51 @@ def test_basic_url_to_with_alts(pad):
     assert slave_de.url_to(slave_en) == '../../../projects/slave/'
 
 
-@pytest.mark.parametrize("alt, absolute, external, base_url, expected", [
-    ("de", None, None, None, '../../de/projects/sklave/'),
-    ("de", True, None, None, '/de/projects/sklave/'),
-    ("de", True, True, None, '/de/projects/sklave/'), #
-    ("de", True, True, '/content/', '/de/projects/sklave/'),#
-    ("de", None, True, None, '/projects/slave1/de/projects/sklave/'),
-    ("de", None, True, '/content/', '/projects/slave1/de/projects/sklave/'),#
-    ("de", None, None, '/content/', '../de/projects/sklave/'),
-    (None, True, None, None, '/projects/slave/'),
-    (None, True, True, None, '/projects/slave/'),#
-    (None, True, True, '/content/', '/projects/slave/'),
-    (None, True, None, '/content/', '/projects/slave/'),
-    (None, None, True, None, '/projects/slave1/projects/slave/'),
-    (None, None, True, '/content/', '/projects/slave1/projects/slave/'),
-    (None, None, None, '/content/', '../projects/slave/'),
+@pytest.mark.parametrize('path, alt, absolute, external, base_url, expected', [
+    ('/projects/slave', '', None, None, None, '../../projects/slave/'),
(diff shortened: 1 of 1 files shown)


nixjdm:
> Also I have to set pad.db.config.base_url when external = True, is that an expected behavior ?

It is, though it is pretty confusing right now and should be cleaned up. This `url_to` ends up calling `make_url` in db.py, which is also used elsewhere. The `base_url` when set, is sent straight to https://github.com/lektor/lektor/blob/master/lektor/db.py#L1444 to be used to generate relative urls, it is the "from" url and looks like `path/of/page`. It is not the same as the `config.base_url` used to generate the external url, this one is more the site root url when they are external and looks like `example.com`.

This is much better test coverage. It doesn't cover every permutation, but this is definitely an improvement. A couple things not included: the reflection of going back from de pages to en pages, `wolf_en.url_to("/projects/slave/", None, None, None, None)`, [paths starting with a bang](https://github.com/lektor/lektor/blob/05a541120e6f6add0d5d9d30d0585166f838594c/lektor/sourceobj.py#L105), and maybe others. It's not super obvious at a glance what the scheme for going through the options is so it's a little hard to pick out exactly what isn't covered.

Eventually (after some cleanup optionally) we should cover the very similar tests of `make_url` directly, to for instance test this [line](https://github.com/lektor/lektor/blob/master/lektor/db.py#L1436), which isn't testable through https://github.com/lektor/lektor/blob/05a541120e6f6add0d5d9d30d0585166f838594c/lektor/sourceobj.py#L88

but that doesn't need covered here, though it's certainly welcome.

Could you try to cover a few more cases, consistently use single quotes, and get rid of those extra hash marks? After that I'll pull this.

mickaelReuze:
Hi,

Added some tests:
 - When a path start with "!"
 - When a path is empty ""
 - When `alt` value is not defined
 
Also added some Error Cases to trigger the RunTimeError you pointed out [here](https://github.com/lektor/lektor/blob/master/lektor/db.py#L1436).
_(As far as I know I can trigger it when `absolute` is False, `external` is True and there is nothing under `config.base_url`)_

All tests are PASSED

nixjdm:
Thank you very much. This is much more comprehensive.

That RuntimeError message is a little example of why this needs cleaning up. It's confusing that 'absolute' in the message is the more colloquial 'absolute', but it really what the rest of Lektor terms 'external', not Lektor's 'absolute'.

Issue #554, opened 2018-06-13
ftp deploy on python 3 broken


nixjdm:
From @uwekamper on gitter:

> This line causes an error for me: https://github.com/lektor/lektor/blob/05a541120e6f6add0d5d9d30d0585166f838594c/lektor/publisher.py#L280
when the self.con.login() is called, I get an error message saying "must be str, not bytes"

It looks like ftp doesn't work on python 3.

uwekamper:
My immediate work-around for this bug was to create simple upload script:

    #!/bin/bash
    ( cd $(lektor project-info --output-path) ; \
        find . -type f -exec curl -u <username>:<password> --ftp-create-dirs -T {} ftp://<ftp-server- hostname>/{} \; )

nixjdm:
This was fixed when #555 was merged, and didn't autoclose.

Issue #553, opened 2018-06-13
Add tests for ftp deploy


nixjdm:
We got a report that ftp may not work with python 3. Also there are [a couple other issues and an open pull reequest](https://github.com/lektor/lektor/issues?utf8=%E2%9C%93&q=is%3Aopen+ftp+) that would benefit from having a testing harness for ftp deploys.

PR #555, opened 2018-06-14
Fix some problems with FTP in Python 3.6

uwekamper:
I was able to do some digging in the FTP upload of lektor. It now works for me when I do FTP uploads. FTP-TLS is still broken, though.

The API of pythons `ftplib` has changed slightly with Python 3. Most of the errors where caused by ftplib expecting `str` instead of `bytes` and vice versa.

Furthermore, `ftplib` uses exceptions for a lot of things. There were lots of instances where I found code like this:

     try:
          ...
     except Exception as e:
          ...

This made it a lot harder to find the root-cause of many errors. I changed that to only catch ftplib's own `Error` class (which is renamed to `FTPError` during import).

I would not know how to write unit tests for this … sorry.

Hope this helps. The corresponding issue is #554 .

### DIFF
--- a/lektor/publisher.py
+++ b/lektor/publisher.py
@@ -10,9 +10,10 @@ import threading
 from contextlib import contextmanager
 
 from werkzeug import urls
+from ftplib import Error as FTPError
 
 from lektor._compat import (iteritems, iterkeys, range_type, string_types,
-    text_type, queue, StringIO)
+    text_type, queue, BytesIO, StringIO, PY2)
 from lektor.exception import LektorException
 from lektor.utils import locate_executable, portable_popen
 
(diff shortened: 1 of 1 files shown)


nixjdm:
Thanks @uwekamper! This looks good to me! It is thankfully pretty easy to read this considering we don't have this code tested. Considering it is a bugfix, and I'm fairly confident in my read of it, I'll accept this without a test. If you are able to take a shot at writing tests for this though under #553, that would be greatly appreciated.

Fixes #554

nixjdm:
In addition to these fixes, I also appreciate you making some of those `except Exception`s more specific. That's much more useful.

PR #563, opened 2018-06-28
Adding Node 10 and Python 3.7 to tests

nixjdm:
Python 3.7 is out, so I'm adding that to the tests. Node 10 is as well, so it's in too. This can be pulled once appveyor and travis actually support these versions, which I don't think they do quite yet, and any newly uncovered bugs are fixed.

I also lowered the number of total tests by not testing every permutation. The tests now only run a minimum of 1 times per version, not 3 as before. This should be sufficient and is a little faster.


nixjdm:
And it turns out that we're not compatible with node 10.

nixjdm:
The problem with node 10 is with the `natives==1.1.0`, that is, `<1.1.3`. `natives` is brought by `graceful-fs`, which is brought by `less`. So one way or the other an upgrade to natives `>1.1.3` will allow builds again. A targeted upgrade is probably fine for this PR. Though of course the packages should be overhauled at some point.

nixjdm:
Node 10 builds now. Just waiting on Python 3.7 to land.

I also added some badges for this and updated the trove classifiers.

PR #566, opened 2018-07-04
Be more robust when getting SVG info

eigengrau:
SVGs in the wild don’t always include an XML declaration, and as per [1], they
don’t necessarily have to. Additionally, height and width declarations may include more
than just an integer.

[1]: https://www.w3.org/TR/xml/#sec-prolog-dtd

### DIFF
--- a/lektor/imagetools.py
+++ b/lektor/imagetools.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import os
 import imghdr
+import re
 import struct
 import posixpath
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


nixjdm:
This looks pretty straight forward. There are no tests for this in test_images.py. Are you up for putting a couple tests in there for this? You could test this directly by importing and using `get_image_info` directly, without any Lektor context or project set up.

You could hardcode a very simple svg with and without the XML declaration. Just remember to create the file pointer with `fp = io.BytesIO` for 2/3 compatibility.

eigengrau:
Hello Joseph! Thanks for the pointer. I amended the commit with a regression test. I also pushed another fix to make determining SVG size a bit more robust. Let me know if this works for you.

nixjdm:
Thanks @eigengrau!

PR #578, opened 2018-07-15
Add test for pagination based on attachment items

eigengrau:


nixjdm:
Have you seen that this doesn't pass the tests? I think if you rebase off of master some the build tests might pass at least.

yagebu:
This seems to have gotten quite stale, so I'm closing this PR.

PR #579, opened 2018-07-18
[WIP] Support s/e shortcuts to switch between edit / preview mode.

dwt:
Hi there,

I really liked the way it is possible to switch from the edit screen to the preview screen with `command-s`, however I also wanted to switch back to the edit screen from the preview.

Thus what I do in this pull request.

Some things that I would love some feedback:
* When switching between the views, sometimes the opposing key-event seems to be auto generated by the browser
* I tried using `command+…` shortcuts but those seem to work really strangely - do you happen to know why?
* I'd like to have tooltips on the menu buttons (edit/preview) on the left - but I don't know react very well - how would one do that?

### DIFF
--- a/lektor/admin/static/js/views/EditPage.jsx
+++ b/lektor/admin/static/js/views/EditPage.jsx
@@ -54,7 +54,7 @@ class EditPage extends RecordEditComponent {
       console.log('_onKeyPress', event)
     if (
         (event.which === 83 && utils.isMetaKey(event)) // meta+s
-        || event.which === 83 // s
+        // || event.which === 83 // s
     ) {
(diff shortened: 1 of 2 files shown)


dwt:
Well, I quickly noticed that those event handlers, also get called for every keypress _inside_ a textarea or input - so back to the meta key variants it is.

I do however see a very strange bug. That is, if you change between the two views with the `command-e`/`command-s` shortcuts, and then press just `command` - the views will switch again.

In the debugger it looks like the shortcut was pressed again - which it certainly isn't. Do you have any hint why that might be?

yagebu:
Superseded by #876

PR #580, opened 2018-07-27
enabled do extension in jinja environment

ericmjl:
@nixjdm @singingwolfboy My apologies, as this is an unsolicited PR, but I genuinely think that the `do` extension is a useful thing to have as part of the default Jinja2 environment that Lektor uses.

Please let me know if this is something you'd like or not; as this is my first PR into Lektor, I wasn't quite sure if I needed to add tests and the likes. Please do let me know, I am more than happy to continue adding things if needed.

### DIFF
--- a/lektor/environment.py
+++ b/lektor/environment.py
@@ -428,7 +428,9 @@ class Environment(object):
 
         self.jinja_env = CustomJinjaEnvironment(
             autoescape=self.select_jinja_autoescape,
-            extensions=['jinja2.ext.autoescape', 'jinja2.ext.with_'],
+            extensions=['jinja2.ext.autoescape',
+                        'jinja2.ext.with_',
+                        'jinja2.ext.do'],
             loader=jinja2.FileSystemLoader(
                 template_paths)
         )
--- a/tests/test_environment.py
+++ b/tests/test_environment.py
@@ -1,0 +1,6 @@
+def test_jinja2_extensions(env):
+    extensions = env.jinja_env.extensions
+
+    assert 'jinja2.ext.AutoEscapeExtension' in extensions.keys()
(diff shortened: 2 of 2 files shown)


nixjdm:
Thanks for the PR @ericmjl! I'm fine with adding in this extension. Could you add a test so that we know that the extension is picked up and works? The code looks, fine, it's more to make sure that it doesn't break in the future.

You can ignore all of the pylint errors that are unrelated. It looks like pylint updated and we need to fix a slew of new linting errors again.

ericmjl:
@nixjdm thank you! :smile: 

If I could ask, where would you suggest I put the test? I've been looking through the `tests/` directory, but could not figure out where would be the best place to put it.

nixjdm:
I think a new `test_env.py` is probably the most appropriate.

ericmjl:
Ok, got it! Just added in a test, locally the tests pass:

```
$ pytest tests/test_environment.py
============================================================================ test session starts ============================================================================
platform darwin -- Python 3.6.6, pytest-3.6.3, py-1.5.4, pluggy-0.6.0
rootdir: /Users/[obscured_username]/github/software/lektor, inifile: pytest.ini
plugins: pylint-0.11.0
collected 2 items
-----------------------------------------------------------------
Linting files
Using config file /Users/[obscured_username]/github/software/lektor/pylintrc
-----------------------------------------------------------------

tests/test_environment.py ..                                                                                                                                          [100%]

========================================================================= 2 passed in 0.45 seconds ==========================================================================
```

ericmjl:
@nixjdm please let me know if there's anything else I can do to help move this along! :smile:

nixjdm:
Lgtm! Thanks @ericmjl

PR #592, opened 2018-09-04
include alt when resolving virtual paths with page_num, fixes #216

### LINKED ISSUE #216: [bug] Static alternative url generation if pagination is enabled on a multi-lang project
A project with some alternative content defined:

```
$ cat blabla.lektorproject
...
[alternatives.one]
name = One
url_prefix = /one/
locale = XX
primary = yes

[alternatives.two]
name = Two
url_prefix = /two/
locale = YY
...
```

If pagination is enabled for a model:

```
$ cat models/wtf.ini
...
[pagination]
enabled = yes
per_page = 10
...
```

, and the model have some alternative content

```
$ cat content/wtf.lr
bla = blabla

$ cat content/wtf+two.lr
bla = albalb
```

, trying to fetch an alternative url for this content/model

```
$ cat templates/wtf.html
This is an <a href="{{ '.'|url(alt='two') }}">Alternative</a>
```

**returns a wrong path** -> ever returns the primary alternative

```
This is an <a href="/one/wtf">Alternative</a>
```

**instead of:**  //expected result

```
This is an <a href="/two/wtf">Alternative</a>
```

Reproduced with different _slug for each alternative, and without it.

eljoest:


nixjdm:
Much appreciated, but I had to pick one :)

PR #601, opened 2018-09-30
Lektor-599 Updated theme.ini formats

Andrew-Shay:
- theme.ini files updated
- ConfigParser loads them

```
pytest ./tests/
============================== warnings summary =============================== 
lektor\lekdev\lib\site-packages\typed_ast\ast3.py:58: DeprecationWarning: invalid escape sequence '\s'
  return _ast3._parse(source, filename, mode, feature_version)
lektor\lekdev\lib\site-packages\typed_ast\ast3.py:58: DeprecationWarning: invalid escape sequence '\s'
  return _ast3._parse(source, filename, mode, feature_version)
lektor\lekdev\lib\site-packages\typed_ast\ast3.py:58: DeprecationWarning: invalid escape sequence '\s'
  return _ast3._parse(source, filename, mode, feature_version)
lektor\lekdev\lib\site-packages\typed_ast\ast3.py:58: DeprecationWarning: invalid escape sequence '\s'
  return _ast3._parse(source, filename, mode, feature_version)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
=================== 183 passed, 4 warnings in 30.98 seconds ===================
```

PR #602, opened 2018-10-03
Update README.md

dtranmobil:


nixjdm:
No thanks, I like it better the way it is. If you want to break into editing this repo though, that hash is in need of updating because lektor-themes-nix and this copy just changed (see #601), so you can submit a PR for that.

PR #603, opened 2018-10-06
Adding `lektor dev new-theme`

### LINKED ISSUE #527: Add `lektor dev new-theme` command
This command should be very similar to the existing `lektor dev new-plugin` command, and use the same style prompts to query for basic information that will be used in creating a very basic theme. This theme should then be placed in `themes/` with a specified name, and contain a `theme.ini`. The `theme.ini` should be populated by the answers to the prompts.

I think there should also be a trivial amount of content in the new theme. Perhaps a single `layout.html` template file with very basic contents.

Like `lektor dev new-plugin`, this new command should be able to accept a `--path` option, and a `theme_name` argument.

nixjdm:
Fixes #527 

This provides a new command that works just like the `new-plugin` command. This drops a dir in `themes/` for the new theme, and gives a readme, a partially filled theme.ini, an `example-site/` and `images/` with a come helper files to get started.

Since a theme doesn't have to have any particular set of components, it does not presume or give any. Instead instructs in the readme that they should be made.


tests/test_devcli.py
```diff
@@ -269,3 +265,248 @@ def test_new_plugin_path_and_name_params(project_cli_runner):
...
+
+        # If porting an existing theme
+        [original]
+        author =
+        homepage =
+        repo =
+    """).strip()
+    with open(os.path.join(path, 'theme.ini')) as f:
+        theme_ini_contents = f.read().strip()
+    assert theme_ini_contents == theme_ini_expected
```
runfalk: This is a matter of taste I guess, but isn't it better to test the parsed output? That seems closer to reality and what we guarantee. Reordering fields or adding a new field or section is not a breaking change.

tests/test_devcli.py
```diff
@@ -269,3 +265,248 @@ def test_new_plugin_path_and_name_params(project_cli_runner):
...
+        - flowblocks
+        - models
+        - templates
+
+        Just put those folders here and they'll be picked up when your
+        theme is used.
+    """).strip()
+    with open(os.path.join(path, 'README.md')) as f:
+        readme_contents = f.read().strip()
+    assert readme_contents == readme_expected
```
runfalk: This is basically just duplicating the README template. Isn't it enough to verify that the file exists and is not empty?

I'm also concerned that this i duplicated in multiple tests.

tests/test_devcli.py
nixjdm: Yes, you're right. I've changed it to only test for the values we construct, reading it as the ini it is. So then there's a general test of structure, and then the few specific things that are more likely to break.

tests/test_devcli.py
nixjdm: Yes. I got a little overzealous with an existing pattern in the tests. Now I read and test for the existence of the theme name, which is the only variable in the readme template.

lektor/quickstart.py
```diff
@@ -266,3 +269,78 @@ def plugin_quickstart(defaults=None, project=None):
...
+        theme_name = g.prompt('Theme Name', default=None,
+            info='This is the human readable name for this theme')
+
+    theme_id = theme_name.lower()  # pylint: disable=no-member
+    if theme_id != 'lektor' and theme_id.startswith('lektor'):
+        theme_id = theme_id[6:].strip()
+    if theme_id != 'theme' and theme_id.startswith('theme'):
+        theme_id = theme_id[5:]
+    if theme_id != "theme" and theme_id.endswith('theme'):
+        theme_id = theme_id[:-5]
```
nixjdm, line 291: This is a little more complicated than what exists for the new-plugin command. Namely it doesn't handle well the case where someone just uncreatively names their plugin `lektor plugin`. I bet you can guess how I found that out. That should be updated similarly, but I'll do that in a separate PR.

tests/test_devcli.py
```diff
@@ -269,3 +268,185 @@ def test_new_plugin_path_and_name_params(project_cli_runner):
...
+    (
+        ('Lektor New Theme', 'lektor-theme-new'),
+        ('New Theme', 'lektor-theme-new'),
+        ('New', 'lektor-theme-new'),
+        ('Theme', 'lektor-theme-theme'),
+        ('Lektor', 'lektor-theme-lektor'),
+        ('Lektor Theme New', 'lektor-theme-new'),
+    ),
+)
+def test_new_theme_varying_names(project_cli_runner, theme_name, expected_id):
```
nixjdm, line 441: New test for edge cases of `theme_id` constructions

PR #604, opened 2018-10-06
Replacing custom slugify with python-slugify

nixjdm:


setup.py
```diff
@@ -28,18 +28,19 @@
     zip_safe=False,
     platforms='any',
     install_requires=[
+        'Babel',
+        'EXIFRead',
+        'Flask',
         'Jinja2>=2.4',
         'click>=6.0',
-        'watchdog',
```
xlotlu: it would be nice if this great rearrangement were a different commit
/OCD :)

nixjdm:
This should be pulled just prior to the next minor version bump. It is compatible with the existing slugify for all of the the easy and common cases, but not for many other things. It could potentially rebuild a site with new paths and thus break any hard-coded links.

xlotlu:
@nixjdm

runfalk:
I really think we should put this in the release announcement since it's a breaking change (i.e. indexed pages on search engines may become 404).

Issue #455, opened 2017-09-20
Unclosed file objects in deploy


nixjdm:
See from the most recent deploy of the getlektor website, https://travis-ci.org/lektor/lektor-website.

There are what appear to be some unclosed file objects around line 736:

```
Deploying to Production
  Build cache: /home/travis/.cache/lektor/builds/7895ab62e64441423911ee654bd1b85f
  Target: rsync://deploy@flow.srv.pocoo.org/srv/websites/getlektor.com/static
  Warning: Permanently added the ECDSA host key for IP address '148.251.50.164' to the list of known hosts.
  sending incremental file list
  index.html
  
  sent 21885 bytes  received 410 bytes  6370.00 bytes/sec
  total size is 24250704  speedup is 1087.72
Done!
/home/travis/virtualenv/python3.5.4/lib/python3.5/site.py:165: DeprecationWarning: 'U' mode is deprecated
  f = open(fullname, "rU")
/home/travis/virtualenv/python3.5.4/lib/python3.5/site-packages/lektor/cli.py:326: ResourceWarning: unclosed file <_io.BufferedReader name=5>
  for line in event_iter:
/home/travis/virtualenv/python3.5.4/lib/python3.5/site-packages/lektor/cli.py:326: ResourceWarning: unclosed file <_io.BufferedReader name=7>
  for line in event_iter:
```

bbbart:
same problem here with Lektor 3.1.1 and Python 3.6.5

is this related to the `temp` directory not being removed after deploy in the project's root?

RussellJTech:
I, too, get the same problem with Lektor 3.1.1 and Python 3.6.5.

I just used lektor quickstart to create [a demo site](https://russelljtech.github.io/lektor-demo-site/) and deployed it to [GitHub Pages](https://github.com/RussellJTech/lektor-demo-site). My deploy command was "lektor deploy --output-path ./output". My repos' [README.md](https://github.com/RussellJTech/lektor-demo-site/blob/master/README.md) details the steps that I followed, and the attached file gives my Windows command line output. 
[20180616 lektor deploy output.txt](https://github.com/lektor/lektor/files/2108577/20180616.lektor.deploy.output.txt)

Thankfully, my [demo site](https://russelljtech.github.io/lektor-demo-site/) seems to be functioning fine.

In addition to the "ResourceWarning: unclosed file ..." warnings, there was the following error:

  Fetching origin
  fatal: Couldn't find remote ref refs/heads/gh-pages
[...]

PR #615, opened 2018-10-25
Fix minor deployment issues

uyar:
This PR fixes the warning about unclosed file objects during deploy (#455) and also removes the temporary -empty- directory.


uyar:
The edit link patch is mistakenly included, I'm closing this PR.

PR #620, opened 2018-11-17
Fix #588 - modified utils.py

### LINKED ISSUE #588: `url` filter returns url with one relative `../` too many when called from a dotted slug
When calling the `url` filter from a page with a dotted slug (e.g. `/takecontactform.php`), 
the filter returns a relative url to the designated target that has one relative level `../`
too many.

A template containing `<a href="{{ '/home'|url }}">Home</a>` would point to `../../home`
instead of `../home` when rendered for `/takecontactform.php`.

The same issue applies to the `asseturl`filter.

f-seven:
Modified def make_relative_url(base, target): so it checks whether the base path belongs to a dotted slug. If so, the depth is reduced by one to point to the target at the right level.

Addresses issue #588 in Github.

### DIFF
--- a/lektor/utils.py
+++ b/lektor/utils.py
@@ -537,6 +537,13 @@ def make_relative_url(base, target):
     else:
         depth = ('/' + base.strip('/')).count('/')
         prefix = ''
+        # If the last part of the base path contains a dot, the page will 
+        # render into, for example "404.html" instead of "404/index.html".
+        # Any relative links from it therefore need one less level of depth.
+        if '.' in base.strip('/').split('/')[-1]:
+            depth -= 1
+            if depth == 0:
+                prefix = './'
 
     ends_in_slash = target[-1:] == '/'
     target = posixpath.normpath(posixpath.join(base, target))
--- a/lektor/utils.py
+++ b/lektor/utils.py
@@ -537,7 +537,7 @@ def make_relative_url(base, target):
     else:
         depth = ('/' + base.strip('/')).count('/')
         prefix = ''
-        # If the last part of the base path contains a dot, the page will 
+        # If the last part of the base path contains a dot, the page will
         # render into, for example "404.html" instead of "404/index.html".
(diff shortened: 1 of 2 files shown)


DeeUnderscore:
Whitespace complaints raised by the linter aside, this seems to solve the issue.

runfalk:
Sorry about the silence on this issue. I haven't taken the time to understand the issue this solves.

If you make sure the linter is happy I'll make sure that this PR gets a proper review. I'd also like to see a unit test that covers this behavior.

f-seven:
Thanks for your feedback. I found the issue the linter was complaining about and fixed it.

A unit test would indeed be good. I am very new to this and therefore unsure if I could help. Are there any infos/links on how unit tests are handled in lektor?

runfalk:
I'll be happy to walk you through it. As there doesn't seem to be any existing tests for it you should add a new one to the end of [tests/test_utils.py](https://github.com/lektor/lektor/blob/master/tests/test_utils.py).

```python
@pytest.mark.parametrize("base, target, expected", [
    # A few different edge cases goes here, one testcase per line
    ("/", "./", "/"),
    # ...
    # Add as many test cases as you see fit. You definitely want one test case for the
    # new behavior.
])
def test_make_relative_url(base, target, expected):
    assert make_relative_url(base, target) == expected
```

This uses [pytest](https://pytest.org/)'s [parametrized tests](https://docs.pytest.org/en/latest/parametrize.html) functionality if you need help with the documentation.

To run the tests locally you should simply be able to execute the `pytest` command in the root of the repository (provided that you have installed it).

This is how I setup my development environment: I create a [virtual environment](https://realpython.com/python-virtual-environments-a-primer/) (follow the python 3 instructions) and then I run `pip install -e .[test]` inside that environment to install everything.

f-seven:
Thanks for your help. It has really helped me get started.
 
I committed the changes to test_utils.py.

PR #637, opened 2019-03-07
On reinstall plugins now won't be initialized and won't receive events

germn:
If you run `lektor plugins reinstall` every plugin's `on_setup_env` event will be triggered.

It's unexpected behavior and this PR fixes it. Plugins now will be just reinstalled.


nixjdm:
Thank you for the PR! It indeed should be unexpected behavior.

Without this, someone could more easily run plugin code on install, which is not something I want to encourage.

----
The failing tests are unrelated to this.

PR #644, opened 2019-04-04
Fix no-else-raise pylint warning

vnznznz:


skorokithakis:
This is kind of a silly warning, but I prefer the separate-if shortcircuiting anyway, thanks!

vnznznz:
I just *had* to fix your red CI badges ;)

PR #651, opened 2019-05-12
Lektor 613 - Passing extra_flags to all plugin events

### LINKED ISSUE #613: Make extra flags more broadly available
Right now `extra_flags` is being inconsistently passed to plugin events;

- directly to the `server_spawn`
- nested inside the `builder` object for the events that receive that.
- not at all for others.

I think this should open up, so that all plugin events should directly receive `extra_flags`. This will offer greater functionality for the events that currently don't have it at all (like `setup_env`), and a standard so a plugin developer has a single way of accessing it. After this the docs should be amended to say a little more about the `extra-flags`, and how to access them.

This should not break most existing plugins since the plugin docs have for a long time advised receiving `**extra` in case additional arguments are passed in the future, just like this.

nixjdm:
Fixes #613.

This passes extra_flags to every plugin event, and allows extra_flags in a few more commands like `clean` and `dev shell`. This includes tests for most of the added functionality.

The tests could be improved, but the bigger deficiencies are not easy to overcome:

-  Testing the events hooked by `lektor server`; `server_spawn` and `server_stop`. Maybe there's an easy way I haven't thought of?
- Including cli triggered hooking of the markdown config commands. Tough because of the markdown thread local.
- Cleaning up after creating and installing plugins within the single pytest command. That's much more difficult than I originally realized. See inline comment block.

IMO, none of those are bad enough to block this. I'll happily accept recommended changes though if someone thinks of something I haven't :)

Ping @Andrew-Shay, for https://github.com/lektor/lektor-website/pull/256.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ pex:
 
 test-python:
 	@echo "---> running python tests"
-	py.test . --tb=long -svv --cov=lektor
+	py.test . --tb=long -vv --cov=lektor
 
 coverage-python: test-python
 	coverage xml
--- a/lektor/cli.py
+++ b/lektor/cli.py
@@ -116,17 +116,19 @@ class Context(object):
             return rv
         return self.get_project().get_output_path()
 
-    def get_env(self):
+    def get_env(self, extra_flags=None):
         if self._env is not None:
             return self._env
         from lektor.environment import Environment
-        env = Environment(self.get_project(), load_plugins=False)
+        env = Environment(self.get_project(), load_plugins=False,
+                          extra_flags=extra_flags)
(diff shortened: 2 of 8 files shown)


skorokithakis (approved):
This looks good to me at a first glance, but I'm not super duper familiar with the codebase.

nixjdm:
Looks like I messed up the working directory pathing of the tests a bit. I'm fixing that.

Thanks for the look!

nixjdm:
I think I'm done modifying this, tweaking the tests.

goanpeca:
Looks pretty good in general :-)

PR #656, opened 2019-06-06
[lektor-654] plugin compatibility and deprecation notice

### LINKED ISSUE #654: Improve error reporting when plugins don't catch extra_flags / **extra
By #652, some plugins may not be actually catching `**extra` or `extra_flags`. Since #651, they will need to. If they don't, Lektor should report a more useful error, indicating at least which plugin is causing trouble, so that it can be fixed.

nixjdm:
resolves #654 

This adds back in compatibility with plugins not ready for extra_flags on all events, and enables that with a DeprecationWarning.

### DIFF
--- a/lektor/pluginsystem.py
+++ b/lektor/pluginsystem.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import warnings
 from weakref import ref as weakref
 from inifile import IniFile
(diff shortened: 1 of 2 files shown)


runfalk:
Looks good to me

PR #667, opened 2019-07-26
Update react to v16 and react-router to v3

### LINKED ISSUE #430: Fix React-Router warnings
When you open console on admin edit pages, there are react router warnings. Better fix them, before it's too late (:

```
Warning: [react-router] It appears you have provided a deprecated history object to `<Router/>`, please use a history provided by React Router with `import { browserHistory } from 'react-router'` or `import { hashHistory } from 'react-router'`. If you are using a custom history please create it with `useRouterHistory`, see http://tiny.cc/router-usinghistory for details.
warning
Warning: [react-router] `props.history` and `context.history` are deprecated. Please use `context.router`. http://tiny.cc/router-contextchanges
warning
Warning: [react-router] adding multiple leave hooks for the same route is deprecated; manage multiple confirmations in your own code instead
```

### LINKED ISSUE #431: Fix React warnings

davidferguson:
### Issue(s) Resolved

Fixes #430 
Fixes #431 

### Related Issues / Links

Related to #458 and all other admin interface issues.

### Description of Changes

* [x] Used [react-codemod](https://github.com/reactjs/react-codemod) to automatically modify all instances of React.createClass to native classes, or `createReactClass` where mixins were used.
* [x] Upgraded `react` to v16.8.6 (latest as of writing)
* [x] Upgraded `react-router` to v3.2.2 (a version that supports react v16, but is also mostly compatible with react-router v2)

Having the admin UI run on react v15 is a pain as it prevents adding react components that only support react v16. There has been discussion previously in #430 and #431 about upgrading, however it was always put off due to the issues raised in #458 about people wanting the admin UI re-written, maybe not in React, due to current limitations.

This PR isn't trying to get into that debate. It is more of a "bodge job" than an actual fix, as it does not re-write the code to use pure v16 technologies, and instead relies on packages like `createReactClass` which is "*A drop-in replacement for React.createClass*". Additionally, `react-router` is only upgraded to v3.2.2 and not v4 (or v5 - they seem to have odd versioning) as would be ideal, as v3 is mostly compatible with v2, and also support react v16.

The goal of this PR is to update react in the admin UI, to make it more usable in it's current state.


lektor/admin/package.json
```diff
@@ -19,7 +19,7 @@
     "extract-text-webpack-plugin": "^3.0.0",
     "file-loader": "^0.11.2",
     "font-awesome": "^4.3.0",
```
runfalk, line 22: I think we should depend on `create-react-class` directly. Currently it's indirectly provided by `react-router`.

runfalk (changes requested):
Thank you for this PR.Glad to see the warnings are gone!

I see no problem in merging this as a stop-gap measure until other issues are solved. I'll merge this as soon as my requested change is sorted out.

lektor/admin/package.json
davidferguson: Fixed, I think I assumed react-codemod would add this in automatically, but it didn't.

davidferguson:
This appears to introduce an error that I didn't spot during testing - when you save a content page, it errors out with ``TypeError: Cannot read property 'pushState' of undefined``. I'm looking into this now.

davidferguson:
Fixed, it involved replacing calls to `props.history` with their replacements in `context.router`. This should be good to close now.

lektor/admin/package.json
```diff
@@ -2,7 +2,9 @@
   "name": "lektor",
   "version": "0.0.0",
   "private": true,
-  "dependencies": {},
+  "dependencies": {
+    "create-react-class": "^15.6.3"
```
runfalk: It's a bit weird to have this as a normal dependency when everything else is a dev-dependency :o)

lektor/admin/static/js/components/Component.jsx
```diff
@@ -1,17 +1,17 @@
 'use strict'
 import dialogSystem from '../dialogSystem'
 import BaseComponent from './BaseComponent'
+import PropTypes from 'prop-types'
 
 class Component extends BaseComponent {
-  constructor (props) {
-    super(props)
+  constructor (props, context) {
+    super(props, context)
```
runfalk, line 8: I'm not really sure how context works here. This should call `BaseComponent`'s constructor here (which is the same as `React.Component`). I can't find anything about a second argument for it (https://reactjs.org/docs/react-component.html#constructor). Where is this passed from?

davidferguson: See https://reactjs.org/docs/legacy-context.html#referencing-context-in-lifecycle-methods - I didn't realise it, but it's actually a legacy API that will be removed after React 16. I don't think that's a huge issue though, as it works, and as mentioned this is a stopgap fix to get react to 16, rather than a rewrite to make it work with 16 concepts and APIs.

lektor/admin/package.json
davidferguson: Fixed in 2dbb05e - I think I must have run `--save` rather than `--save-dev`!

lektor/admin/static/js/components/Component.jsx
runfalk: I agree, let's get this merged.

PR #670, opened 2019-07-29
flow rendering fix - admin UI

davidferguson:
### Issue(s) Resolved

*no open issues about this*

Fixes issue of constant unnecessary mounting/unmounting of flow widgets. For complex widgets that parse values on mount (such as gui markdown editor, and others), this has big performance issues.

### Description of Changes

* [x] Make serialization of data occur only when data is loaded by the API
* [x] Make flow's `deserializeValue` a pure function

---

On `EditPage.jsx`, every time a field is updated (ie. user changes text in a strings field), the render method is called.

This causes `renderFormFields` to be called, which in turn calls `renderFormField` for each field. The value prop for each field is set to `getValueForField`.

`getValueForField` looks first at `state.recordData`, but this might not be set, as by default, this is empty, and is only populated when a field calls its `onChange` prop. Therefore, if `state.recordData` is empty (which it will be if a field hasn't been changed and onChange hasn't fired), it grabs the value from `state.recordInitialData`, which is populated by the result of the initial API call.

This value might be serialized, and it wants the deserialized value, so it checks if the static function `deserializeValue` exists on the appropriate widget class, and if so, calls it on the serialized value to get the deserialized value.

This is all fine, except that `deserializeValue` is not a pure function for a flow widget. `deserializeValue` increments the `lastBlockId` counter inside `flowWidget.jsx`, which is used to keep track of how many flow blocks are displayed (ie. when a new flow block is added, `lastBlockId` is incremented).

When deserializing the flow widget content, `lastBlockId` is used to create unique `localId`s for each flow block, which is then used for the `key` prop in the widget. However, as `deserializeValue` isn't pure, and can be called many many times, this leads to the deserialized blocks getting new `localId`s every time it is called, and as this is used for the `key` prop, this forces a unmount/mount of the flow blocks.

The fix for this is to remove `state.recordInitialData`, and make it so that when the data is fetched from the API, instead of setting `state.recordInitialData` to the API data and `state.recordData` to empty, it goes through the API data, deserializes where appropriate, and puts this in `state.recordData`.

This means that deserialization happens **once** - when the data is loaded (which matches up nicely with serialization, which happens once when the page is saved).

Additionally, static method `deserializeValue` was converted into a pure function, which removes the need for the global `lastBlockId` in `flowWidget.jsx`.

### DIFF
--- a/lektor/admin/static/js/views/EditPage.jsx
+++ b/lektor/admin/static/js/views/EditPage.jsx
@@ -13,7 +13,6 @@ class EditPage extends RecordEditComponent {
     super(props)
 
     this.state = {
-      recordInitialData: null,
       recordData: null,
       recordDataModel: null,
       recordInfo: null,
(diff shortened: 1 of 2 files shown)


nixjdm (approved):

nixjdm:
Still looks good to me! Merging. Thanks @davidferguson!

PR #673, opened 2019-07-29
speed up EXIF parsing for imagetools.make_image_thumbnail

### LINKED ISSUE #672: imagetools.make_image_thumbnail is slow because of EXIF parsing
I run a personal photo blog that (1) uses pagination and (2) contains a lot of large image files. It seems that when I change a single file, then all pages of the main pagination need to be rebuilt (I see `index.html`, `page/2/index.html`, ..., `page/14/index.html` in the `lektor build` output), I think this is described in #180.

On my local machine, I used `python -m cProfile lektor build` and investigated the resulting prof file with `snakeviz`. Total build time is about 138 seconds, of which 118 seconds are spent in `imagetools.make_image_thumbnail`, of which 117 seconds are spent in `imagetools.read_exif`. (Note that there are actually no thumbnails to be generated in this `lektor build` run.)

Bildschirmfoto von 2019-07-29 22-52-42

I understand that processing the EXIF header is quite complicated and probably is hard to optimize. However, it seems that the only thing in that header that is used in the `make_image_thumbnail` function is the information whether the image is rotated:

https://github.com/lektor/lektor/blob/076e4fb71d8aeb11b97804de8f20fe654d760623/lektor/imagetools.py#L389-L397

(It seems that actually I myself introduced this call in #321 :sweat_smile:)

Now I wonder, if 85% of the `lektor build` time is spent in parsing the whole EXIF header just for the sake of getting the `is_rotated` information, maybe we can add some dirty hack to get just that particular value out of the JPEG header faster.

I'll definitely look into this myself, because right now editing this blog using the admin interface is not fun...

tgpfeiffer:
The only EXIF information that is used in `imagetools.make_image_thumbnail` is the `is_rotated` field. However, parsing the complete EXIF header (decoding thumbnails etc.) just for that field comes with a significant overhead, as described in #672. I have added a small helper function `get_is_rotated_fast(fp)` that only parses the header until it finds the `Orientation` header and returns the `is_rotated` value.

**The time for `lektor build` that I described in #672 has gone down from 138 seconds to 22 seconds.**

### Issue(s) Resolved

Fixes #672 

### Description of Changes

* [X] Wrote at least one-line docstrings (for any new functions)
* [X] Added unit test(s) covering the changes (if testable)

### DIFF
--- a/lektor/imagetools.py
+++ b/lektor/imagetools.py
@@ -407,6 +407,12 @@ def read_exif(fp):
     return EXIFInfo(exif)
 
 
+def get_is_rotated_fast(fp):
+    """Fast version of read_exif(fp).is_rotated, using an exif header subset."""
+    exif = exifread.process_file(fp, stop_tag='Orientation', details=False)
+    return EXIFInfo(exif).is_rotated
+
+
 def find_imagemagick(im=None):
     """Finds imagemagick and returns the path to it."""
     # If it's provided explicitly and it's valid, we go with that one.
--- a/lektor/imagetools.py
+++ b/lektor/imagetools.py
@@ -392,7 +392,7 @@ def get_image_info(fp):
         # to make decisions based on the "visual", not the "real" dimensions.
         # thumbnail code also depends on this behaviour.)
         fp.seek(0)
-        if get_is_rotated_fast(fp):
+        if is_rotated(fp):
             width, height = height, width
     else:
         fmt = None
@@ -406,7 +406,7 @@ def read_exif(fp):
     return EXIFInfo(exif)
 
 
-def get_is_rotated_fast(fp):
+def is_rotated(fp):
     """Fast version of read_exif(fp).is_rotated, using an exif header subset."""
     exif = exifread.process_file(fp, stop_tag='Orientation', details=False)
     return EXIFInfo(exif).is_rotated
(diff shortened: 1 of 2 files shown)


nixjdm:
Good find! It seems that [this exifread feature was made for just this sort of thing.](https://github.com/ianare/exif-py#stop-at-a-given-tag) I was unaware of it.

This leaves `read_exif` alone, so [Image.exif](https://github.com/lektor/lektor/blob/076e4fb71d8aeb11b97804de8f20fe654d760623/lektor/db.py#L759-L765) should still work just fine, and we can still use it for other future purposes easily. I'll try to remember this trick though if we can use it again.

LGTM. @xlotlu @runfalk, am I missing anything? It seems straight forward enough.

lektor/imagetools.py
```diff
@@ -407,6 +406,12 @@ def read_exif(fp):
     return EXIFInfo(exif)
 
 
+def get_is_rotated_fast(fp):
```
xlotlu: This is only nitpicking: I'd skip the "fast" in the name, since it's a specialised function and there's no "slow" version of it. And the docstring makes it clear enough why it's faster.

LE: In fact I'd just settle for "is_rotated()". This might even be exposed to the template later on.

xlotlu:
+1

lektor/imagetools.py
tgpfeiffer: Renamed as you suggested.

PR #674, opened 2019-07-30
Lock mistune version

lepture:
I'm rewriting mistune now. Mistune 2.0 will break its API.

I'll make another PR when 2.0 is released.


runfalk:
Thank you for the commit!

PR #679, opened 2019-08-01
fix: make negative numbers valid in integer type widget

davidferguson:
### Issue(s) Resolved

*none open*

Fixes bug where `integer` type widget only accepts positive numbers.

### Description of Changes

From the docs on the [integer type](https://www.getlektor.com/docs/api/db/types/integer/):

> The integer type is one of the most basic ones in Lektor. It can store arbitrary natural numbers **both negative and positive**.

However the validation regex for the integer input widget was `/^\d+$/`, therefore not allowing negative numbers to be added.


runfalk:
This looks good to me. It definitely looks like this was the intended use-case to begin with.

PR #680, opened 2019-08-02
replace imghdr with filetype

### LINKED ISSUE #653: Replace use of imghdr to guess image formats
[PEP 594 is drafted to remove imghdr from Python](https://www.python.org/dev/peps/pep-0594/#imghdr). We use it to guess the format of an image only on this line: https://github.com/lektor/lektor/blob/723e9ecd97a305f128de369a4466a2863307b31a/lektor/imagetools.py#L334

We know imghdr provides a poor guess anyway. This ticket removes the use for an upstream fix for #624, so I'll close that.

xlotlu:
### Issue(s) Resolved

Fixes #653

### Description of Changes

The existing tests take care of everything.

### DIFF
--- a/lektor/imagetools.py
+++ b/lektor/imagetools.py
@@ -333,11 +333,7 @@ def get_image_info(fp):
         return get_svg_info(fp)
 
     _type = filetype.image(head)
-    fmt = _type.extension if _type else None
-
-    # preserve backward-compatibility with what imghdr would return
-    if fmt == "jpg":
-        fmt = "jpeg"
+    fmt = _type.mime.split("/")[1] if _type else None
 
     width = None
     height = None
--- a/lektor/imagetools.py
+++ b/lektor/imagetools.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 import decimal
 import os
-import imghdr
 import re
 import struct
 import posixpath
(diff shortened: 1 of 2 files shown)


runfalk:
The patch looks good to me. Looking at the CI there seems to be an issue with Python 2.7 however. I don't think we should break the compatibility now, unless we don't plan a release of Lektor before 2020.

lektor/imagetools.py
```diff
@@ -331,7 +332,8 @@ def get_image_info(fp):
     if any(map(head.strip().startswith, magic_bytes)):
         return get_svg_info(fp)
 
-    fmt = imghdr.what(None, head)
+    _type = filetype.image(head)
```
runfalk: This will throw if the format is not recognized. Looking at the `imghdr.what` source code it will never throw. Should we catch the exception here?

xlotlu:
> there seems to be an issue with Python 2.7 however. I don't think we should break the compatibility now, unless we don't plan a release of Lektor before 2020.

Yeah, I noticed that. I didn't intend to get rid of py2 just like that.. :)

At first I thought this thing is py3-only, but it's only a bug (debatably...) with raw bytes under py2. Whatever you pass as a str it tries to open as a file. When passed bytes in py3 it does the right thing.

lektor/imagetools.py
xlotlu: It doesn't throw, it returns None (see the next line). The exception is only in py2 due to the thing below.

runfalk: You are correct I think. I got confused by the documentation: https://h2non.github.io/filetype.py/v0.1.3/match.m.html#filetype.match.image

xlotlu: Yeah, the docs are pretty lacking.

PR #685, opened 2019-09-08
WIP: Fix base URL for attachments in alternatives

### LINKED ISSUE #313: Attached image and alternatives

tomgross:
### Issue(s) Resolved

Fixes #313 

### Related Issues / Links

Superseeds #470

### Description of Changes

### DIFF
--- a/lektor/markdown.py
+++ b/lektor/markdown.py
@@ -19,8 +19,9 @@ class ImprovedRenderer(mistune.Renderer):
             url = url_parse(link)
             if not url.scheme:
                 # PRIMARY_ALT is used because images aren't generated for each alt
-                link = self.record.url_to("!" + link, alt=PRIMARY_ALT,
-                                          base_url=get_ctx().base_url)
+                link = self.record.url_to(
+                    "!" + link, alt=PRIMARY_ALT, base_url=get_ctx().base_url
+                )
         link = escape(link)
         if not title:
             return '<a href="%s">%s</a>' % (link, text)
@@ -32,8 +33,9 @@ class ImprovedRenderer(mistune.Renderer):
             url = url_parse(src)
             if not url.scheme:
                 # PRIMARY_ALT is used because images aren't generated for each alt
-                src = self.record.url_to("!" + src, alt=PRIMARY_ALT,
-                                         base_url=get_ctx().base_url)
+                src = self.record.url_to(
+                    "!" + src, alt=PRIMARY_ALT, base_url=get_ctx().base_url
+                )
         src = escape(src)
         text = escape(text)
         if title:
--- a/lektor/markdown.py
+++ b/lektor/markdown.py
@@ -6,6 +6,7 @@ from markupsafe import escape
 from markupsafe import Markup
 from werkzeug.urls import url_parse
 
+from lektor.constants import PRIMARY_ALT
 from lektor.context import get_ctx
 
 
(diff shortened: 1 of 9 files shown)


lektor/markdown.py
```diff
@@ -17,7 +18,10 @@ def link(self, link, title, text):
         if self.record is not None:
             url = url_parse(link)
             if not url.scheme:
-                link = self.record.url_to("!" + link, base_url=get_ctx().base_url)
+                # PRIMARY_ALT is used because images aren't generated for each alt
+                link = self.record.url_to(
+                    "!" + link, alt=PRIMARY_ALT, base_url=get_ctx().base_url
+                )
```
dairiki, line 24: I haven't tested this, and I am not that familiar with the alternative feature, but won't this also force all links to pages (that don't include a scheme) to the PRIMARY_ALT as well?  I'm pretty sure that would not be desired behavior.

I think what you really want to do here is remove the "!" prefixing from `link`.  This will (or should) cause `url_to` to resolve the path given by `link` to a db record (e.g. either a `Page` or an `Attachment`) and then ask that record for its URL (via its `.url_path` method).  The record will (or should[^1]) know the correct URL path to itself.

[^1]: Edit: having poked a bit more it seems that `lektor.db.Attachment.url_path` (inherited from `Record.url_path`) is broken.  It currently returns a URL doctored according to the attachments _alt_.  I don't think that's right.  Since there's only one copy of each attachment emitted for alt actually present in the source, I think it should be doctored according to the attachments `_source_alt`.

I've never understood why that `"!"` prefix is inserted here.  It seems like if the author had a use case where they wanted to prevent resolution through the db, they could add the `!` prefix themselves.  Even if there isn't a good reason for it — and there may be — it may be too late to change the behavior now.
To spitball: maybe we could introduce a new prefix which would force resolution through the db.  Or introduce a new _model option_ for the `markdown` type which when set would disable this automatic `!` prefixing.

setup.cfg
```diff
@@ -31,6 +31,7 @@ zip_safe = False
 python_requires = >=3.6
 install_requires =
     Babel
+    beautifulsoup4
```
dairiki, line 34: Is it at all possible to achieve the necessary amount of HTML parsing without introducing new requirements?

lektor/types/formats.py
```diff
@@ -17,3 +23,45 @@ class MarkdownType(Type):
...
+
+    def __html__(self):
+        return self.html
+
+
+class HTMLDescriptor:
+    def __init__(self, source):
+        self.source = source
+
+    def __get__(self, obj, type=None):
```
dairiki, line 41: This adds a lot of new magic to `HtmlType`.
Aside from concerns about rewriting links to pages to force them to the PRIMARY_ALT, this will probably result in surprises to existing users, who will find their links suddenly mangled.

I'm not convinced this is the way to go, at all.  But if it is, this functionality should probably be enabled by a new _model option_ for the `html` type so as not to break existing sites.

lektor/types/formats.py
```diff
@@ -17,3 +23,45 @@ class MarkdownType(Type):
 
     def value_from_raw(self, raw):
         return MarkdownDescriptor(raw.value or u"")
+
+
+# Wrapper with an __html__ method prevents Lektor from escaping HTML tags.
```
dairiki, line 28: Is there any reason not to use [`markupsafe.Markup`][1] here (instead of `HTML`)?

[1]: https://markupsafe.palletsprojects.com/en/2.0.x/escaping/

dairiki (commented):
Thank you for the PR!

The sites I maintain are all monolingual, so I am not overly familiar with the _alternatives_ feature.  Nevertheless, here are some comments.

tomgross:
Thanks for looking into this. I think #958 fixes the issue in a better way.

Issue #611, opened 2018-10-18
Lektor thumbnail: set colorspace in imagemagick


rponline:
People tend to feed lektor any JPG files - also with e.g. Photoshop-CMYK-JPGs.

But JPGs with uncommon colorspaces tend to compress very badly and so on.  
For me, it resulted in JPG files wtith a few MB filesize on the website.

 I propose to add `-colorspace sRGB` (it's the most common in the web) in every imagemagick operation.

rponline:
https://github.com/lektor/lektor/compare/master...rponline:patch-1

xlotlu:
On a related note, it seems that much of the filesize is in fact due to the embedded color profile itself:

```shell
$ convert cmyk.jpg -strip cmyk-x.jpg 
$ convert rgb.jpg -strip rgb-x.jpg 
$ stat -c '%n: %s' cmyk*.jpg rgb*.jpg 
cmyk.jpg: 1041608
cmyk-x.jpg: 80345
rgb.jpg: 80148
rgb-x.jpg: 71582
```

nixjdm:
I don't know, but if there is no safe and good colorspace (as asked in the PR), which intuitively seems reasonable to me, then we shouldn't force one. Maybe make it an optional parameter?

rponline:
Optional parameter sounds good. 
Maybe also for the strip flag.

xlotlu:
I personally lean towards the initial suggestion: hardcode it for any thumbnail operation. Given that browsers have to do the conversion anyway, and a thumbnail ready for print doesn't make sense, I don't even see a point in making it a default and giving control to the users. (... it may be a good idea for some future-planned operations, like simple conversion, but not for thumbnailing).

The one question that needs answering is if it's the right / safe thing to do in all cases. A short and inconclusive test shows that imagemagick doesn't touch grayscale images, despite the `-colorspace sRGB`, and in all other cases... everything uses sRGB, correct?

P.S.: I'd even go so far to make `-strip` default, or at least `+profile '*'` (as strip clears both profiles and comments). I've read someplace that browsers assume sRGB if they don't find an embedded profile, but this should be investigated.

skorokithakis:
What do we want to do here? Just add `-strip` or also `-colorspace sRGB`? It's a pretty straightforward change, and since `process_image` is only called by `make_image_thumbnail`, it shouldn't be too dangerous.

xlotlu:
[...]

Issue #609, opened 2018-10-16
allow for additional thumbnail modes


xlotlu:
We should allow for additional image processing modes (this is also a more appropriate name than "thumbnail modes"), which can be provided by plugins.

One candidate might be [seam-carving](https://upload.wikimedia.org/wikipedia/commons/b/bf/Broadway_tower_seam_carving_interactive.svg) ("liquid rescale"), which can be provided by imagemagick, so it would be based on the current code.

Another implementation could for example be provided by using [libvips](https://github.com/libvips/libvips/), either by direct bindings or by using the [vipsthumbnail](https://libvips.github.io/libvips/API/current/Using-vipsthumbnail.md.html) binary, much like the current code. (Besides speed improvements, of much interest is its [smart crop](https://libvips.github.io/libvips/API/current/Using-vipsthumbnail.md.html#cropping) feature).

Of course, one plugin could even provide the same existing modes, but by using native imagemagick bindings instead.

Technically, this is probably best achieved by transferring  the current thumbnail-generation code into stand-alone classes / functions, and providing these callables as the "image processing modes".

PR #692, opened 2019-10-03
Add an `extra_params` argument to `process_image` (#691)

### LINKED ISSUE #691: Allow process_image to accept extra parameters to ImageMagick
I have found that I can halve my image size if I pass `-strip` to ImageMagick. Due to my thumbnail plugin being what it is, I would greatly benefit by `process_image` being able to accept a list of extra parameters to pass to ImageMagick.

It would not be appropriate to include these parameters in `process_image` in mainline because not everybody wants thumbnail-quality, but for my plugin they're ideal. Would this be a desirable PR to open? If so, I can submit it ASAP.

What do you think, @nixjdm?

skorokithakis:
### Issue(s) Resolved

Fixes #691 

### Description of Changes

As per #691, this adds the `extra_params` argument to `process_image()` so plugins can pass their own params to ImageMagick.

### DIFF
--- a/lektor/imagetools.py
+++ b/lektor/imagetools.py
@@ -477,7 +477,7 @@ def compute_dimensions(width, height, source_width, source_height):
 
 def process_image(ctx, source_image, dst_filename,
                   width=None, height=None, mode=ThumbnailMode.DEFAULT,
-                  quality=None):
+                  quality=None, extra_params=None):
     """Build image from source image, optionally compressing and resizing.
 
     "source_image" is the absolute path of the source in the content directory,
@@ -508,6 +508,9 @@ def process_image(ctx, source_image, dst_filename,
                     '-extent', resize_key]
     else:
         cmdline += ['-resize', resize_key]
+
+    if extra_params:
+        cmdline.extend(extra_params)
 
     cmdline += ['-quality', str(quality), dst_filename]
 


skorokithakis:
Some tests are failing but I'm not sure why, this PR looks unrelated.

goanpeca (approved):

skorokithakis:
@goanpeca thanks! Do we merge or wait for @nixjdm?

goanpeca:
>Do we merge or wait for @nixjdm?

Let's wait ;-), he should not take long :-)

skorokithakis:
Sure thing, I'm not in a rush (due to backwards compatibility I had to vendor the function anyway in thumbnail-generator 0.4).

xlotlu:
> > Do we merge or wait for @nixjdm?
> 
> Let's wait ;-), he should not take long :-)

Yes, please, wait. As far as I'm concerned this is a very big hammer that ties us closer to imagemagick as a cli application, which I don't think is a good idea. I was looking forward to replaceable backends, and I think a clean API is more desirable.

This was also previously discussed under #611, and the general consensus was to make `-strip` default for thumbnails.

runfalk:
I agree with @xlotlu that we shouldn't tie the API to ImageMagick commandline. It would prevent us from ever changing from it in a backwards compatible way with plugins. I making `-strip` a default argument like #611 suggests is fine.

goanpeca:
>I agree with @xlotlu that we shouldn't tie the API to ImageMagick commandline.

How is adding a parameter named `extra_params` where anything can be passed to whatever CLI tool lies underneath is making that method tied to ImageMagick?

skorokithakis:
@goanpeca that wouldn't work, because then the plug in would be mandating what you can use, instead of the user choosing whichever backend they wanted.

You basically can't reconcile pluggable backends with this, so I'll close it.

xlotlu:
> How is adding a parameter named `extra_params` where anything can be passed to whatever CLI tool lies underneath is making that method tied to ImageMagick?

@goanpeca one main reason for this was to not use a CLI tool, but a native python library (presumably it should be way faster than spawning a new process for each thumbnail). The idea is documented under #609.

xlotlu:
@skorokithakis would you like to take a swing at #611?  I'd do it myself but will be afk for ~1 week, also as an API user you're bound to come up with the best solution I guess.

nixjdm:
I agree with @xlotlu, a swappable backend and cleaner API is the way forward for imagetools, and this would make it more difficult to decouple things and not break unknown plugins.

I'm in favor of adding in `-strip` to core with #611.

skorokithakis:
Sure, I can have a look at #611, especially since with this recent familiarity with the image processing pipeline. I'll comment there about that and `-strip`.

PR #612, opened 2018-10-18
add colorspace sRGB to imagemagick commands

rponline:
https://github.com/lektor/lektor/issues/611

PR #694, opened 2019-10-03
Set the color space for thumbnails and strip profiles/comments (#611)

skorokithakis:
This implements the changes proposed in #611 by adding the desired command line arguments to `process_image()`.


runfalk:
I guess this change means we can close #612?

runfalk:
I think it's fine as it is too.

PR #695, opened 2019-10-03
Add support for output_path in project file

### LINKED ISSUE #688: Output patch in project file
Hi,

wouldn't it be easier to specify the output path already in the project file? The default output path on MacOS is really weird and it would make life much easier than specifying the path each time on the command line.

Regards, Thomas

georgeyk:
### Issue(s) Resolved

Fixes #688 

### Description of Changes

Lektor project file now supports "output_path" key.
The preference order will look like:
* Values set via cli `-O/--output-path`
* `LEKTOR_OUTPUT_PATH` environment variable
* `output_path` in lektor project file
* default cache folder (os-specific)

I haven't found any similar tests in the codebase, so it would be helpful if I get some direction here.

### DIFF
--- a/lektor/project.py
+++ b/lektor/project.py
@@ -98,7 +98,8 @@ class Project(object):
         config = self.open_config()
         output_path = config.get('project.output_path', '')
         if output_path:
-            return os.path.normpath(output_path)
+            return os.path.join(self.tree, os.path.normpath(output_path))
+
         return os.path.join(get_cache_dir(), 'builds', self.id)
 
     def get_package_cache_path(self):
--- a/lektor/project.py
+++ b/lektor/project.py
@@ -96,7 +96,7 @@ class Project(object):
     def get_output_path(self):
         """The path where output files are stored."""
         config = self.open_config()
-        output_path = config.get('project.output_path', '')
+        output_path = config.get('project.output_path')
         if output_path:
             return os.path.join(self.tree, os.path.normpath(output_path))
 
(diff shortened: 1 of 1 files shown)


runfalk:
Thank you for your contribution!

I haven't tested this yet, but I have a question. How does this work with relative files? Will it depend on where the `lektor` command is launched, or will it always be relative to the project file?

I think it makes more sense to make it relative to the project file for predictability.

georgeyk:
hey @runfalk 
I did some tests using `output_path = _build` in the project file and invoking `lektor` inside `templates/` folder (for example) and it worked as expected.
Maybe I could add something like:
```
if not os.path.isabs(output_path):
    output_path = os.path.join(self.project_path, output_path)
```
It is more explicit, imo.

runfalk:
I guess Lektor always changes it's working directory to the project files'. This behavior will likely never change as it could break things. Is this consistent for both `lektor server` and `lektor build`?

I like your proposed fix as it is a bit more explicit. I think you don't even need the `isabs`-check:

```
>>> os.path.join("/some/abs/path", "/another/abs/path/here")
'/another/abs/path/here'
```

georgeyk:
>  Is this consistent for both lektor server and lektor build?

I've played around a bit after the last commit and I think so.
I've configured output_path to a nested folder relative to project-root and also absolute paths nested and on `/tmp/blabla' (build, serve and clean commands).

runfalk:
I think this is fine now. I was a bit concerned about the usage of `self.open_config()` since it makes it impossible to have projects without a project file. However, that functionality seems to have been broken since February 2016 (3 months after the initial commit).

I will merge this after testing it.

georgeyk:
Ok. I'll try to update docs right after.

Issue #693, opened 2019-10-03
Add black/isort to the CI and codebase


skorokithakis:
What does everyone think about adding pre-commit with black/isort/flake8 to the codebase? It would make everything look more consistent and would enforce the style at the CI level. flake8 would also get rid of some common bugs.

davidism:
I did this relatively recently for the Pallets projects and am definitely happy with the results. I went with [reorder-python-imports][1] instead of isort because it followed the same "no configuration" policy as black. You can check out Werkzeugs's [pre-commit config][2] and [flake8 config][3] if you'd like a starting point.

[1]: https://github.com/asottile/reorder_python_imports
[2]: https://github.com/pallets/werkzeug/blob/e7ba08f209477cb453f15113f9a4d527a6e81bfe/.pre-commit-config.yaml
[3]: https://github.com/pallets/werkzeug/blob/e7ba08f209477cb453f15113f9a4d527a6e81bfe/setup.cfg#L30-L54

skorokithakis:
Thanks for that, I've been looking for a project that is compatible with black and is opinionated. I will look into it, but do you know if `reorder-python-imports` can recognize first/third party imports without looking at the env? This is the biggest problem I've had with `isort` in `pre-commit`, although @asottile's `seed-isort-config` kind of mitigates it.

davidism:
I haven't had to do anything to get it to split into built-in / third-party / local groups.

skorokithakis:
Fantastic, thank you.

asottile:
it can yes, I wrote both reorder-python-imports and seed-isort-config (they use the same machinery to determine the third-partied-ness of modules)

skorokithakis:
This is great, @asottile, thank you. I'm going to switch everything to `reorder-python-imports`, since it looks like it solves a lot of my isort frustrations.

asottile:
:tada: I also wrote pre-commit and maintain flake8 if you need help there too <3

skorokithakis:
Did you? Damn, you wrote pretty much my entire linting stack, props. Would you like to join [Code Shelter](https://www.codeshelter.co/) as a maintainer, by any chance?

nixjdm:
I'm definitely want to blacken Lektor. I use it in several other projects and love it. Thank you @davidism for your input too. `reorder-python-imports` is something I could use more of in my life.

[...]

PR #696, opened 2019-10-03
Beautify code (implements #693)

skorokithakis:
I have added pre-commit to CI and added `black/reorder-python-imports/flake8` to it. This should help keep the code style consistent and weed out many bugs before even committing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,15 @@
+repos:
+-   repo: https://github.com/ambv/black
+    rev: 19.3b0
+    hooks:
+    - id: black
+-   repo: https://github.com/asottile/reorder_python_imports
+    rev: v1.7.0
+    hooks:
+    -   id: reorder-python-imports
+-   repo: https://gitlab.com/pycqa/flake8
+    rev: 3.7.8
+    hooks:
+    - id: flake8
+      args: ["--config=setup.cfg"]
+      language_version: python3
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,15 +1,15 @@
 repos:
--   repo: https://github.com/ambv/black
-    rev: 19.3b0
-    hooks:
-    - id: black
--   repo: https://github.com/asottile/reorder_python_imports
-    rev: v1.7.0
-    hooks:
-    -   id: reorder-python-imports
--   repo: https://gitlab.com/pycqa/flake8
-    rev: 3.7.8
-    hooks:
-    - id: flake8
-      args: ["--config=setup.cfg"]
-      language_version: python3
+- repo: https://github.com/ambv/black
+  rev: 20.8b1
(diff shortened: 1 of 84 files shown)


nixjdm:
Ref - Fixes #693

skorokithakis:
Oh huh, I thought the title reference would work, thanks.

nixjdm:
All maintainers who see this, wait on the merge until after 3.2 is cut please. :) I want to do that this month.

skorokithakis:
Fine, fine, I won't merge yet :P

runfalk:
I think we should also update the *Want to develop on Lektor?* section in the README on how to set up a pre-commit. Alternatively we should add support to the `make install-git-hooks` command.

I also think this is a good time to change the README to be Python 3 favored (use `python3 -m venv --prompt="(lektor)" venv` rather than `virtualenv venv`).

Maybe we should also add `black` to `setup.py` as an `extra_requires` target (named `dev`)?

skorokithakis:
I agree with everything @runfalk says, except the additional dependency, since pre-commit installs its tools itself in a virtualenv.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,15 @@
+repos:
+-   repo: https://github.com/ambv/black
+    rev: 18.9b0
```
nixjdm: These versions should be updated and the tools rerun before merge to keep out as much churn as we can.

davidism: `pre-commit autoupdate` will set the latest versions, then `pre-commit run --all-files` will reapply them.

nixjdm:
Thank you @runfalk for [mentioning git hyper merge](https://github.com/lektor/lektor/issues/693#issuecomment-538289403). I hadn't heard of it. It sounds good to me to add a short blurb about it to the readme. At the least, adding a `.git-blame-ignore-revs` seems like a good move. To support that most, all of the formatting changes should be in a single commit (or at least black). For instance, this whole PR could be squashed into a single commit, and one additional that adds `.git-blame-ignore-revs` with the previous commit's hash.

The readme could also use a black badge.

Sound good @skorokithakis?

skorokithakis:
Sure thing, done.

skorokithakis:
What's the status of this? I imagine it has gotten pretty stale.

runfalk:
I guess we need to release 3.2 first, which is blocked by #698

yagebu:
@skorokithakis: 3.2 is out, so we can do this now :) Do you have time to get this up-to-date or do you want me to take this over?

skorokithakis:
It's fine @yagebu, I can do it, though we should do the following:

* Scrap this branch.
* Branch again from master.
* Add the configs and beautify everything with them.
* Add the commit SHA to the ignore file.
* Merge.

skorokithakis:
Closing in favor of #823.

PR #701, opened 2019-10-20
Added a 'Save and Preview' button and made the 'Save' button fire an alert when successful and reload the page so changes are reflected.

### LINKED ISSUE #92: Transition Behavior of "Save Changes" on the Admin page
After saving changes in the admin, we're transported to the "Preview" screen. I like keeping a preview open in a separate browser and alternating between admin and that preview.

With the use case above, it'd be nice to **not** have this in `EditPage.jsx` `saveChanges`

```
this.transitionToAdminPage('.preview', {
  path: this.getUrlRecordPathWithAlt(path)
});
```

I can think of different designs, including a two button ("save changes" and "save and preview"). Alternatively, if there is an alert/status system in the code (I haven't checked), it could display an alert that changes have been saved with a link to the preview.

Is this open to discussion, or does it make sense to back burner such a thing at this stage?

Thanks,
John

hannylicious:
### Issue(s) Resolved

#92  

Fixes #

### Description of Changes

https://user-images.githubusercontent.com/3648225/67443853-3605e400-f5cc-11e9-8ae3-422934e2af32.gif

Per the recommendation in issue #92 I added a secondary button. The 'Save Changes' button now alerts the user on successful changes and the 'Save and Preview' button saves the changes and then takes the user to the preview (same as before)


lektor/admin/static/js/views/EditPage.jsx
```diff
@@ -130,6 +130,29 @@ class EditPage extends RecordEditComponent {
   }
 
   saveChanges () {
+    const path = this.getRecordPath()
+    const alt = this.getRecordAlt()
+    const newData = this.getValues()
+    utils.apiRequest('/rawrecord', { json: {
+      data: newData, path: path, alt: alt },
+    // eslint-disable-next-line indent
```
runfalk: The formatting around here looks a bit iffy. Is this suppression necessary?

lektor/admin/static/js/views/EditPage.jsx
```diff
@@ -130,6 +130,29 @@ class EditPage extends RecordEditComponent {
...
+      method: 'PUT' }, makeRichPromise)
+      .then((resp) => {
+        this.setState({
+          hasPendingChanges: false
+        }, () => {
+          this.transitionToAdminPage('.edit', {
+            path: this.getUrlRecordPathWithAlt(path)
+          })
+          window.location.reload()
```
runfalk: I don't think code after this point is guaranteed to run.

lektor/admin/static/js/views/EditPage.jsx
```diff
@@ -130,6 +130,29 @@ class EditPage extends RecordEditComponent {
...
+          if (!resp.okay) {
```
runfalk: Why is this statement inverted?

lektor/admin/static/js/views/EditPage.jsx
```diff
@@ -225,6 +248,8 @@ class EditPage extends RecordEditComponent {
         <div className='actions'>
           <button type='submit' className='btn btn-primary'
             onClick={this.saveChanges.bind(this)}>{i18n.trans('SAVE_CHANGES')}</button>
+          <button type='submit' className='btn btn-primary'
```
runfalk: Should both buttons be marked as primary?

lektor/admin/static/js/views/EditPage.jsx
```diff
@@ -130,6 +130,29 @@ class EditPage extends RecordEditComponent {
...
+      .then((resp) => {
+        this.setState({
+          hasPendingChanges: false
+        }, () => {
+          this.transitionToAdminPage('.edit', {
+            path: this.getUrlRecordPathWithAlt(path)
+          })
+          window.location.reload()
+          if (!resp.okay) {
+            window.alert(i18n.trans('EDIT_SAVED_SUCCESFULLY'))
```
runfalk: If possible I think it's better if the user doesn't have to dismiss the dialog manually. Maybe a 3 second timeout can be used?

I think there already exist classes to support similar dialog messages in the existing code.

runfalk (changes requested):

runfalk:
Thank you for your contribution! I've added some comments to your PR.

lektor/admin/static/js/views/EditPage.jsx
hannylicious: Is there a different CSS selection that should be used here? My logic here was "they both save - which is the primary function - but one refreshes the page, one takes you to preview"; so I felt that leaving the styling the same would make that evident. If not, I can change it - no worries!

lektor/admin/static/js/views/EditPage.jsx
hannylicious: I really dig that idea. I'll look into that!

lektor/admin/static/js/views/EditPage.jsx
runfalk: I think you can use bootstraps `btn-secondary` for one of them. I think it's best to have the least surprising (or most common) action be marked `btn-primary`. That way `btn-primary` means: *"This is likely what you want"*.

The real question which is most sensible to promote as default. I'm leaning slightly towards saving and not loading the preview as the primary choice, but I'll change my position if there are any compelling arguments :smile:

lektor/admin/static/js/views/EditPage.jsx
hannylicious: I added a successDialog component in the most recent pull request to make it similar to the other dialogs. I couldn't find any of them set to auto close after a particular timeout, so I made this one work identical to the existing ones. If it's decided that a timeout feature to close automatically is a need, I figure we can revisit it.

lektor/admin/static/js/views/EditPage.jsx
hannylicious: I have made the requested changes here and made it a secondary button.

lektor/admin/static/js/views/EditPage.jsx
hannylicious: I found that in a similar spot - but I've since removed it to handle things more in line with elsewhere in the code.

lektor/admin/static/js/views/EditPage.jsx
hannylicious: I've removed that commented line. It was in the core that way - so I just left it in there in case it served some purpose at some point - but I've removed it.

goanpeca:
@hannylicious thanks or working on this. Could you add a screenshot of  how it looks on the description of this issue?

Since it is a UI change, it helps to see the actual image of how it looks.

hannylicious:
Peek 2019-10-23 19-29

This is a gif showing the new buttons in action and the page refresh on successDialog close. Don't mind the index.html template not showing up.

PR #707, opened 2019-11-03
Add options to rsync publisher

### LINKED ISSUE #705: rsync missing --delete option
`rsync` does not delete anything removed from source on the destination directory during deploy. In order to keep two directories truly in sync, it's necessary to add `--delete` option to `rsync`. Can we change the behavior of `lektor deploy` by adding this option beside `-rclzv` options on `class RsyncPublisher(Publisher)`?

m-lnrs:
### Issue(s) Resolved

Fixes #705 

### Description of Changes

* [x] added `--delete` argline option
* [x] added `--exclude` argline options
* [x] added tests for `--delete`
* [x] added tests for `--exclude`

### DIFF
--- a/lektor/publisher.py
+++ b/lektor/publisher.py
@@ -204,9 +204,14 @@ class RsyncPublisher(Publisher):
 
     def get_command(self, target_url, tempdir, credentials):
         credentials = credentials or {}
-        argline = ['rsync', '-rclzv', '--delete', '--exclude=.lektor']
+        argline = ['rsync', '-rclzv', '--exclude=.lektor']
         target = []
         env = {}
+
+        options = target_url.decode_query()
+        delete = options.get('delete') in ('on', 'yes', 'true', '1')
+        if delete:
+            argline.append('--delete-delay')
 
         keyfile = _write_ssh_key_file(os.path.join(
             tempdir, 'ssh-auth-key'), credentials)
--- a/lektor/publisher.py
+++ b/lektor/publisher.py
@@ -211,7 +211,7 @@ class RsyncPublisher(Publisher):
         options = target_url.decode_query()
         exclude = options.getlist('exclude')
         for file in exclude:
-            argline.append("--exclude='{}'".format(file.strip("'").strip('"').replace("'", "\\'").replace('"', '\\"')))
+            argline.extend(('--exclude', file))
 
         delete = options.get('delete', False) in ('', 'on', 'yes', 'true', '1', None)
         if delete:
--- a/lektor/publisher.py
+++ b/lektor/publisher.py
@@ -209,6 +209,10 @@ class RsyncPublisher(Publisher):
         env = {}
 
         options = target_url.decode_query()
+        exclude = options.getlist('exclude')
+        for file in exclude:
+            argline.append("--exclude='{}'".format(file.strip("'").strip('"').replace("'", "\\'").replace('"', '\\"')))
+
         delete = options.get('delete', False) in ('', 'on', 'yes', 'true', '1', None)
         if delete:
             argline.append('--delete-delay')
--- a/lektor/publisher.py
+++ b/lektor/publisher.py
@@ -209,7 +209,7 @@ class RsyncPublisher(Publisher):
         env = {}
 
         options = target_url.decode_query()
-        delete = options.get('delete') in ('on', 'yes', 'true', '1')
+        delete = options.get('delete', False) in ('', 'on', 'yes', 'true', '1', None)
         if delete:
             argline.append('--delete-delay')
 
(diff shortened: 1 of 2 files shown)


runfalk:
Thank you for your contribution. I think your fix is sensible and the current behavior confused me the first time I tried it as well. My only concern is that changing this potentially is destructive and could cause data loss if user have added their own sub directories to their deployments.

I don't know how to mitigate this problem (if we even think there is one?). Maybe we could provide two protocols `rsync://` and `rsync+delete://`? Or maybe we could add more configuration options to the target section in the lektor project file.

xlotlu:
>  * added `--delete` as a `argline` option

This should probably be `--delete`, `--delete-after` (or better, `--delete-delay`), to make sure there won't be any deletions in case of failure.

xlotlu:
> I think your fix is sensible and the current behavior confused me the first time I tried it as well.

> I don't know how to mitigate this problem (if we even think there is one?).

I think keeping stale files around is a serious issue, something that under certain circumstances could be a security threat, or have potential for public embarrassment, or who knows what else.

> Maybe we could provide two protocols `rsync://` and `rsync+delete://`? Or maybe we could add more configuration options to the target section in the lektor project file.

~~It's kinda bad that provider-specific options aren't supported in the config.~~

Edit: Scratch that, other providers implement this using url query syntax, see https://www.getlektor.com/docs/deployment/ghpages/#cname-support and `?passive` for ftp.

This should be implemented just the same, with `?delete` defaulting to false.

m-lnrs:
I've read all the valuable comments and tryed to make it the same way it's done on [Lektor FTP](https://www.getlektor.com/docs/deployment/ftp/). Now to turn on the `--delete-delay` option on `rsync`, it's only necessary to include `?delete=yes` on the `rsync` URL. What do you think?

m-lnrs:
> `delete = options.get('delete', False) in ('on', 'yes', 'true', '1', None)`

In order to `?delete` implies `True` I've needed to change this line this way:

`delete = options.get('delete', False) in ('', 'on', 'yes', 'true', '1', None)`

xlotlu:
@m-lib just one more thing please, update the documentation, it's (unfortunately) under a different project: https://github.com/lektor/lektor-website/tree/master/content/docs/deployment

m-lnrs:
Hello @xlotlu, for sure I can update the documentation too :-)

But before that, I recently realized what @runfalk told with:

> My only concern is that changing this potentially is destructive and could cause data loss if user have added their own sub directories to their deployments.

Here in my work, I recently needed to add some symbolic links inside the deploy directory on the server, but with the delete option enabled these links also were removed as expected. Those links only make sense on the server and thats the reason I don't want to keep it on my dev machine... at the same time, I also need that `rsync` revome directories that no longer exists to prevent people accessing informations that should not be there anymore so... to mitigate this, I think we can add this lines as well for these cases where the user need to choose what must be ignored by `rsync` when deploying with delete option:

```
options = target_url.decode_query()
exclude = options.getlist('exclude')
if exclude:
    for file in exclude:
        argline.append('--exclude=' + file)
```

Thus, we can pass one or more files to be ignored by `rsync` on the `url`:

`target = rsync://000.00.0.000:0000/path/to/deploy?delete=yes&exclude=file_one&exclude=file_two`

What do you guys think?

Thanks.

xlotlu:
> But before that, I recently realized what @runfalk told with:
> 
> > My only concern is that changing this potentially is destructive and could cause data loss if user have added their own sub directories to their deployments.

It might be a good idea to add a warning about that.

> What do you guys think?

I think it's a very good addition, but...

> ```
>     for file in exclude:
>         argline.append('--exclude=' + file)
> ```

I expect this "`'...' + file `" to need some escaping. What happens if the file contains space, or an apostrophe, or quotes?

Also, minor nitpick,

> ```
> exclude = options.getlist('exclude')
> if exclude:
> ```

Doesn't `.getlist()` always return a list? You can probably skip the `if` and jump straight to the `for`.

davidism:
Can someone change the title to something more descriptive? I keep forgetting what this discussion is when it pops up in notifications.

m-lnrs:
Thanks for the feedback @xlotlu

I added some escaping to file names and removed the unnecessary `if` statement. If everything is ok now, please tell me so I can start editing the documentation page.

xlotlu:
@m-lib thanks for all the work. I see the tests look just fine in the unified diff, so after one last modification this should be it.

m-lnrs:
Now I see...

I changed the manual escaping for `argline.extend(('--exclude', file))`. It's seems much better now :-D

I also removed some tests that was added by myself. That is why `diff` looks so strange. Sorry for that.

Thanks for the all help!

m-lnrs:
> @m-lib just one more thing please, update the documentation, it's (unfortunately) under a different project: https://github.com/lektor/lektor-website/tree/master/content/docs/deployment

Done: https://github.com/lektor/lektor-website/pull/290

Thanks :-)

xlotlu:
@m-lib sorry for the delay, and thank you for the code

Issue #668, opened 2019-07-26
_raw_slug?: support URLs ilke "über" or "VM"


evilham:
I'm studying migrating some sites to lektor and find two cases that are not well handled in lektor. It's a bit of an issue because I'd like the migration to not break / break as few URLs as possible.

## 1. Uppercase in slug
This one is a matter of taste, but acronyms are uppercase, so are proper nouns and in languages like German all nouns.
I design URLs to be *readable* and *hackable* but not with the expectation that a visitor will type them (though they still could if they wanted).

### Solution:
In this case, the patch is removing `.lower` in following line works as I'd expect it.
https://github.com/lektor/lektor/blob/83a2ba58eb964e2c7b8ae258aead456677d267a6/lektor/utils.py#L384

Since that'd be nasty to existing instances of lektor, I am tempted to propose a `_raw_slug` field that basically is just stripped of whitespace and used as-is.
That would also enable following case

## 2. Unicode chars in slug

I found an archive of a gitter discussion about this, where it was argued that it was a mess because of OS support, while I empathise with that, unicode support has gotten significantly better.
And it while it can be useful for lektor to have an opinion in many cases, sometimes it'd be useful to leave things to the webmaster.

### Solution
I tried patching `lektor.utils.slugify` which does encode/decode tricks to strip non-ascii chars, but that wasn't enough, they are still stripped out somewhere.

With some guidance and knowing what would be acceptable I'd be willing to create a patch that can be part of a new version of lektor, I'd just hate to maintain my local fork of lektor because of these minor details :-).

nixjdm:
Have you seen #604? That is mostly intended to solve handling unicode chars as best as we can, by outsourcing the solution, and be pretty compatible with existing slugs. I'm aiming to pull that change prior to 3.2, which I think is not far out.

With that, Lektor would be using python-slugify. That lib has an option to turn case sensitivity off. We could use python-slugify for this by allowing page slugs to use the [python-slugify options](https://github.com/un33k/python-slugify#options). We'd have to think about how to integrate that exactly, e.g. another system field?

evilham:
I actually hadn't seen that, I'd have to take a deeper look but that sounds like the ultimative solution for this kind of issues as long as the slugify options can be passed in a config file.

FWIW: I don't consider the fact that it *could* break some weird setups a huge issue, as long as it's in the release notes and documented that users should check their settings and builds.
In general for this kind of software one should pin specific versions and update as needed, not blindly running latest shiny.

[...]

PR #714, opened 2019-11-24
Relative URLs, fixes #713

### LINKED ISSUE #713: Allow relative urls abbreviated when they are on the same level
Can we have relative URLs that do not always go to the lowest folder and then back up again?

For example if I browse the following page:
`/articles/2019/07/brownies`
which has a subpage I would like to link to:
`/articles/2019/07/brownies/image-gallery`

The resulting relative url will be:
`../../../../articles/2019/07/brownies/image-gallery/`
instead of just:
`image-gallery/`

Also linking to any other resource may be shorter (e.g., same folder):
`../chocolate-cake/`

relikd:
### Issue(s) Resolved
Fixes #713

### Description of Changes

Actual relative URLs that omit unnecessary folder traversing.

Example:
`/articles/2019/07/brownies` -> link to ->
`/articles/2019/07/brownies/image-gallery`

Previous output:
`../../../../articles/2019/07/brownies/image-gallery/`
Ouput with this PR:
`image-gallery/`

**Note:**
Resources in the same directory are linked without `./` in front.
With the only exception if the link target is the same.
`/blog/`
-> link to `/blog/` will result in `./`
-> link to `/blog/a.jpg` will result in `a.jpg`

Tiny URLs for all :)

### DIFF
--- a/lektor/utils.py
+++ b/lektor/utils.py
@@ -531,26 +531,39 @@ def bool_from_string(val, default=None):
 
 def make_relative_url(base, target):
     """Returns a relative URL from base to target."""
-    if base == '/':
-        depth = 0
-        prefix = './'
-    else:
-        depth = ('/' + base.strip('/')).count('/')
-        prefix = ''
+    path_dest = posixpath.normpath(posixpath.join(base, target))
+    if target[-1:] == '/' and path_dest[-1:] != '/':
+        path_dest += '/'
+
+    depth = 0
+    if base != '/':
+        path_components = base.strip('/').split('/')
+        depth = len(path_components)
         # If the last part of the base path contains a dot, the page will
         # render into, for example "404.html" instead of "404/index.html".
         # Any relative links from it therefore need one less level of depth.
-        if '.' in base.strip('/').split('/')[-1]:
+        if '.' in path_components[-1]:
             depth -= 1
-            if depth == 0:
-                prefix = './'
 
-    ends_in_slash = target[-1:] == '/'
-    target = posixpath.normpath(posixpath.join(base, target))
-    if ends_in_slash and target[-1:] != '/':
-        target += '/'
+    if depth > 0:
+        path_src_comp = posixpath.normpath(base).lstrip('/').split('/')
+        path_dest_comp = path_dest.lstrip('/').split('/')
+        identical = 0
+        for x, y in zip(path_src_comp, path_dest_comp):
+            if x == y:
+                identical += 1
+            else:
+                break  # quit loop after first mismatch
+        if identical > 0:
+            path_dest = '/' + '/'.join(path_dest_comp[identical:])
+            depth -= identical
 
-    return (prefix + '../' * depth).rstrip('/') + target
+    if depth == 0:
+        if path_dest == '/':
+            return './'
+        return path_dest.lstrip('/')  # use './' syntax regardless?
+    else:
+        return ('../' * depth).rstrip('/') + path_dest
 
 
 def get_structure_hash(params):
--- a/tests/test_builder.py
+++ b/tests/test_builder.py
@@ -155,7 +155,7 @@ def test_basic_template_rendering(pad, builder):
 
     assert '<title>My Website</title>' in rv
     assert '<h1>Welcome</h1>' in rv
-    assert '<link href="./static/style.css" rel="stylesheet">' in rv
+    assert '<link href="static/style.css" rel="stylesheet">' in rv
     assert '<p>Welcome to this pretty nifty website.</p>' in rv
 
 
--- a/tests/test_images.py
+++ b/tests/test_images.py
@@ -145,7 +145,7 @@ def test_thumbnail_dimensions_reported(builder):
         html = f.read()
 
     for t, (w, h) in _THUMBNAILS.items():
-        assert '<img src="./%s" width="%s" height="%s">' % (t, w, h) in html
+        assert '<img src="%s" width="%s" height="%s">' % (t, w, h) in html
 
 
 def test_thumbnail_dimensions_real(builder):
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -75,7 +75,7 @@ def test_markdown_linking(pad, builder):
     prog, _ = builder.build(blog_index)
     with prog.artifacts[0].open('rb') as f:
         assert (
-            b'Look at my <a href="../blog/2015/12/post1/hello.txt">'
+            b'Look at my <a href="2015/12/post1/hello.txt">'
             b'attachment</a>'
         ) in f.read()
 
@@ -84,8 +84,7 @@ def test_markdown_linking(pad, builder):
     prog, _ = builder.build(blog_post)
     with prog.artifacts[0].open('rb') as f:
         assert (
-            b'Look at my <a href="../../../../blog/2015/12/post1/hello.txt">'
-            b'attachment</a>'
+            b'Look at my <a href="hello.txt">attachment</a>.'
         ) in f.read()
 
 
@@ -95,7 +94,7 @@ def test_markdown_images(pad, builder):
     prog, _ = builder.build(blog_index)
     with prog.artifacts[0].open('rb') as f:
         assert (
-            b'This is an image <img src="../blog/2015/12/'
+            b'This is an image <img src="2015/12/'
             b'post1/logo.png" alt="logo">.'
         ) in f.read()
 
@@ -104,8 +103,7 @@ def test_markdown_images(pad, builder):
     prog, _ = builder.build(blog_post)
     with prog.artifacts[0].open('rb') as f:
         assert (
-            b'This is an image <img src="../../../../blog/'
-            b'2015/12/post1/logo.png" alt="logo">.'
+            b'This is an image <img src="logo.png" alt="logo">.'
         ) in f.read()
 
 
--- a/tests/test_urls.py
+++ b/tests/test_urls.py
@@ -21,14 +21,14 @@ def test_basic_url_to_with_alts(pad):
     wolf_de = pad.get('/projects/wolf', alt='de')
     slave_de = pad.get('/projects/slave', alt='de')
 
-    assert wolf_en.url_to(slave_en) == '../../projects/slave/'
-    assert wolf_de.url_to(slave_de) == '../../../de/projects/sklave/'
+    assert wolf_en.url_to(slave_en) == '../slave/'
+    assert wolf_de.url_to(slave_de) == '../sklave/'
     assert slave_en.url_to(slave_de) == '../../de/projects/sklave/'
     assert slave_de.url_to(slave_en) == '../../../projects/slave/'
 
 
 @pytest.mark.parametrize('path, alt, absolute, external, base_url, expected', [
-    ('/projects/slave', '', None, None, None, '../../projects/slave/'),
+    ('/projects/slave', '', None, None, None, '../slave/'),
     ('/projects/slave', 'de', None, None, None, '../../de/projects/sklave/'),
     ('/projects/slave', 'de', True, None, None, '/de/projects/sklave/'),
     ('/projects/slave', 'de', True, True, None, '/de/projects/sklave/'),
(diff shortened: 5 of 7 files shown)


relikd:
Is this the default result to fail for Python 3.x? I can not see how my change should have broken that.

runfalk:
Yes, when pylint adds new checks they break CI. If you look at the test results you can see they are all formatting related.

This PR will eventually fix that: #696

xlotlu:
> Yes, when pylint adds new checks they break CI. If you look at the test results you can see they are all formatting related.

[This](https://travis-ci.org/lektor/lektor/jobs/616150365#L994) doesn't look pylint-related. It breaks all recent PRs.

xlotlu:
@relikd +1, this brings sanity to them urls.

Would you consider implementing it using `pathlib`? It has one glaring omission, there's no `.normalize()` so you'd still need to hit `normpath()`, but I think it would make the code much more readable. Also, [`.relative_to()`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to) and [`.parents`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parents) would make the logic more straightforward.

Big note while I was doing research on this: [the backport](https://pypi.org/pypi/pathlib2/) fails on python 2 on windows with unicode characters. So this might not be desirable.

relikd:
@xlotlu played around a bit, I don't think this will help readability a lot. E.g., `relative_to()` does not work properly for the following example:
```
base: /fr/blog/2015/11/a
target: /fr/blog
```
That would require additional code for handling these special cases. Which in turn, does not help readability. Also, adding a new dependency besides `posixpath`, which is still needed anyway, seems overkill. Plus, the backport issue you mentioned.

But regarding readability,

```suggestion
534:  path_dest = posixpath.normpath(posixpath.join(base, target))
```
could be replaced with `_norm_join(base, target)`, and
```suggestion
549:  path_src_comp = posixpath.normpath(base).lstrip('/').split('/')
```
could be simplified if `base` is guaranteed to be normpath as input.

xlotlu:
> `relative_to()` does not work properly for the following example:
> 
> ```
> base: /fr/blog/2015/11/a
> target: /fr/blog
> ```

I was thinking something along these lines

```python
from pathlib import PurePosixPath
from posixpath import normpath

def get_relpath(source, target):
    """
    Returns the relative path needed to navigate from `source` to `target`.
    """

    source = PurePosixPath(normpath(source))
    target = PurePosixPath(normpath(target))

    # is the source an ancestor of the target?
    try:
        relpath = target.relative_to(source)

    except ValueError:
        # nope, it isn't. how about the other way around?
        try:
            inverse = source.relative_to(target)

        except ValueError:
            # there's no relationship between these two
            srclen = len(source.parts)
            relpath = (
                "/".join([".."] * srclen) # go all the way down
                 / target                 # and then all the way up
            )

        else:
            # yey, it's a descendant. we'll just need to go down.
            distance = len(inverse.parts)
            relpath = "/".join([".."] * distance)

    return relpath
```

```python
>>> get_relpath("fr/blog", "fr/blog/2015/11/a")
PurePosixPath('2015/11/a')
>>> get_relpath("fr/blog/2015/11/a", "fr/blog")
'../../..'
>>> get_relpath("fr/blog/2015/11/a", "en/blog")
PurePosixPath('../../../../../en/blog')
```

This looks quite nice to me. Well, ignoring the different data types returned. I guess it could be made more readable by returning from the `try` and the `else`, and pass-ing inside `except`, to flatten the logic.

> Also, adding a new dependency besides `posixpath`, which is still needed anyway, seems overkill.

`pathlib` is standard library since 3.4. We'd only add the backport as a dependency to support 2.7, something we did before, see the requirements for enum34 and functools32 in [setup.py](https://github.com/lektor/lektor/blob/master/setup.py#L30).

> Plus, the backport issue you mentioned.

Now, that is indeed an issue. I downplayed it given it only affects a small fraction of people: the intersection between python 2.7, windows, system encoding in a specific way, and unicode characters in the path. Also, given that 2.7 support will be phased out anyway, it was worth a shot, but it's a delicate decision anyway. @nixjdm, @runfalk, opinions?

Edit: do we support unicode paths anyway? Is this a non-issue? I remembered the discussion about python-slugify https://github.com/lektor/lektor/pull/604 and this: https://github.com/lektor/lektor/issues/668, which seems to suggest otherwise.

relikd:
18 Tests will fail :o
```
base = '/fr/blog/2015/11/a/', target = '/images/b.svg', expected = '../../../../../images/b.svg', actual = '/images/b.svg'
base = '/fr/blog/2015/11/a.php', target = '/fr/blog/', expected = '../../', actual = '../../..'
base = '/fr/blog/2015/11/a/', target = '/fr/blog/', expected = '../../../', actual = '../../..'
base = '/', target = './a', expected = 'a', actual = '../a'
base = '/', target = './a/', expected = 'a/', actual = '../a'
```
notably the first is an absolute URL and needs to be relative-ized.
The remainder will need special handling code like before.

xlotlu:
> ```
> base = '/fr/blog/2015/11/a/', target = '/images/b.svg', expected = '../../../../../images/b.svg', actual = '/images/b.svg'
> base = '/', target = './a', expected = 'a', actual = '../a'
> base = '/', target = './a/', expected = 'a/', actual = '../a'
> ```

My code is buggy when dealing with absolute urls. Compare that with

```
>>> get_relpath('fr/blog/2015/11/a/', 'images/b.svg')                                        
PurePosixPath('../../../../../images/b.svg')
>>> get_relpath('', './a')                                                                   
PurePosixPath('a')
>>> get_relpath('', './a/')                                                                  
PurePosixPath('a')
```

Regarding that last `target = './a/', expected = 'a/'`, so we want an ending slash in the input to be preserved in the output. I guess that makes sense, to tell directories apart. I think the way to go about it is just a big `if target.endswith("/"): ...` on the first line, before normalizing.

> ```
> base = '/fr/blog/2015/11/a.php', target = '/fr/blog/', expected = '../../', actual = '../../..'
> ```

This is a bug when calculating descendants. The file part should be blown away to begin with.

> ```
> base = '/fr/blog/2015/11/a/', target = '/fr/blog/', expected = '../../../', actual = '../../..'
> ```

This is purely cosmetic, both are valid and they point to the same thing. I guess the `endswith("/")` will take care of it too.

> The remainder will need special handling code like before.

There should be no special handling. If there's special handling there's something wrong with the logic.

I'll look into why the code fails for absolute urls, I never considered that. I guess any mix of relative and absolute should do the right thing.

xlotlu:
@relikd so in principle you like the idea? Does it seem more readable to you?

If I have your ok I'll create that get_relpath() function, make sure it works for all cases, give it some tests, and you can pull it into this PR, or even better another, to preserve this as-is.

relikd:
Sure, go ahead, even on this branch if you like.
Just make sure that under no circumstance absolute URLs are generated.

For example, I currently have a Lektor instance running under `my-server.com/subpath/`. Allowing absolute URLs will kill the navigation in this situation. Redirecting `/a/image.jpg` to `my-server.com/a/image.jpg` instead of the correct `my-server.com/subpath/a/image.jpg`

xlotlu:
@relikd https://github.com/xlotlu/lektor/tree/relative-paths

I was thinking a new PR might be better in case the powers that be decide this approach is better. Saves you some git revert-ing / force pushing, also they're both available for comparison.

I'm not sure what to do about the function name. I like mine better, obviously, but renaming everything existing is meh. I'll leave the decision to you.

relikd:
Have you run the test suite? There were 5 other test classes that need to be updated besides `test_utils`.

xlotlu:
> Have you run the test suite? There were 5 other test classes that need to be updated besides `test_utils`.

I only created a new function and tests for it. Right now it doesn't do anything useful, everything else is as-is.

You have already done the work with whatever other changes need to be made, so I thought you'd pull my commit and either rename my function, or rename the old function, or simply not rename anything and wrap the new function inside the old one.

relikd:
Sorry for the delay, I have not a lot of time right now. But somehow `pathlib`'s `relative_to` does not seem to work properly. Here is my debug log added in between the lines:

```
source: /projects/wolf
target: /projects/slave
src is absolute
tgt is absolute
is the source an ancestor of the target?
no, how about the other way around?
no
return ../../projects/slave
```

relikd:
Further, I did some performance test on both methods, yours and mine:
```
for src, tgt, res in [
    ("/", "./a/", "a/"),
    ("/", "./a", "a"),
    ("/fr/blog/2015/11/a/", "/fr/blog/2015/11/a/a.jpg", "a.jpg"),
    ("/fr/blog/2015/11/a/", "/fr/blog/", "../../../"),
    ("/fr/blog/2015/11/a.php", "/fr/blog/", "../../"),
    ("/fr/blog/2015/11/a/", "/images/b.svg", "../../../../../images/b.svg")
    ]:
    print('A', timeit.timeit(f'xlotlu("{src}", "{tgt}")', "from __main__ import xlotlu", number=10000))
    print('B', timeit.timeit(f'relikd("{src}", "{tgt}")', "from __main__ import relikd", number=10000))
```
Here is the output:
```
A 0.2667111
B 0.08100069500000001
A 0.256260281
B 0.07027780699999997
A 0.686655378
B 0.1920104789999999
A 0.8033440859999998
B 0.18568602499999987
A 0.821497468
B 0.1804938680000001
A 1.0361282459999996
B 0.1735235399999997
[Finished in 4.9s]
```
EDIT: may not change much, but Lektor should be improved in general...
The build time for my 140 recipe page takes more than a minute which I think is infeasible.

xlotlu:
> Sorry for the delay, I have not a lot of time right now.

Sorry for the delay right back, I was busy too, and was kinda waiting for another maintainer to chime in.

> But somehow `pathlib`'s `relative_to` does not seem to work properly.

It's not pathlib's fault, I never handled that scenario (the most common one, d'oh). I fixed the code and added to the tests.

> Further, I did some performance test on both methods, yours and mine:
> ...
> EDIT: may not change much,

It's probably gonna be even worse with the modifications, but I don't think this is reason for worry. Not trying to push my implementation, but I'll personally take readability over performance any day if the overall impact is negligible. (This is also why I was hoping for one of the other guys to get involved).

So, I did a test. Considering a site with 1000 pages, 100 links each, distributed across some of the urls in the tests:

```pycon
>>> stuff = [ 
...     ("/fr/blog/2015/11/a/", "/fr/blog/2015/11/a/a.jpg", "a.jpg"), 
...     ("/fr/blog/2015/11/a/", "/fr/blog/", "../../../"), 
...     ("/fr/blog/2015/11/a.php", "/fr/blog/", "../../"), 
...     ("/fr/blog/2015/11/a/", "/fr/blog/2016/", "../../../2016/"), 
...     ("/fr/blog/2015/11/a/", "/fr/blog/2016/c.jpg", "../../../2016/c.jpg"), 
...     ("/fr/blog/2016/", "/fr/blog/2015/a/", "../2015/a/"), 
...     ("/fr/blog/2016/", "/fr/blog/2015/a/d.jpg", "../2015/a/d.jpg"), 
...     ("/fr/blog/2015/11/a/", "/images/b.svg", "../../../../../images/b.svg"), 
...     ("/fr/blog/", "2015/11/", "2015/11/"), 
... ]                                                                                               
>>> %%time 
... for x in range(int(1000 * 100 / len(stuff))): 
...     for s, t, _ in stuff: 
...         get_relative_path(s, t) 
...                                                                                                 
CPU times: user 3.84 s, sys: 4.1 ms, total: 3.84 s
Wall time: 3.84 s
``` 

> but Lektor should be improved in general...

Yes, the awful performance comes from the query system.

xlotlu:
Devs, it would be nice if 3.2 were released with this fix. Please review this and compare with my alternate PR? @relikd and I had a long talk ^^ about which approach is best.

nixjdm:
Closing this in favor of https://github.com/lektor/lektor/pull/776. Thank you very much for spurring this on and helping to get the final solution @relikd! We probably wouldn't have that now without this.

PR #725, opened 2020-03-04
lint: silence one type of warning and two false positives

yagebu:
### Description of Changes

This should fix the current CI build by disabling one pylint warning and adding comments in two places to silence false positives.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -15,17 +15,15 @@ cache:
 
 environment:
   matrix:
-    # Node 8 - Py 2.7
+    # Node 10 - Py 2.7
     - PYTHON: "C:\\Python27"
-      NODE_JS_VERSION: "8"
+      NODE_JS_VERSION: "10"
     # Node 10 - Py 3.5
     - PYTHON: "C:\\Python35"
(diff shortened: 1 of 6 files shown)


yagebu:
This is ready to be merged IMHO (after CI passes, which it now should).

runfalk:
I've looked at the current state think most things look good. I do however think we should keep testing Python 3.7 with Node 12. Would you mind adding that back?

For anyone else who might see this Node 8 is no longer maintained as per their release page (https://nodejs.org/en/about/releases/).

What I would like before merging this is a squash of your commits with a more descriptive message of what changes that have been performed. I've listed them for your convenience:

* Removed Node 8 and 11 from CI since it's no longer supported
* Add Node 13 to CI
* Fixed linting issues by moving imports from tests to top of file

A big thank you for fixing this :clap:

yagebu:
@runfalk Added 3.7 back in and squashed.

runfalk:
I'll merge this as it is. I think we should switch to GitHub actions and ditch both Appveyor and Travis in the longer run. Though I haven't discussed this with other maintainers.

PR #727, opened 2020-03-04
change test that breaks git configuration

yagebu:
This test was testing a one-line function that basically just called out
to `git config` to get the configured email with a test fixture that
had 3 calls to `git config` and would break the local git config.

Testing some functionality by having a test fixture that just duplicates
that code seems somewhat pointless, even more so when that breaks the
test runners git configuration.

If removed the fixture, instead we now only test that the function
returns a string.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,10 +1,7 @@
 import os
 import shutil
-import subprocess
 import textwrap
 import pytest
-
-from lektor.utils import is_windows
 
 
 @pytest.fixture(scope='function')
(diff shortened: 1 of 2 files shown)


runfalk:
I agree that the current test fixture takes the wrong approach here. A fixture should really avoid side effects like this.

What we could do instead is to monkey patch subprocess.Popen and use a mock. I'm normally not a huge fan of monkey patching based on implementation details, but in this case we can at least ensure that it tried to run the command we expect and that is uses the stripped return value.

yagebu:
I don't really see the point in that though. That would be testing that the code is implemented exactly as it is now - that's not a test IMHO

runfalk:
I guess, I don't really have a strong opinion on removing this test without replacing it.

I can merge this as it is.

runfalk:
When the tests pass, please squash your commits I'll merge them tomorrow.

blueyed:
@yagebu 
> would break the local git config

Just out of curiosity: how did it break the local config?  It should restore the old value at least after all, but might be setting something where nothing was set?
(just stumbled upon this and wondered about it)

runfalk:
Correct me if I'm wrong @yagebu, I think the main issue here is that if the tests are killed at the wrong time they may leave your Git config in an inconsistent state. Also there will be a brief moment where your e-mail is wrong and if you happen to do a commit at that time you may end up with the wrong e-mail address. So in majority of cases the current test case will work fine, but it's very confusing if one of the above cases ever happen.

I think the correct way of solving this would be to create a new temporary git repository and set the e-mail address there.

yagebu:
> Just out of curiosity: how did it break the local config? It should restore the old value at least after all, but might be setting something where nothing was set?

@blueyed: It would set an empty value where no value was set before.

> Correct me if I'm wrong @yagebu, I think the main issue here is that if the tests are killed at the wrong time they may leave your Git config in an inconsistent state.

It was actually worse than that. The tests would run through successfully and after that it would set the `user.email` to empty in the local git config (in contrast to not having any entry in the local git config). That would override my globally configured email so that I'd end up with commits without my email address.

PR #728, opened 2020-03-04
Npm dependencies

yagebu:
### Related Issues / Links

Ref #449

In this PR, I'll try to get the npm dependencies up to date.

Currently we get the following nice line when installing the dependencies...:

    found 613 vulnerabilities (226 low, 25 moderate, 360 high, 2 critical)


lektor/admin/static/js/widgets/flowWidget.jsx
```diff
@@ -263,7 +263,7 @@ const FlowWidget = createReactClass({
               className='btn btn-default btn-xs'
               title={this.props.value[idx].collapsed ? i18n.trans('Expand') : i18n.trans('Collapse')}
               onClick={this.props.value[idx].collapsed
-                       ? this.expandBlock.bind(this, idx) : this.collapseBlock.bind(this, idx)}>
+                ? this.expandBlock.bind(this, idx) : this.collapseBlock.bind(this, idx)}>
```
runfalk, line 266: May I suggest breaking this up instead?

```js
const isCollapsed = this.props.value[idx].collapsed;
const onClickHandler = isCollapsed ? this.expandBlock : this.collapseBlock;
...
title={isCollapsed ? i18n.trans('Expand') : i18n.trans('Collapse')}
onClick={onClickHandler.bind(this, idx)}
```

yagebu: This change was made by `standard --fix` - I didn't really want to do any non-trivial code changes in this PR ;)

runfalk: Fair enough, I figured that was the case :)

runfalk:
I'll try to test this tonight. I think the changes themselves mostly look OK.

lektor/admin/static/webpack.config.js
```diff
@@ -22,41 +23,54 @@ module.exports = {
...
+      /*
+      cacheGroups: {
+        styles: {
+          name: 'styles',
+          test: /\.(le|c)ss$/,
+          chunks: 'all',
+          enforce: true
+        }
+      }
+      */
```
runfalk: I think we should either remove or use this.

lektor/admin/static/webpack.config.js
```diff
@@ -22,41 +23,54 @@ module.exports = {
     filename: '[name].js'
   },
   devtool: 'source-map',
+  optimization: {
+    splitChunks: {
+      chunks: 'all',
+      name: 'vendor'
```
runfalk, line 28: Is there a reason why we need two CSS files now instead of one like before?

lektor/admin/static/webpack.config.js
```diff
@@ -1,8 +1,9 @@
 var webpack = require('webpack')
 var path = require('path')
```
runfalk: Feel free to update these to const as well.

runfalk (changes requested):

lektor/admin/static/webpack.config.js
yagebu: Sorry, this PR is not quite done, I should have been a bit clearer in the description. But you're right of course :)

runfalk: It is possible to mark it as a draft if that's the case :)

lektor/admin/static/webpack.config.js
yagebu: Well I tried to bundle the CSS into one with the commented part. I assume it's still possible with Webpack 4, but I don't see the point. If we split the JS into two, it makes just as much sense to also split the CSS.

lektor/admin/static/webpack.config.js
yagebu: Will do

yagebu:
I'd call this ready. This does now contain minor version bumps for the dependencies in the frontend and bumps all development dependencies to the latest version. I'll try to update the frontend dependencies in a separate PR to keep things manageable.

lektor/admin/static/webpack.config.js
runfalk: Fair enough, I don't mind this.

runfalk (approved):

runfalk:
Thank you!

PR #734, opened 2020-03-05
PR: Move to github actions

### LINKED ISSUE #733: Switching to GitHub actions
Currently we use both Travis and Appveyor. I think we should consider moving to GitHub actions which is completely free for all open source projects. I also see this as an opportunity to optimize the CI process a bit. The current one has some problems (not all are related to Travis and Appveyor):

* JS is built for all Python versions even though we never use it for the tests
* Windows tests are slow because they can't run in parallel
* We have no macOS tests, which would probably be a requirement to start building dmg files again
* Linting is run for every single Python version which is completely unnecessary and takes some time

My suggestion is to split the build into many different stages where the first stage runs lints, the second builds JS and runs tests using different Node versions and the last step would run different Python version on different platforms.

Anyone who have negative experiences with GitHub actions or are concerned about moving to it?

goanpeca:
## Todo
- [x] Add caching for pip -> https://github.com/actions/cache/blob/master/examples.md#python---pip 
- [x] Add caching for node

### Issue(s) Resolved

Fixes #733


goanpeca:
We have a JS lint and a Python lint job that start in parallel. All the node tests will only start if the js lint job passes. All the Python tests will only run if the python lint passes.

Removed the pytest-pylint dependency. Added only py3 tests, but I could also add 2.7 if we plan to support that? But I suggest we just move on with 3.x support only.

Need to add the coverage. I could create a coverage per job and upload the artifacts to a final job that   could combined them and then upload the combined coverage.

Thoughts?

runfalk:
I assume we want to release Lektor 3.2 with support for Python 2.7. After that I think it's OK to break compatibility. Python 2.7 isn't EOL yet and macOS which still ship with it AFAIK. For now I think it should stay.

I think you should update the Makefile and add `pylint` to `test-python`. Maybe you could also take the opportunity to change from the archaic `py.test` to `pytest` as well.

I also think we should take the opportunity to produce wheels as build artifacts. This should probably only apply to new tags (which CI also should run for). Later we can make this automatically publish to PyPI.

I also think there is some way to save the package cache between runs. Maybe we should consider doing that to make the runs even speedier. (https://github.com/actions/cache/blob/master/examples.md#python---pip)

Great work :clap:

runfalk (changes requested):
Apparently I need to write something here to request changes

xlotlu:
> I also think we should take the opportunity to produce wheels as build artifacts. This should probably only apply to new tags (which CI also should run for). Later we can make this automatically publish to PyPI.

Maybe leave this aside as a future issue?
 
> I also think there is some way to save the package cache between runs. Maybe we should consider doing that to make the runs even speedier. (https://github.com/actions/cache/blob/master/examples.md#python---pip)

Unless it's really trivial, leave this for a future commit as well?

> Great work clap

Indeed.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,296 @@
...
+          python -m pip list
+      - name: Run python lint
+        run: pylint lektor
+      - name: Cache pip
+        uses: actions/cache@v1
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/setup.py') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
```
runfalk: I think these go before the installation of dependencies, otherwise the cache won't be restored before it's used.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,296 @@
...
+        run: |
+          python --version
+          python -m pip list
+      - name: Run python lint
+        run: pylint lektor
+      - name: Cache pip
+        uses: actions/cache@v1
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/setup.py') }}
```
runfalk: I think Python version and OS must be part of this string, otherwise it'll try to reuse the same cache across Python versions.

runfalk (changes requested):

.github/workflows/ci.yml
```diff
@@ -34,20 +41,20 @@ jobs:
...
   lint-js:
     name: Lint JS
     runs-on: ubuntu-latest
     steps:
       - name: Checkout branch
         uses: actions/checkout@v2
+      - name: Cache node
+        uses: actions/cache@v1
+        with:
+          path: ~/.npm
```
runfalk, line 52: I think you can even go as far as saving node_modules itself.

.github/workflows/ci.yml
xlotlu: > I think Python version and OS must be part of this string

The OS yes, but the Python version?

> otherwise it'll try to reuse the same cache across Python versions.

Isn't that the point?

runfalk: OS is already used I see, but we still need to consider Python or Node version

.github/workflows/ci.yml
goanpeca: Fixed

.github/workflows/ci.yml
goanpeca: Fixed. Added py version. Some packages might not be universal wheels...

.github/workflows/ci.yml
goanpeca: Do I need to add the node version to the cache key as well @runfalk ?

runfalk: I'm not sure but I think it's safest to do so.

goanpeca:
>I also think we should take the opportunity to produce wheels as build artifacts. This should probably only apply to new tags (which CI also should run for). Later we can make this automatically publish to PyPI.

I would leave this for a future PR

runfalk:
I'm fine with leaving that for a future PR. This is a lot better than what we have currently.

.github/workflows/ci.yml
goanpeca: Fixed

.github/workflows/ci.yml
```diff
@@ -0,0 +1,324 @@
...
+    name: Lint JS
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v2
+      - name: Cache node
+        uses: actions/cache@v1
+        with:
+          path: ~/.npm
+          key: linux-node13-lint-${{ hashFiles('**/lektor/admin/package-lock.json') }}
```
runfalk: I think it's fine to drop the `-lint` part. That way the linux-node tests can re-use the same cache.

Also `node13` -> `node-13` to ensure other jobs can re-use it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,324 @@
...
+        uses: actions/cache@v1
+        with:
+          path: ~\AppData\Roaming\npm-cache
+          key: ${{ runner.os }}-node-${{ matrix.NODEJS_VERSION }}-${{ hashFiles('**\lektor\admin\package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-${{ matrix.NODEJS_VERSION }}-
+      - name: Cache node modules
+        uses: actions/cache@v1
+        with:
+          path: /lektor/admin/node_modules
```
runfalk: Will forwardslashes really work here?

goanpeca: I will leave the node_modules cache for a separate PR as well. I am not really sure about it

.github/workflows/ci.yml
goanpeca: Fixed!

goanpeca:
@runfalk @xlotlu missing the coverage. There is no straightforward way to combine all the different generated files so for now I will use python3.8 plus linux as the coverage. 

Thoughts?

runfalk:
I'm fine with using only Linux Python 3.8 for now, but I think it's solvable using: https://coverage.readthedocs.io/en/coverage-5.0.3/cmd.html#combining-data-files and https://help.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts

If you don't fix it in this PR I think we should open an issue for it.

goanpeca:
Ok I think this is good enough for now :-) @runfalk & @xlotlu

goanpeca:
@nixjdm I need the codecov repository upload token :-p (Github actions does not fetch it automatically like travis does)

goanpeca:
Nevermind, I have access!

runfalk (changes requested):
One last thing. I think the `restore-keys` aren't necessary since you never create a key that matches it. From the documentation:

> `restore-keys` - An ordered list of keys to use for restoring the cache if no cache hit occurred for key

.github/workflows/ci.yml
```diff
@@ -0,0 +1,285 @@
...
+          python -m pip install codecov
+      - name: Show python environment
+        run: |
+          python --version
+          python -m pip list
+      - name: Run python tests
+        run: pytest . --tb=long -svv --cov=lektor
+      - name: Publish coverage
+        if: matrix.PYTHON_VERSION == '3.8'
+        run: codecov -t 39974034-91d8-48d1-9698-de48e0667a09
```
runfalk, line 281: I don't think we should expose the token like this unless we really have to. There are ways to set environment variable in a more secret way. (https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets)

If we keep this here I think forks will report their coverage centrally as well.

goanpeca: Secrets are not passed to PRs from forks AFAIK

https://github.community/t5/GitHub-Actions/Make-secrets-available-to-builds-of-forks/td-p/30678

Besides a coverage token is a very non risky thing to expose anyway 🤷‍♂

.github/workflows/ci.yml
```diff
@@ -0,0 +1,285 @@
...
+    name: Linux node${{ matrix.NODEJS_VERSION }} tests
+    runs-on: ubuntu-latest
+    needs: lint-js
+    env:
+      CI: True
+      NODEJS_VERSION: ${{ matrix.NODEJS_VERSION }}
+    strategy:
+      fail-fast: false 
+      matrix:
+        NODEJS_VERSION: ['10', '12', '13']
```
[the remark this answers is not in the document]
goanpeca: It runs in parallel containers so it makes no difference in time

goanpeca:
>@goanpeca: Is there a reason for dropping Python 3.5?

I see no point in supporting such an "old" version, there is 3.6, 3.7 and 3.8 now to pick from.

yagebu:
> I see no point in supporting such an "old" version, there is 3.6, 3.7 and 3.8 now to pick from.

I somewhat agree, but it's less EOL than 2.7, so I'd drop support for them at the same time.

.github/workflows/ci.yml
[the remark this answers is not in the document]
yagebu: But there's really no point to running them more than once. It still wastes resources and introduces a mental overhead for any contributor who see 20 checks of which 9 are the same thing.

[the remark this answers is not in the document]
runfalk: I think there is a point to run them all. Since all are still supported we may have contributors who use them. There are differences between them and sometimes one of them may break. I really think we should test for all supported releases here.

I do however think it's OK to only test latest for macOS and Windows and all on Linux. That way we reduce the amount of build jobs but can still catch version inconsistencies.

Though I don't really mind keeping all the tests either.

runfalk:
I think we should have tests for all maintained versions of Python (https://devguide.python.org/#status-of-python-branches). Though I'm fine with just testing latest on windows and latest + 2.7 on macOS.

runfalk:
I looked a bit att codecov @goanpeca and it seems like it should automatically merge multiple coverage reports. This means that we can probably just att the codecov upload at the end of each job.

goanpeca:
For python I suggest we test on linux 2.7, 3.5, 3.6, 3.7, 3.8 (the fastest container)
For osx 2.7, 3.5 and 3.8
For win 2.7, 3.5 and 3.8 

This way we test all on a fast machine and test both the oldest and the newest plus 2.7 on other systems. 

For node versions yo be honest I do not know much to have an opinion so whatever you guys agree on is fine.

goanpeca:
I suggest we make a 3.x release with as many of the PRs with support for 2.7 and then leave master (which will become 4.x with support for python 3 only.

runfalk:
My suggestion would be.

```
Linting
  Linux with Node 13
  Linux with Python 3.8

Python
  Linux with 2.7, 3.5, 3.6, 3.7, 3.8
  macOS with 2.7 and 3.8
  Windows 3.8

Node
  Linux with 10, 12 and 13
  macOS with 13
  Windows with 13
```

We mostly want to test for version differences and I think it's acceptable to skip that on Windows and macOS to make builds faster. If we ever run into a bug because of it I think we should reconsider and add it back.

I'd like keeping the linting separate from the rest of the tests for to make it easier to see if the commit failed on linting or failing tests. I think the extra cost of 1 minute per lint is worth it.

I also really think we should test all node versions since slower moving distributions probably have not updated to the latest version.

goanpeca:
I disagree with your proposal @runfalk for python. Usually the older python versions are the ones that cause trouble so if the plan is to keep supporting python 2.7 and 3.5 then we should ALWAYS test that on all systems, specially on windows where a lot of problems occur. Macos and Linux are very similar and usually not testing things on mac is ok. Not testing things on window is a big mistake.

```
Linting
  Linux with Node 13. # Already like this
  Linux with Python 3.8. # Already like this

Python
  Linux with 2.7, 3.5, 3.6, 3.7, 3.8.  # Test all
  macOS with 2.7, 3.5 and 3.8 # Test "corner" cases
  Windows with 2.7, 3.5 and 3.8 # Test "corner" cases

Node   # I have no opinion on this...
  Linux with 10, 12 and 13
  macOS with 13
  Windows with 13
```

In general we will always be limited by the older versions (2.7 and 3.5) , not the newer ones and newer versions might include subtle changes, hence why testing 3.8 makes sense.

goanpeca:
>I looked a bit att codecov @goanpeca and it seems like it should automatically merge multiple coverage reports. This means that we can probably just att the codecov upload at the end of each job.

I do not think this is the case. And I do not want to delay the PR much longer, so I think coverage can be updated later, for now it checks py38 on linux.

goanpeca:
>I'd like keeping the linting separate from the rest of the tests for to make it easier to see if the commit failed on linting or failing tests. I think the extra cost of 1 minute per lint is worth it.

Agree, there is no point in running things if linting is not ok. I would argue that running linux tests for let's say py3.8 could also serve as a smoke tests for simple errors in test writing so that if an error occurs (typo, failed import) we could catch it sooner and not launch all the 10+ something containers that run the rest.

The current setup is, run lint tests for python and js first. If Js lint tests pass, the node "tests"(we are not really testing, just installing) are launched. If python lint tests pass all the python tests are launched.

runfalk:
I'm fine with testing more versions for both Windows and macOS.

Regarding the automatic merging of codecov reports: https://github.com/codecov/codecov-node/issues/87

I don't necessarily think linting should gate the other tests from starting. There is value in letting them finish even though lints are not successful. That way we can see that a PR has passing test but the linting is not correct yet. I as a reviewer would at least find that useful.

goanpeca:
>I don't necessarily think linting should gate the other tests from starting. There is value in letting them finish even though lints are not successful. That way we can see that a PR has passing test but the linting is not correct yet. I as a reviewer would at least find that useful.

I as a reviewer do not want to "waste" time on a PR if it has not passed the linting tests. Also why I suggested running only 1 of the linux tests (let's say for py38) in tandem with lints so the user can still catch silly test errors (missing imports, syntax errors) but not waste containers by running ALL tests if there is a clear problem on all of them. This way that tests serves as a smoke test. This is specially useful for sprints or when a project receives many PRs in a short period of time, plus is less wasteful with github resources (even if they are "for free").

nixjdm:
The issue I think is that if the build lints first, and stops on a lint error, it is very likely that occasionally a linter will update and fail all builds until old code is fixed to appease the linter. Updated linters breaking on old code is something we've seen repeatedly, so I think it's likely to happen.

Maybe we can swap the order and run pytest first, and then lint? Then I could still choose to not waste my time on a broken build - or if I have the time, I could see that the PR is actually fine and the linter updating has caused the failure.

Pinning the linter doesn't seem like a great idea either.

runfalk:
The problem with gating things behind the linter is that we end up with PRs fixing totally unrelated linting problems (if there has been a new version released) only to pass the CI loop. This is why I think we should consider pinning the linter version.

I think it's beyond the scope of this PR but we could run a separate job that checks bi-weekly or something if master is broken with the latest release of pylint (https://help.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events-schedule combined with https://github.com/JasonEtco/create-an-issue). This way master will never break due to pylint, PRs won't have to fix unrelated commits, we can gate tests behind pylint and still not be unaware of new potential pylint bugs.

runfalk:
Sorry @goanpeca I missread your comment about Linux 3.8 as a smoke test. I think that strikes a pretty good balance.

.github/workflows/ci.yml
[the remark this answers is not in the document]
yagebu: > I think there is a point to run them all. Since all are still supported we may have contributors who use them. There are differences between them and sometimes one of them may break.

The way I see it, what we're testing with these runs is whether Webpack and Babel work on all these versions of Node. Which seems a bit overkill for me, since those projects already test for that. But if these builds turn out to be non-flaky there's not much harm either :)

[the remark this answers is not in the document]
runfalk: I see what you mean. What this protect us from is if someone makes a PR that updates a dependency that breaks some Node version we still want to support. JS projects are infamous for breaking backwards compatibility so I think we should still consider this.

In the future I think we can be a bit more clever and only run JS tests if the commit touches JS files and vice versa.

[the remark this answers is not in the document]
goanpeca: Agreed, we need JS tests in the first place :-p

runfalk (approved):
I'm fine with this as it is right now.

goanpeca:
@runfalk thanks for the comments, let me make one final push to add the smoke test as discussed, and we are good to go!

xlotlu:
> The problem with gating things behind the linter is that we end up with PRs fixing totally unrelated linting problems (if there has been a new version released) only to pass the CI loop. This is why I think we should consider pinning the linter version.

Yes, the current situation is absurd. The linter really should be pinned, and figure out later the proper workflow to deal with updating it.

As for gating tests and linting, my personal preference is we shouldn't.

I'm thinking here from the point of view of a contributor: if I submit a change I'd like to see right away if it passes the tests, and also see if it might be rejected due to linting errors. Getting a failure because of some silly style violation is just meh, if after fixing said error I then get a test error, and need to move on to fixing the real issue. The other way around is pretty annoying as well. If you think about it, it's the equivalent of exiting the tests on first failure. While early exit makes sense in development, when pushing a commit to an existing PR, all failures should be visible, not do incremental development.

Yes, if any of these fails on latest py / linux don't run anything else, but both tests and linting are valuable, non-gated.

P.S.: I didn't really participate in this one, but I just wanted to say great work guys.

goanpeca:
Updated the PR, so lint and smoke tests are run first, and the next tests depend only on the smoke tests (not on the lint ones).

Screen Shot 2020-03-07 at 21 20 57

setup.py
```diff
@@ -55,11 +55,12 @@
         'License :: OSI Approved :: BSD License',
         'Operating System :: OS Independent',
         'Programming Language :: Python',
+        'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.5',
```
runfalk, line 60: This is removed but that's one of the versions we are testing.

goanpeca: Oops, will fix!

goanpeca: Fixed!

runfalk:
Is this waiting for anything or should we just merge it? It looks good to me.

goanpeca:
>Is this waiting for anything or should we just merge it? It looks good to me.

It is fine on my side 🤷‍♂

nixjdm:
Thank you @goanpeca and @runfalk for spearheading this. Fantastic work!

PR #739, opened 2020-03-08
Upgrade react-router

yagebu:
### Related Issues / Links

Ref #449 

### Description of Changes

This upgrades react-router to from version 3 to 5. This introduces quite a few changes to how the `<Route>` components work. Instead of basically just registering routes, they now do the actual rendering. Also, there's now the `<Prompt>` component to prevent navigation.

Also use function components or components only inheriting from `React.Component` in some places.


runfalk:
Very nice. This is something I've wanted to do for a long time.

I'll try to get this reviewed during next week.

runfalk:
Sorry for the delay in handling this. It looks great! :tada:

PR #744, opened 2020-03-27
Replace RobotoSlab font with full UTF-8 version #743

### LINKED ISSUE #743: Replace RobotoSlab font with full UTF-8 version
Current RobotoSlab version doesn't contain cyrillic symbols - so mixed character texts look ugly.

Please consider replacing font with full UTF-8 version. It's available at Google:

https://fonts.google.com/specimen/Roboto+Slab

barser:
### Issue(s) Resolved

Fixes #743 

### Description of Changes

* UTF-8 version of Roboto Slab Fonts (with cyrillic characters support)


runfalk:
I definitely think we should upgrade this to the full UTF-8 version. I do however think it may be better to use https://www.npmjs.com/package/@openfonts/roboto-slab_all instead of vendoring the dependency. That way we avoid increasing the repository size if we ever need to update the font again.

I think you should just be able to add this as a dependency to `package.json`, update `webpack.config.js` to copy the fonts and add a `require('@openfonts/roboto-slab_all')` in the JS code.

xlotlu:
Closing this in favour of #775

PR #748, opened 2020-04-12
PR: Add checks to skip builds and cancel previous builds

goanpeca:
Ok, this adds conditional running of builds.

Used paths based on @runfalk comment. 

I also added support for `[ci skip]` and `[skip ci]` 

The regex might need some additions, for package.json etc...

Thoughts @runfalk  @xlotlu ?

### DIFF
--- a/.github/workflows/js_tests.yml
+++ b/.github/workflows/js_tests.yml
@@ -1,0 +1,272 @@
+name: JS tests master
+
+on:
+  push:
+    branches:
+      - master
+      - 3.x-maintenance
+    paths:
+      - '.github/scripts/*.sh'
+      - '.github/workflows/*.yml'
+      - 'gui/package.json'
+      - 'requirements/*.txt'
+      - 'Makefile'
+      - 'MANIFEST.in'
+      - '**.bat'
+      - '**.jsx?'
+      - '**.sh'
+
+  pull_request:
+    branches:
+      - master
+      - 3.x-maintenance
+    paths:
+      - '.github/scripts/*.sh'
+      - '.github/workflows/*.yml'
+      - 'gui/package.json'
+      - 'requirements/*.txt'
+      - 'MANIFEST.in'
+      - '**.bat'
+      - '**.jsx?'
+      - '**.sh'
+
+jobs:
+  ############################################################################
+  # Close previous builds on the same branch
+  ############################################################################
+  cleanup-previous-runs:
+    name: Cleanup previous builds
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo $GITHUB_EVENT_NAME
+      - uses: rokroskar/workflow-run-cleanup-action@v0.2.2
+        if: github.event_name == 'pull_request'
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+
+  ############################################################################
+  # Lint jobs
+  ############################################################################
+  lint-js:
+    name: Lint JS
+    needs: cleanup-previous-runs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Fetch branches
+        run: git fetch --prune --unshallow
+      - name: Check build skips
+        shell: bash -l {0}
+        run: source .github/scripts/check_skip.sh
+      - name: Cache node
+        if: env.RUN_BUILD == 'true'
+        uses: actions/cache@v1
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-13-${{ hashFiles('**/lektor/admin/package-lock.json') }}
+      - name: Setup node
+        if: env.RUN_BUILD == 'true'
+        uses: actions/setup-node@v1
+        with:
+          node-version: '13'
+      - name: Install node dependencies
+        if: env.RUN_BUILD == 'true'
+        run: |
+          cd lektor/admin/
+          npm install .
+          npm run webpack
+      - name: Show node environment
+        if: env.RUN_BUILD == 'true'
+        run: |
+          node --version
+          npm --version
+      - name: Run JS lint
+        if: env.RUN_BUILD == 'true'
+        run: |
+          cd lektor/admin/
+          npm run lint
+
+  ############################################################################
+  # Smoke tests
+  ############################################################################
+  linux-smoke-node:
+    name: Linux node${{ matrix.NODEJS_VERSION }} smoke test
+    needs: cleanup-previous-runs
+    runs-on: ubuntu-latest
+    env:
+      CI: True
+      NODEJS_VERSION: ${{ matrix.NODEJS_VERSION }}
+    strategy:
+      fail-fast: false 
+      matrix:
+        NODEJS_VERSION: ['13']
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Fetch branches
+        run: git fetch --prune --unshallow
+      - name: Check build skips
+        shell: bash -l {0}
+        run: source .github/scripts/check_skip.sh
+      - name: Cache node
+        if: env.RUN_BUILD == 'true'
+        uses: actions/cache@v1
+        with:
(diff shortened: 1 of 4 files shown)


runfalk:
Wouldn't it be simpler to split things into two workflows (GitHub Actions YAML files) and use the [paths filter](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths)? I'm thinking one workflow for JS and one for Python. That should also allow us to skip the check on every job I suppose?

runfalk:
Is canceling the previous job something that we need to do because of races, or are we just trying to be good citizens and save compute power?

goanpeca:
>Wouldn't it be simpler to split things into two workflows (GitHub Actions YAML files) and use the paths filter?

Yep, that indeed sounds like a better approach, I missed that.

>That should also allow us to skip the check on every job I suppose?

yep

>Is canceling the previous job something that we need to do because of races, or are we just trying to be good citizens and save compute power?

Both.  If we do not cancel them new builds with stay queued.

No more time to work on this atm, so I may take a look at it later on. Splitting sounds like a good idea, just for easier reading of the files (too much info in the ci.yml)

goanpeca:
@runfalk made the changes. I still left the script for [ci skip] as I think that could still be useful.

Any other suggestions to the paths to check?

.github/workflows/js_tests.yml
```diff
@@ -0,0 +1,272 @@
...
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Fetch branches
+        run: git fetch --prune --unshallow
+      - name: Check build skips
+        shell: bash -l {0}
+        run: source .github/scripts/check_skip.sh
+      - name: Cache node
+        if: env.RUN_BUILD == 'true'
```
runfalk, line 65: Can this work instead?

https://github.com/srz-zumix/ci-skip/blob/master/.github/workflows/main.yml

.github/workflows/js_tests.yml
```diff
@@ -0,0 +1,272 @@
...
+  ############################################################################
+  lint-js:
+    name: Lint JS
+    needs: cleanup-previous-runs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
runfalk, line 58: Will this be set when it is not a PR? Will it now checkout the correct commit without this ref spec?

.github/workflows/js_tests.yml
```diff
@@ -0,0 +1,272 @@
...
+  ############################################################################
+  cleanup-previous-runs:
+    name: Cleanup previous builds
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo $GITHUB_EVENT_NAME
+      - uses: rokroskar/workflow-run-cleanup-action@v0.2.2
+        if: github.event_name == 'pull_request'
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
```
runfalk, line 45: Will this even be available when pull requests run? If it is, is there any risk this can leak?

.github/workflows/py_tests.yml
```diff
@@ -1,111 +1,98 @@
...
+    paths:
+      - '.github/scripts/*.sh'
+      - '.github/workflows/*.yml'
+      - 'requirements/*.txt'
+      - 'Makefile'
+      - 'MANIFEST.in'
+      - 'pylintrc'
+      - '**.bat'
+      - '**.py'
+      - '**.sh'
```
runfalk, line 17: Maybe setup.cfg and pytest.ini as well.

.github/workflows/js_tests.yml
```diff
@@ -0,0 +1,272 @@
...
+
+on:
+  push:
+    branches:
+      - master
+      - 3.x-maintenance
+    paths:
+      - '.github/scripts/*.sh'
+      - '.github/workflows/*.yml'
+      - 'gui/package.json'
```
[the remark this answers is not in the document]
runfalk: *.less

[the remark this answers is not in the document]
runfalk: *.ttf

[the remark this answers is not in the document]
runfalk: *.json

runfalk (changes requested):
I don't really like the concept of `[skip ci]` as it clutters the YAML files, and I don't really see the reason to have it at all. However, it seems to be supported by other maintainers so I won't reject the PR based on it.

I'm a bit uncomfortable with the build cancelling since it has access to a secret token. I really want to be sure there is no risk that this leaks if we're going to have it. Now that we skip a lot of steps unless you change both Python and JS in the same commit, is this still necessary?

.github/workflows/js_tests.yml
goanpeca: Not for prs

.github/workflows/js_tests.yml
goanpeca: Dont know

.github/workflows/js_tests.yml
goanpeca: The ref in a Pr points to the merge commit this just makes it point to the actual last commit. Nothing will not eork because of this

.github/workflows/js_tests.yml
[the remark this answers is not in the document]
goanpeca: > run gui/package.json in CI since it has been broken for ages.

Is that worth fixing?

Why are any of the other extensions relevant? we are not even running tests at this point on JS

.github/workflows/js_tests.yml
goanpeca: I could add a conditional if you prefer to be sure 🤷

.github/workflows/js_tests.yml
goanpeca: https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets

.github/workflows/js_tests.yml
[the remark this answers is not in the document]
runfalk: The json things are included by webpack, i.e. they change stuff. Especially if a .json is removed builds can break.

I didn't realise we didn't run the JS tests. But they exist and we probably should: https://github.com/lektor/lektor/blob/master/Makefile#L32

I think we should probably scrap `gui/` and rebuild it from scratch. It's based on an ancient version of Electron.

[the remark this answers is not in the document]
runfalk: The same as json goes for less files.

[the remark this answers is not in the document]
goanpeca: Got it.

.github/workflows/js_tests.yml
runfalk: I just want to know it works :laughing:. If you say it does, sure let's keep it.

.github/workflows/js_tests.yml
runfalk: I'm not sure how to interpret that actually. It doesn't explicitly say anything as far as I can tell. I tried googling yesterday as well without success.

The attack vector I want  to know is: can someone make a PR that adds/modifies a workflow and then snag the access token?

goanpeca: https://github.community/t5/GitHub-Actions/Token-permissions-for-forks-once-again/td-p/33839

from forks I think not...?

runfalk: If it's not run for forks, then this only works when one of us maintainers does a PR directly on the core lektor repository I guess? Is that case worth optimizing for?

.github/workflows/js_tests.yml
goanpeca: Nope actually it does not work, will make the push for the fix later today.

.github/workflows/js_tests.yml
goanpeca: >If it's not run for forks, then this only works when one of us maintainers does a PR directly on the core lektor repository I guess? Is that case worth optimizing for?

Good question I need to check this

yagebu:
This has gotten pretty stale, so I'm closing this PR. Feel free to reopen if you want to continue working on this :)

PR #755, opened 2020-04-18
Prevent fsync storm during update_all_source_infos

dairiki:
This applies the fix for #144 that was applied in 701a37c8 to [Builder.build_all](https://github.com/lektor/lektor/blob/master/lektor/builder.py#L1128) to [Builder.update_all_source_infos](https://github.com/lektor/lektor/blob/master/lektor/builder.py#L1151).

### Issue(s) Resolved

This resolves performance issues related to fsync-ing by sqlite during the "update all source info" phase of the initialization of the dev server.  I.e. a “_source info update_” is performed at the beginning of each `lektor server` execution.

On my old machine, working on a lektor site with about 100 pages, this change reduces the time for the initial _source info update_ to complete from 17 seconds (with lots of audible disk thrashing) to under 2 seconds.

### Related Issues / Links

See #144.

### Description of Changes

Hold a dummy database connection open during the source info update. This seems to prevent fsync-ing during the source info updating.  I’ll admit that I haven’t fully determined how or why this works, but it’s the same fix applied in 701a37c8 to [Builder.build_all](https://github.com/lektor/lektor/commit/701a37c838321c1eacfb9e27f197dc40e5ef4746#diff-9a190466ef89d08b8977fd667bd08e13R1097).

### DIFF
--- a/lektor/builder.py
+++ b/lektor/builder.py
@@ -1152,13 +1152,19 @@ class Builder(object):
         """Fast way to update all source infos without having to build
         everything.
         """
-        with reporter.build('source info update', self):
-            with self.new_build_state() as build_state:
-                to_build = self.get_initial_build_queue()
-                while to_build:
-                    source = to_build.popleft()
-                    with reporter.process_source(source):
-                        prog = self.get_build_program(source, build_state)
-                        self.update_source_info(prog, build_state)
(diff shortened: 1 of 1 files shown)


xlotlu:
thank you

dairiki:
Thank you!

PR #758, opened 2020-04-23
Improve edit page

### LINKED ISSUE #179: Values get saved even when validation fails
Tested for Integer and Float, in both cases it is possible to save invalid values from admin ui to contents.lr even if the validation failed because onChange gets called before the validation.
Need to check for other types.

Also the float validation should be reworked because of the limitations of the builtin `parseFloat`.

yagebu:
### Issue(s) Resolved

Fixes #179 

### Description of Changes

The first few commits consist mostly of basic refactorings of the widgets code (the editing UI components), simplifying and documentings parts and turning some functions that were basically just composing UI components into simple React components.

The last few commits address #179 by disallowing saving of invalid values and using a regex for float parsing instead of `parseFloat`.

### DIFF
--- a/lektor/admin/static/js/events.jsx
+++ b/lektor/admin/static/js/events.jsx
@@ -14,14 +14,14 @@ Event.getEventType = function () {
   return this.name
 }
 
-class RecordEvent extends Event {
+export class RecordEvent extends Event {
   constructor (options) {
     super(options = options || {})
     this.recordPath = options.recordPath
   }
 }
 
-class AttachmentsChangedEvent extends RecordEvent {
+export class AttachmentsChangedEvent extends RecordEvent {
   constructor (options) {
     super(options = options || {})
     this.attachmentsAdded = options.attachmentsAdded || []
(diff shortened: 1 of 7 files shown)


yagebu:
Added two more commits, one removing an unused dependency (querystring), another for an `npm update`

goanpeca:
@yagebu thanks for working on this. Could you post a screenshot or better yet an animation to showcase the changes you made?

You can use something like https://www.cockos.com/licecap/

Thanks!

yagebu:
@goanpeca: the only user-visble change is the use of the standard form validation utilities, the look of these might of course differ from browser to browser, here is an example:

https://googlechrome.github.io/samples/report-validity/

I discovered that some of the `<button>`s had the wrong types, which lead to them also submitting the form. That is fixed now.

yagebu:
I've removed the `npm update` commits and moved them to a separate PR.

I didn't manage to take a screenshot of the shown message since taking the screenshot would take over focus from the browser and remove it... Also, I didn't find an easy way to capture it as an animation (working under Wayland on Linux here). If you want to see this general browser feature in action, I think the link I posted above probably gives a good enough impression.

xlotlu:
@yagebu sorry to bug you, I started to review this and it's quite a lot of information. Would it be too difficult to separate the refactorings away from the #179 bugfix? Or does it depend on that?

yagebu:
> Would it be too difficult to separate the refactorings away from the #179 bugfix? Or does it depend on that?

I didn't really manage to untangle it, sorry. I think without the refactorings I'd basically have to redo the (larger) bugfixes from scratch.

runfalk:
I know I started looking at this a while back and maybe I misunderstood it but it didnt seem to work then. I just tried it and it does seem to work. Maybe I forgot to recompile it, I don't know.

I think the code changes are fine. I'll leave it up to @nixjdm to decide whether to include it or not. I personally think we should include it. Worst case we have to release a bugfix version.

EDIT: And sorry for being slow on this one.

PR #766, opened 2020-05-02
Bug in Pagination.iter_pages()

dairiki:
### Issue(s) Resolved

As is, [`Pagination.iter_pages()`][1] does not work correctly.

Assuming that I understand correctly how `iter_pages` is supposed to work, there are two issues with the current implementation:

1. There is an off-by-one error in the interpretation of the `right_current` argument.  As things stand now, if one wants to include one page on either side of the current page in the page list, then one must set `left_current=1`, but `right_current=2`.  (Setting `right_current=1` will result in no pages being listed to the right of the current page.  Setting `right_current=0` can result in the current page not even being listed.)

2. If one sets `right_edge=0` to prevent the listing of any pages at the end of the page range,  the current implementation does not include a trailing `None` (to represent the elided pages) in the resulting page list.  (It should do so when the "current" page range does not include the final page.)

[1]: <https://github.com/lektor/lektor/blob/04509f83319472e883e5b04ab7c530263743e343/lektor/pagination.py#L78>

### Description of Changes

This pull request first introduces some unit tests for `iter_pages`.  (It seems to be totally untested in the current test suite.)  Many of these tests fail.

The second commit in this pull request fixes the bugs in `iter_pages`.

### Discussion

This does change the behavior of `iter_pages`, and so may affect any sites which currently use it.  Worst case, though, it will result in one extra page button in the pager.

### DIFF
--- a/lektor/pagination.py
+++ b/lektor/pagination.py
@@ -86,10 +86,10 @@ class Pagination(object):
                range is specified by the ``left_edge`` argument (which
                may be zero).
 
-            2. A range around the current page.  This range will include
-               ``left_edge`` pages before the current page, and
-               ``right_edge`` pages after the current page.  This
-               range always includes the current page.
+            2. A range around the current page.  This range will
+               include ``left_current`` pages before the current page,
+               and ``right_current`` pages after the current page.
+               This range always includes the current page.
 
             3. Finally, a range (always ending at the last page) at
                the end of the page sequence.  The length of this range
--- a/lektor/pagination.py
+++ b/lektor/pagination.py
@@ -77,9 +77,28 @@ class Pagination(object):
 
     def iter_pages(self, left_edge=2, left_current=2,
                    right_current=5, right_edge=2):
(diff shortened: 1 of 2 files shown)


dairiki:
Note: the failing CI checks appear to be due to #767

xlotlu:
@dairiki please rebase and push force?

runfalk:
I don't think it's really clear how the arguments are supposed to work (and this seems to be undocumented so it's probably OK to change). Would you mind adding some information to the docstring what each parameter does? I think it'd be beneficial if someone else want to depend on this method.

dairiki:
@xlotlu Rebased. It looks like tests pass now.   Whew! :-)

@runfalk I will work on the docstring.  Standby.

dairiki:
(Bump.)
Can this be merged, or is something holding it up?

xlotlu:
@dairiki thank you, and sorry for the delay.

PR #769, opened 2020-05-02
handle filetype API change

### LINKED ISSUE #767: filetype==1.0.7 breaks get_image_info()
In version 1.0.7, `filetype` renames `filetype.image` to `filetype.image_match` (h2non/filetype.py@e7babc0).  That breaks [this line][2] in `lektor.imagetools.get_image_info`.

Since this rename happened without so much as a minor version bump from `filetype`, I guess `filetype.image` is not considered part of the public API of the `filetype` package.  It looks like the intended public API is through `filetype.match`.

I haven't tested this, but I think replacing `filetype.image(...)` with `filetype.match(..., matchers=filetype.image_matchers)` might do the trick.

[2]: <https://github.com/lektor/lektor/blob/04509f83319472e883e5b04ab7c530263743e343/lektor/imagetools.py#L335>

xlotlu:
fixes #767

### DIFF
--- a/lektor/imagetools.py
+++ b/lektor/imagetools.py
@@ -332,7 +332,7 @@ def get_image_info(fp):
     if any(map(head.strip().startswith, magic_bytes)):
         return get_svg_info(fp)
 
-    _type = filetype.image(bytearray(head))
+    _type = filetype.image_match(bytearray(head))
     fmt = _type.mime.split("/")[1] if _type else None
 
     width = None
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ setup(
         'click>=6.0',
         'enum34;python_version<"3.4"',
         'EXIFRead',
-        'filetype',
+        'filetype>=1.0.7',
         'Flask',
         'functools32;python_version<"3.2.3"',
         'inifile',


lektor/imagetools.py
```diff
@@ -332,7 +332,7 @@ def get_image_info(fp):
     if any(map(head.strip().startswith, magic_bytes)):
         return get_svg_info(fp)
 
-    _type = filetype.image(bytearray(head))
+    _type = filetype.image_match(bytearray(head))
```
dairiki, line 335: Another option here would be:
```
_type = filetype.match(bytearray(head), filetype.image_matchers)
```
This would have the advantage of being compatible with all versions of `filetype` and so wouldn't require the version pin.  (Note that `filetype.image`/`filetype.image_match` is a one-line function which does exactly this.)

xlotlu: I know, but that would be giving up an undocumented feature for an even lower-level undocumented feature.

My first choice was doing a `try / import as / except ImportError`. But there's really no point in that, the very few people that do a pip install --upgrade will be served by the version pin, and those that somehow manage to install an old version will get bitten by this regardless, and there's nothing we can do about it now.

PR #770, opened 2020-05-02
Load loopcontrols and debug jinja extensions

nixjdm:
This loads all available jinja extensions, adding the 3 of the 6 built-in jinja extensions that were previously excluded: i18n, loopcontrols, and debug. https://jinja.palletsprojects.com/en/2.11.x/extensions/

Rather than hardcoding them, this method will pick up however many are available. This will keep up with any additions, infrequent thought they may be, and pick up any that plugins potentially add. Writing it this way provides a little future-proofing.

### DIFF
--- a/lektor/environment.py
+++ b/lektor/environment.py
@@ -428,9 +428,9 @@ class Environment(object):
 
         self.jinja_env = CustomJinjaEnvironment(
             autoescape=self.select_jinja_autoescape,
-            extensions=['jinja2.ext.autoescape',
-                        'jinja2.ext.with_',
-                        'jinja2.ext.do'],
+            extensions=['jinja2.ext.debug',
+                        'jinja2.ext.do',
+                        'jinja2.ext.loopcontrols'],
             loader=jinja2.FileSystemLoader(
                 template_paths)
         )
--- a/tests/test_environment.py
+++ b/tests/test_environment.py
@@ -1,6 +1,6 @@
 def test_jinja2_extensions(env):
     extensions = env.jinja_env.extensions
 
-    assert 'jinja2.ext.AutoEscapeExtension' in extensions.keys()
-    assert 'jinja2.ext.WithExtension' in extensions.keys()
+    assert 'jinja2.ext.DebugExtension' in extensions.keys()
     assert 'jinja2.ext.ExprStmtExtension' in extensions.keys()
+    assert 'jinja2.ext.LoopControlExtension' in extensions.keys()


xlotlu:
> Rather than hardcoding them, this method will pick up however many are available.

Don't you think it's a bit too magic? Non-explicit?

> This will keep up with any additions, infrequent thought they may be, and pick up any that plugins potentially add.

Umh.

nixjdm:
You've got me to reconsider, and I agree. I'll update the PR to simply add ~~the three that are missing.~~ loopcontrols and debug. i18n isn't useful without more work.

---

I was thinking this was a straightforward enough and easy stepping stone, with #772, to get the ability to load custom jinja extensions in plugins. But there's other ways to go about it that I didn't see yesterday. For instance, `Environment` could have its signature extended to be able to pass a list of additional jinja extensions to it, then we simply append those to the list on this line. That's more explicit.

xlotlu:
It would be nice if the `Environment` were easily configurable...

davidism:
`with_` and `autoescape` don't do anything they're built-in. They'll start issuing deprecation warnings in 2.0.

nixjdm:
Good to know, thanks @davidism. I'll remove them in this PR.

davidism:
`loopcontrols` usually indicates you're doing too much in the template, although maybe that's not an issue for Lektor? Regardless, it's not very discoverable as probably very few people are aware it exists, or that it's enabled in Lektor.

`debug` *might* be useful during template development, but could be problematic if it gets left in during builds, although I don't think it would leak anything in this context.

#778 seems like a better idea, let users control their environment. Also, if extensions are enabled by default, how can they be *disabled* in #778?

nixjdm:
@davidism I finally remember another plugin that deals with i18n, and looked and it does add an extension after creation of the Jinja env https://github.com/numericube/lektor-i18n-plugin/blob/master/lektor_i18n.py#L240

So I looked and see the `add_extension` method, which looks like it adds extensions in the same way the Jinja Environment's `__init__` does. https://github.com/pallets/jinja/blob/6478c22f29bb85b2bae635602148d95df5e5c7ce/src/jinja2/environment.py#L348-L364 Should we just use this? This PR and #778 might be unnecessary if we can just use that.

All that goes against the line in the docs. From https://jinja.palletsprojects.com/en/2.11.x/extensions/#adding-extensions: 

> Once the environment is created additional extensions cannot be added.

goanpeca:
Documentation aside (it could be outdated) is there a reason for not being able to update an extension in the Env after it has been created?, it would seem like the most straightforward way of doing it in general?

We may need to ask if that is valid or not over there?

davidism:
Hmm, looks like `add_extension` does what you want. I'll make an issue about updating those docs.

nixjdm:
Thanks @davidism. Considering we can just use that, I'll close this and just make a small new PR to trim the unneeded `with_` and `autoescape`, and follow up with those related issues.

PR #775, opened 2020-05-03
use full version of RobotoSlab font; install with npm

### LINKED ISSUE #743: Replace RobotoSlab font with full UTF-8 version

yagebu:
### Issue(s) Resolved

Fixes #743

### Related Issues / Links

Closes #744

### Description of Changes

This change installs the RobotoSlab font that is used for the admin UI with npm instead of vendoring it in the repository and uses the full UTF-8 version. For compatibility with IE, I've added both the `woff` and the `woff2` format.


lektor/admin/static/less/bootstrap-overrides.less
```diff
@@ -35,13 +35,21 @@
   font-family: 'Roboto Slab';
   font-style: normal;
   font-weight: 300;
-  src: local('Roboto Slab Light'), local('RobotoSlab-Light'), url('../fonts/RobotoSlab-Light-webfont.ttf') format('truetype');
```
xlotlu, line 38: @yagebu you switched from Light to Regular

xlotlu: Oops, nevermind me, I wasn't paying attention.

xlotlu:
As always, thanks.

PR #776, opened 2020-05-03
proper relative paths

### LINKED ISSUE #272: Broken relative URLs on pages without a trailing slash
Relative URLs don't take account of whether the current page URL ends with a slash or not. So a relative URL on a page with URL `/about.html` will be given the same relative URL as one on a page with URL `/about/` but the `/about.html` should have one less `../`.

This error doesn't show up if the site is hosted at the root of a domain since the browser ignores the `../` which takes it above the root, but it causes broken links on sites hosted below the root.

I have published a modified version of the "quickstart" that show this error at https://codesimple-tests.gitlab.io/lektor-dotted-url-issue/. The `about/` page works as expected but the `about2.html` and `about3.html` pages have invalid links.

This issue is easily recreated by generating the "quickstart" project and then making of copy of `content/about` as `content/about2.html`. If the site is then generated and hosted on a sub-path, the issue can be seen by visiting the `about2.html` page.

I have created a modified version of the `utils.make_relative_url` function at https://gitlab.com/codesimple-tests/lektor-dotted-url-issue/blob/master/new_make_relative_url.py which fixes the issue for me but I'm not that familiar with the Lektor code base so don't know if this change would have any other unwanted effects.

### LINKED ISSUE #713: Allow relative urls abbreviated when they are on the same level

xlotlu:
Fixes #713, alternative to #714
Fixes #272 (possibly, it may have been fixed before).


xlotlu:
Note that the new function isn't applied to the code yet. Just waiting for review before altering the rest of the code.

runfalk:
I think this looks good. Maybe update the changelog as well before merging?

xlotlu:
@runfalk thanks, but please compare the solution over in #714 too. We had a long debate there about (perceived) readability vs performance, and I'm obviously biased towards this one.

runfalk:
I compared them without diving too deep. I think your might be slightly easier to follow. At first glance I liked @relikd's more since it looked simpler and is shorter, but the more I try to understand it the less sense it makes to me (maybe I'm too tired).

The thing that makes both solutions tricky for me are the nested if statements that may or may not return. I think I like your a bit better though. If you could come up with a way to un-nest the if statements that would be the clear winner.

I'd probably also move the `_mktarget(path)` outside the function and call it `_path_as_str(path, is_dir)` as mktarget doesn't convey what it actually does to me.

nixjdm:
> I think your might be slightly easier to follow. At first glance I liked @relikd's more since it looked simpler and is shorter, but the more I try to understand it the less sense it makes to me

I found this easier to follow. It could be made even easier perhaps by adding some additional helper functions, if that lets us cut out things try/elses and continues, but it's not bad as-is. I'm not sure adding a few helper functions is worth the performance cost if this is terribly sensitive. It's hard to judge the performance hit from that without trying it, and I haven't. If this isn't a big slowdown, great :slightly_smiling_face:  From the the last few comments on #714, it seemed like this is *not* a huge performance slow down, or am I misreading?

From that thread:

> The build time for my 140 recipe page takes more than a minute which I think is infeasible.

1 min, for 1 page? Is that real? That doesn't seem to match the other numbers in that comment, or the other comments. @relikd

> I'd probably also move the _mktarget(path) outside the function and call it _path_as_str(path, is_dir) as mktarget doesn't convey what it actually does to me.

Seconded. I'd put the additional helpers (if any) in that larger scope as well. I think that'd be slightly easier to read and test potentially, and slightly faster than nested (not redefining every run).

xlotlu:
> 1 min, for 1 page?

I think he meant 140 pages. Either way it's not the url resolving that's the culprit.

>  From the the last few comments on #714, it seemed like this is _not_ a huge performance slow down, or am I misreading?

Well, it _is_ 4 to 8 times slower than the raw `os.path` approach, resulting in (only) 100k iterations in 4 seconds. :) If we'll ever end up worrying about these numbers then we're just awesome.

nixjdm:
Right, just had to double check :) Yes, this is definitely not one of the slower parts of Lektor then. I think the fix plus readability improvement is worth it. So I'm generally in support of this approach. We might still be able make it a tad easier to read, but I like what I see.

relikd:
the 140 pages are totally unrelated to relative URLs. It was just a general statement that Lektor is quite slow. And though 100k iterations per 4 seconds does not sound slow I still think these numbers can be quite common for normal sized projects. AFAIK relative URLs are not only processed for links within documents, but also used by Lektor internally to follow paths. But I can be wrong, its quite some time since I looked at the code.

I do not have strong feelings for one solution or the other. @xlotlu But I remember you had an open issue back then https://github.com/lektor/lektor/pull/714#issuecomment-560439614. Did you address it or will the handling increase the nesting level further?

xlotlu:
> I do not have strong feelings for one solution or the other. @xlotlu But I remember you had an open issue back then [#714 (comment)](https://github.com/lektor/lektor/pull/714#issuecomment-560439614). Did you address it or will the handling increase the nesting level further?

Yes, I fixed it back then, there's tests taking care of it now.

> It was just a general statement that Lektor is quite slow.

Indeed it is. We'll be taking care of it step by step.

> And though 100k iterations per 4 seconds does not sound slow I still think these numbers can be quite common for normal sized projects.

May be, but the point is other parts are so very slow, that any optimisation in this particular place won't even make a dent to the total build time.

> AFAIK relative URLs are not only processed for links within documents, but also used by Lektor internally to follow paths. But I can be wrong, its quite some time since I looked at the code.

I don't think so, judging by the windows bug I just fixed this never deals with paths. I didn't look at all the code, but I suspect it only gets called by the `url` filter.

xlotlu:
Guys, thank you for your assistance. It's been a pleasure.

@runfalk / @nixjdm I'll update the changelog and consider this done.

lektor/utils.py
```diff
@@ -529,28 +533,74 @@ def bool_from_string(val, default=None):
...
+        return target
+
+    if source.is_absolute() and target.is_absolute():
+        # convert them to relative paths to simplify the logic
+        source = source.relative_to("/")
+        target = target.relative_to("/")
+
+    # is the source an ancestor of the target?
+    try:
+        relpath = target.relative_to(source)
```
runfalk: I think I find `return target.relative_to(source)` slightly better.

lektor/utils.py
```diff
@@ -529,28 +533,74 @@ def bool_from_string(val, default=None):
...
-    target = posixpath.normpath(posixpath.join(base, target))
-    if ends_in_slash and target[-1:] != '/':
-        target += '/'
-
-    return (prefix + '../' * depth).rstrip('/') + target
+        return relpath
+
+    # even if it isn't, one of the source's ancestors might be
+    distance = PurePosixPath(".")
+    for ancestor in source.parents:
```
runfalk, line 591: Feel free to ignore this if you think it makes things more complicated, but you could do something like:

```python
# Traverse upwards to the closest common ancestor. This will always return since the top is guaranteed to be "."
for distance, ancestor in enumerate(chain([source], source.parents)):
    try:
        return PurePosixPath(*[".."] * distance) / target.relative_to(ancestor)
    except ValueError:
        pass
```

That way you'd avoid having basically the same logic twice, at the expense of some readability.

lektor/utils.py
xlotlu: nice

lektor/utils.py
xlotlu: The original was enumerate-based. To me this the version now is highly readable, I think this is pushing it just for the sake of returning from the `try`.

Hmm. How about returning from the `else`? (see new commit)

runfalk: Sure, I think this is OK.

lektor/utils.py
```diff
@@ -529,28 +533,70 @@ def bool_from_string(val, default=None):
...
+        source = source.relative_to("/")
+        target = target.relative_to("/")
+
+    # is the source an ancestor of the target?
+    try:
+        return target.relative_to(source)
+    except ValueError:
+        pass
+
+    # even if it isn't, one of the source's ancestors might be
```
runfalk, line 588: At this point one of the source's ancestors *must be* an ancestor of target, right?

xlotlu: If there's nothing in common, the root is a common ancestor. (or "the empty path" in this case, as they're relative)

xlotlu: force-pushed a comment to make this clear

runfalk: Fix looks good.

PR #781, opened 2020-05-11
PR: Update setup.py

goanpeca:
### Issue(s) Resolved

Fixes #

### Description of Changes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,6 +44,7 @@ setup(
         'requests[security]',
         'setuptools',
         'watchdog',
+        'Werkzeug<1',
     ],
     tests_require=tests_require,
     extras_require={


setup.py
```diff
@@ -44,6 +44,7 @@
         'requests[security]',
         'setuptools',
         'watchdog',
+        'Werkzeug<1',
```
dairiki, line 47: What is the reason for the pin?

dairiki: @goanpeca While there is nothing wrong with pinning dependencies, especially on an application (as Lektor is) rather than a library, they are still to be avoided, when possible.  (In any case,  they are generally not a viable long-term solution.)

Perhaps it is my error, but I can find no comments either in the source, the commit message, or the pull request as to why this pin is needed.

I would suggest a comment alongside the pin in `setup.py` explaining why it is necessary, and the conditions under which it might be removed.  (Or, alternatively, a comment referencing an issue/pull which describes the reasoning behind the pin.)

goanpeca: Hi @dairiki sorry for the late response. It is an intermediate solution to getting lektor-atom to work properly while we migrate it to use a different Feed generator.

>I would suggest a comment alongside the pin in setup.py explaining why it is necessary, and the conditions under which it might be removed. (Or, alternatively, a comment referencing an issue/pull which describes the reasoning behind the pin.)

Yes, indeed, I should have done that. Since it was master and was temporary I did not pay to much attention to it, but I will add a comment. We use lektor from master to build the lektor website and test latest lektor.

I expect the real fix to come in a couple of weeks so this change should not affect the official relelases.

Cheers!

dairiki: @goanpeca Thank you for the explanation!

Still, pins in the master branch to fix plugins, or specific deployments, do not smell quite right to me...

Putting the version pin into lektor-atom's setup.py would seem ideal, but I guess that doesn't work (due to the workings of Lektor's custom package management?

Since the real motivation here seems to be to fix the lektor website build, perhaps the pin could be placed in the website build script?

runfalk: Yeah @dairiki, maybe it's because Lektor is installed first and then when the plugins are downloaded pip will not downgrade the werkzeug dependency. The correct workaround for now is probably to modify `.travis.yml` for the website repo to manually uninstall and then re-install the pinned werkzeug version.

This also means that our fix release for lektor-atom doesn't really work unfortunately.

goanpeca: Yes!

I will try o fix the feed generator within this week, revert and change on travis.

dairiki: @goanpeca Just as a side note: it is pretty straight-forward to generate an atom feed just with a Jinja template (in a manner similar to how the Lektor docs describe [generating a sitemap](https://www.getlektor.com/docs/guides/sitemap/).)  (I would guess generating RSS rather than Atom is equally straight-forward — they're both just XML — though I haven't actually done that.)

I.e. another option would be to drop the dependency of the website on lektor-atom, and just generate the feed directly.

PR #782, opened 2020-05-17
Added function to check errors in publisher outputs #677

### LINKED ISSUE #677: Travis CI deploy failed actually but report success
## Background

I created a lector repo: ~~https://github.com/Jeff-Tian/jeff-tian~~

## What happened
And I enabled travis deployment. But before I set up the LEKTOR_DEPLOY_USERNAME and LEKTOR_DEPLOY_PASSWORD, the travis reported the deployment was successful, which surprised me.

## Actual 
But if unfold the message you can see a lot of errors:
image

## Expected
It should raise error

arendondiosa:
### Issue(s) Resolved

Fixes #677

### Description of Changes

_The list of outputs in deploy are strings without error handling and the deployment is always successful (Done!)_

Added a function to check if a line (string) is interpreted as an error using a list of error related words. If it is interpreted as an error, an exception is raised and the error is displayed (Finishing the execution of the deployment)

Screen Shot 2020-05-17 at 9 54 47 AM

### DIFF
--- a/lektor/publisher.py
+++ b/lektor/publisher.py
@@ -654,8 +654,8 @@ class GithubPagesPublisher(Publisher):
     def _check_error(line):
         """
         Check that line does not contain a git error keyword.
-        Check if a line (string) is interpreted as an error using a list of error related words. If it is interpreted
-        as an error, an exception is raised.
+        Check if a line (string) is interpreted as an error using a list of error related words.
+        If it is interpreted as an error, an exception is raised.
         :param line: String. Line to evaluate. Ie, 'error: src refspec gh-pages does not match any'
         """
         error_substrings = ['fatal:', 'error:']
--- a/lektor/publisher.py
+++ b/lektor/publisher.py
@@ -650,6 +650,19 @@ class GithubPagesPublisher(Publisher):
             branch = 'gh-pages'
         return branch
 
+    @staticmethod
+    def _check_error(line):
+        """
+        Check that line does not contain a git error keyword.
+        Check if a line (string) is interpreted as an error using a list of error related words. If it is interpreted
+        as an error, an exception is raised.
+        :param line: String. Line to evaluate. Ie, 'error: src refspec gh-pages does not match any'
+        """
+        error_substrings = ['fatal:', 'error:']
+
+        if any(substring in line.lower() for substring in error_substrings):
+            raise PublishError(line)
+
     def publish(self, target_url, credentials=None, **extra):
         if not locate_executable('git'):
             self.fail('git executable not found; cannot deploy.')
@@ -664,10 +677,12 @@ class GithubPagesPublisher(Publisher):
                 return Command(['git'] + args, cwd=path, **kwargs)
 
             for line in git(['init']).output:
+                self._check_error(line)
                 yield line
             ssh_command = self.update_git_config(path, target_url, branch,
                                                  credentials)
             for line in git(['remote', 'update']).output:
+                self._check_error(line)
                 yield line
 
             if git(['checkout', '-q', branch], silent=True).wait() != 0:
@@ -676,10 +691,13 @@ class GithubPagesPublisher(Publisher):
             self.link_artifacts(path)
             self.write_cname(path, target_url)
             for line in git(['add', '-f', '--all', '.']).output:
+                self._check_error(line)
                 yield line
             for line in git(['commit', '-qm', 'Synchronized build']).output:
+                self._check_error(line)
                 yield line
             for line in git(['push', 'origin', branch]).output:
+                self._check_error(line)
                 yield line
 
 
--- a/lektor/publisher.py
+++ b/lektor/publisher.py
@@ -651,17 +651,17 @@ class GithubPagesPublisher(Publisher):
         return branch
 
     @staticmethod
-    def _check_error(line):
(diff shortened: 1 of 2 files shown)


runfalk:
I think this is a bit fragile. What if the user's locale is not English?

Is there a reason why returncode (https://github.com/lektor/lektor/blob/968d570ef44a47e2b3dba365e13e94c3416be5e4/lektor/publisher.py#L131) can't be used?

goanpeca:
Right! We could use that!

lektor/publisher.py
```diff
@@ -676,10 +691,13 @@ def git(args, **kwargs):
             self.link_artifacts(path)
             self.write_cname(path, target_url)
             for line in git(['add', '-f', '--all', '.']).output:
+                self._check_error(line)
                 yield line
```
goanpeca, line 700: Maybe something like this:
```python
cmd = git(['add', '-f', '--all', '.'])
lines = []
for line in cmd.output:
    lines.append(line)
    yield line

self._check_error(cmd.returncode, lines)
```

Since we do not know the line that actually displays an error, we should probably store them and pass them to the helper function to check the return code and print all the lines.

Something like that @runfalk ?

runfalk: I think it'd be better to change the implementation of Command to throw exceptions by default if the return code is != 0. That way we don't need the `self._check_error()` calls. We can add a kwarg to the constructor at a later point if we want to suppress errors somewhere else. Currently it's only used for the git command inside this file.

arendondiosa: I think there is a problem using the return code:
The output is 1 in errors and other lines that are not errors.

@runfalk @goanpeca Any ideas?

Screen Shot 2020-05-26 at 11 52 28 AM

runfalk: I assume we want to abort on any failed command? And 0 normally means success when it comes to return codes.  (Any Nonzero return code is error)

goanpeca: I think that what @arendondiosa is saying is that even non failing commands are rteturning a 1 value for the return code.

runfalk: I assumed the 1 after `.....` was the debug printed return code.

Which OS are you testing on @arendondiosa? When I try git on Linux I get a non-zero return code:

```
$ git remote add test http://x
$ git push test master
fatal: kan inte komma åt "http://x/": Could not resolve host: x
$ echo $?
128
```

arendondiosa: Testing git commands (MacOS):
Screen Shot 2020-05-26 at 4 57 09 PM

@runfalk What I mean: I think some commands should return code 0 (e.g. `On branch gh-pages` or` Initial commit`) 🤔 

If we use the validation `return code == 1`, the execution ends with commands that are not errors

runfalk: I think my quick replies may have introduced some confusion. All commands that succeed will by convention have 0 as its return code. All commands that fail will by convention have a non-zero code, some programs use different codes for different errors while some don't.

The normal way to test would be:

```
is_successful = returncode == 0
```

arendondiosa: @runfalk This is clear to me

> The normal way to test would be:
> `is_successful = returncode == 0`

What I don't understand is why outputs like `On branch gh-pages`, `nothing to commit` or `Initial commit` have as returncode `1` if these are not errors (I think) 🤔

runfalk: Oh, I see! So *nothing to commit* kind of is an error (at least not the expected case). We can either special case this command or add the `--allow-empty` to allow commits with no changes.

In the image you posted earlier everything fails at the fetch step so nothing after that should succeed I would assume? I assume you intentionally misspelled the repo name to trigger an error? And since the fetch fails I wouldn't expect any command after that to succeed.

arendondiosa: @runfalk  You're right, in the example above I caused an error to test the outputs.

In this example, the command is successful but there are outputs with return code = 1 (I think these are not errors). So, I think we couldn't use `returncode == 1` to validate errors.

What do you think?

Screen Shot 2020-05-31 at 3 23 07 PM

lektor/publisher.py
```diff
@@ -651,18 +651,18 @@ def detect_target_branch(self, target_url):
...
         Check if a line (string) is interpreted as an error using a list of error related words.
-        If it is interpreted as an error, an exception is raised.
+        If interpreted as an error, an exception is raised
+        :param cmd: Lektor Command
         :param line: String. Line to evaluate. Ie, 'error: src refspec gh-pages does not match any'
         """
-        error_substrings = ['fatal:', 'error:']
-
-        if any(substring in line.lower() for substring in error_substrings):
+        if cmd.wait() == 1:
```
goanpeca, line 662: This should probably be != 0 @arendondiosa ?

goanpeca: Also since you are now waiting, the logic of the yield keeps working alright?

Since the command fails it might not be able to yield lines one by one. In this case we might need to change the for for something that just waits, does the check and then prints all the lines at once

yagebu:
This looks pretty abandoned to me, so I'm closing this PR. If someone wants to continue this, I think we shouldn't check for some strings in the output but rather check whether `git` has a non-zero returncode

Issue #304, opened 2016-10-26
Disable Admin Link


smth:
It'd be useful to be able to prevent the addition of the admin link (pencil), via an option in the project file perhaps. If this is already possible, I can't find it mentioned in the docs.

singingwolfboy:
I find this surprising. The admin is one of Lektor'a best features -- why do you want to disable that link? Can you explain your use-case?

smth:
Agreed that admin is a nice feature, and I'm not talking about disabling "admin" just the link to it.

My use case is that I'm a visual designer, I often design in code, often running Lektor. The box in the top right of the page can be distracting and sometimes obscures an element that I am designing.

If I want to get to admin section, there are other simple ways to get there, such as typing `admin` in the address bar. So on balance, I'd rather the pencil icon wasn't there.

I could of course use CSS to move/hide it, but I'd prefer not to include Lektor specific code.

seblu:
I have a similar issue, the link hides a menu item. I use an ad-blocker to hide it.

torian257x:
@seblu just saying: you can add a little bit of css to remove the link or move it somewhere else, right? it's just another html element on that page

seblu:
I did not think about it. The following code do the job, thanks.

```
#lektor-edit-link {
 visibility: hidden;
}
```

PR #787, opened 2020-06-13
Import docs

### LINKED ISSUE #756: Moving lektor-website into this repository
The Lektor website currently lives in its own repository. I'd like to streamline the commit process by moving it into this repository as a folder (`website/`). I think this has been discussed in other issues but I can't seem to find any mention of it.

Why I think this is better:

* PRs can fix both documentation and code in one go which means documentation will always be in sync with the code
* Dogfooding Lektor master and ensure that Lektor master can build a decently sized site

What I propose:

* Export the lektor-website commit history verbatim into `website/` (technically not verbatim since commit hashes will be rewritten, but close enough)
* Remove `assets/static/` and let CI build it instead. We may want to prune this before importing to avoid growing the Lektor repository size
* Migrate to GitHub actions from Travis and auto-deploy
* Move issues from lektor-website to lektor
* Remove/archive lektor-website

xlotlu:
This merges the docs from `lektor-website` as a stand-alone site, along with the needed models, templates etc.

Closes https://github.com/lektor/lektor/issues/756
Blocks https://github.com/lektor/lektor-website/issues/226

The script generating this branch is available at <https://gist.github.com/xlotlu/1dd9076e4886faa77797f1b3720f20a3>. It uses <https://github.com/newren/git-filter-repo/>.


xlotlu:
Guys, please help with a checklist for getting things done? Please also commit to this branch liberally and help with related issues?

First thing that comes to mind is https://github.com/lektor/lektor-website/issues/219.

xlotlu:
Unfortunately I never thought about the original #PR / #issue references in the imported commits, which polluted lots of local issues with wrong references. See for example https://github.com/lektor/lektor/issues/304 (yey for another spurious reference).

Force-pushing only made things worse, as the commits are preserved in the reflog and github doesn't remove the links.

dwt:
You could probably rework the script that imports the old repo to regex replace all #$bug_id references with absolute links to the original repo.

But to be fair: If it was just for me, I would be ok with just importing the current status without transporting all the history. It would be nice to get the history, but going forward it will probably not be super neccessary.

xlotlu:
> You could probably rework the script that imports the old repo to regex replace all #$bug_id references with absolute links to the original repo.

@dwt this is already there, it's just that I decided to use the `<project-name>/#<thing-id>` form instead of the full url:
https://gist.github.com/xlotlu/1dd9076e4886faa77797f1b3720f20a3#file-import-docs-sh-L36

> But to be fair: If it was just for me, I would be ok with just importing the current status without transporting all the history. It would be nice to get the history, but going forward it will probably not be super neccessary.

Well, it's true that from a practical standpoint the commit history is not as relevant as code history, more so due to them not being in sync, but... It's history.

Even if it would be entirely irrelevant (which it isn't, diffs are useful information, for docs too), I personally think the commit history should be preserved because it snapshots people's contributions. It's a good homage to their efforts to preserve this info.

dwt:
@xlotlu More power to you - as long as it doesn't prevent getting this merged.

PR #789, opened 2020-06-18
Uniform handling of deferred (descriptor) model data

dairiki:
### Issue(s) Resolved

**TLDR**; Lektor provides a mechanism whereby the computation of model field values can be deferred until such time as the values are accessed (e.g. via template expansion).  (The built-in `markdown` type makes use of this feature, as do a number of  plugin-provided extension types.)  Currently, there are a couple of places where reference to such a deferred field value will not work — worse yet, in these cases, failure is silent and, thus, confusing.  This pull request aims to fix things so that deferred field values will work as expected in those cases.

#### Background

Lektor provides a mechanism whereby the computation of certain values of page model data can be deferred until template-evaluation time.  (For example, the compilation of markdown text within a field of type `markdown` does not happen until template-evaluation time.)  (This late-evaluation happens in [Record.__getitem__](https://github.com/lektor/lektor/blob/16413178e3fe1cec15806611f154ebc4abe80c58/lektor/db.py#L457), where, if field value appears to be a *descriptor*, its `.__get__` method is called to compute the final value.)

Normally, these deferred values are transparently computed when they are accessed, either programmatically (e.g. `page['body']`) or within a Jinja template expression (e.g. `{{ this.body }}`), so for the most part, these deferred types “just work”.

There are a couple of code paths in the current code base which access the raw model data directly, rather than accessing it through `Record.__getitem__`.  In these cases, deferred values do not work.  Worse yet, currently, no warning or error messages are generated when this happens, so the failures are hard for the user to diagnose.

Fortunately, both of these cases are fairly straightforward to fix.  This pull request does that.

#### Problematic code-paths

There are, however, a couple of code-paths where access to model data does not currently go through `Record.__getitem__`.  Attempts to access deferred values in those code-paths will fail (usually in a subtle way, without generating any sort of explicit error or warning message.)

Two code-paths where this happens are:
* The computation of the default slug for a child page, based on the parents [slug_format](https://www.getlektor.com/docs/models/children/#child-slug-behavior).
* In [Record.get_sort_key](https://github.com/lektor/lektor/blob/16413178e3fe1cec15806611f154ebc4abe80c58/lektor/db.py#L446).

There are other places where the raw model data is accessed directly without going through `Record.__getitem__`, but in those cases, specific system fields (e.g. `_model` or  `_hidden`) are being accessed.  I do not believe these system fields ever contain deferred values, so directly accessing them does/should not cause issues.

##### Default slug computation

The computation of the default value for a page’s `_slug` happens in [Database.get_default_slug](https://github.com/lektor/lektor/blob/16413178e3fe1cec15806611f154ebc4abe80c58/lektor/db.py#L1472).

For example, suppose we would like to create URLs for blog posts that contain the year of publication.   Setting `slug_format` to

```html+jinja
{{ "{0.year:04d}/{1}".format(this.pub_date, this._id) }}
```

in the datamodel for the blog parent page will do this.

This currently fails, however, in cases where `pub_date` is a deferred value. This might happen when using the [lektor-git-timestamp](https://github.com/dairiki/lektor-git-timestamp) plugin which can be used to auto-compute `pub_date` from the git commit timestamps. Currently, when the `slug_format` value is jinja-evaluated, `this` is set not to a `Page` instance, but rather to a dict containing the raw page data. `This.pub_date` in this context is the deferred descriptor, not the actual desired value. Currently this will result in [silent failure](https://github.com/lektor/lektor/blob/16413178e3fe1cec15806611f154ebc4abe80c58/lektor/datamodel.py#L336) of the expansion of `slug_format`.

This can be fixed by deferring the computation of the default `_slug` value until after the `Page` instance is fully constructed.

##### Record.get_sort_key

In [Record.get_sort_key](https://github.com/lektor/lektor/blob/16413178e3fe1cec15806611f154ebc4abe80c58/lektor/db.py#L446), the sort key value is fetched directly from the raw model data,\
rather than through `Record.__getitem__`.

As a result, if attempts are made to sort blog posts by, e.g., `pub_date`, and pub_date is a deferred value as described above, this will cause problems.

This can be fixed, simply by using `Record.__getitem__` to access the sort field’s value.

### Description of Changes

* [X] Wrote at least one-line docstrings (for any new functions)
* [X] Added unit test(s) covering the changes (if testable)

* Modified `Database.process_data` to defer computation of the default `_slug`.  This solves the issue with `slug_format`s which access deferred model data.
* Modified `Record.sort_key` to access model field values using `Record.__getitem__` rather than accessing the raw model data (`this._data`) directly.  This fixes the issue with using deferred fields as sort keys.
* Added a diagnostic message to indicate when expansion of `slug_format` fails (instead of silently using the fall-back slug when there is an error in the `slug_format`.)

### DIFF
--- a/lektor/db.py
+++ b/lektor/db.py
@@ -1295,9 +1295,9 @@ def get_default_slug(record, type_=None):
     is computed by expanding the parent’s ``slug_format`` value.
 
     """
-    parent = getattr(record, 'parent', None)
+    parent = getattr(record, "parent", None)
     if parent is None:
-        return ''
+        return ""
     return parent.datamodel.get_default_child_slug(record.pad, record)
 
 
--- a/lektor/db.py
+++ b/lektor/db.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # pylint: disable=too-many-lines
 import errno
 import functools
--- a/lektor/db.py
+++ b/lektor/db.py
@@ -1283,6 +1283,22 @@ def _iter_datamodel_choices(datamodel_name, path, is_attachment=False):
     yield "none"
 
 
+def get_default_slug(record, type_=None):
+    """Compute the default slug for a page.
+
+    This computes the default value of ``_slug`` for a page.  The slug
+    is computed by expanding the parent’s ``slug_format`` value.
+
+    """
+    parent = getattr(record, 'parent', None)
+    if parent is None:
+        return ''
+    return parent.datamodel.get_default_child_slug(record.pad, record)
+
+
+default_slug_descriptor = property(get_default_slug)
+
+
 class Database:
     def __init__(self, env, config=None):
         self.env = env
@@ -1502,21 +1518,10 @@ class Database:
                         ctx.record_dependency(dep_model.filename)
         return record
 
-    def get_default_slug(self, data, pad):
-        parent_path = posixpath.dirname(data["_path"])
-        parent = None
-        if parent_path != data["_path"]:
-            parent = pad.get(parent_path)
-        if parent:
-            slug = parent.datamodel.get_default_child_slug(pad, data)
-        else:
-            slug = ""
-        return slug
-
     def process_data(self, data, datamodel, pad):
         # Automatically fill in slugs
         if not data["_slug"]:
-            data["_slug"] = self.get_default_slug(data, pad)
+            data["_slug"] = default_slug_descriptor
         else:
             data["_slug"] = data["_slug"].strip("/")
 
--- a/lektor/db.py
+++ b/lektor/db.py
@@ -452,7 +452,11 @@ class Record(SourceObject):
             else:
                 field = field.lstrip("+")
                 reverse = False
-            rv[idx] = _CmpHelper(self._data.get(field), reverse)
+            try:
+                value = self[field]
+            except KeyError:
+                value = None
+            rv[idx] = _CmpHelper(value, reverse)
         return rv
 
     def __contains__(self, name):
--- a/lektor/db.py
+++ b/lektor/db.py
@@ -1288,7 +1288,7 @@ def _iter_datamodel_choices(datamodel_name, path, is_attachment=False):
     yield "none"
 
 
-def get_default_slug(record, type_=None):
+def get_default_slug(record):
     """Compute the default slug for a page.
 
     This computes the default value of ``_slug`` for a page.  The slug
--- a/lektor/datamodel.py
+++ b/lektor/datamodel.py
@@ -341,8 +341,7 @@ class DataModel:
                 self._child_slug_tmpl[1].evaluate(pad, this=data).strip().split()
             ).strip("/")
         except Exception as exc:
-            reporter.report_generic(
-                "Failed to expand child slug_format: %s" % exc)
+            reporter.report_generic("Failed to expand child slug_format: %s" % exc)
             return "temp-" + slugify(data["_id"])
 
     def get_default_template_name(self):
--- a/lektor/datamodel.py
+++ b/lektor/datamodel.py
@@ -10,6 +10,7 @@ from lektor.environment.expressions import FormatExpression
 from lektor.i18n import generate_i18n_kvs
 from lektor.i18n import get_i18n_block
 from lektor.pagination import Pagination
+from lektor.reporter import reporter
 from lektor.types import builtin_types
 from lektor.types.base import RawValue
 from lektor.utils import bool_from_string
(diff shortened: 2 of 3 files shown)


dairiki:
@xlotlu @runfalk Is there any chance of getting this merged?

Yesterday, I set up a new site-building environment on a new computer.  I stumbled across the issues fixed by this PR and spent a couple of hours "fixing" them before discovering that I had already fixed them once already in this PR. 😂😐

(I've just now rebased this back onto the current master.)

lektor/db.py
```diff
@@ -1283,6 +1288,22 @@ def _iter_datamodel_choices(datamodel_name, path, is_attachment=False):
     yield "none"
 
 
+def get_default_slug(record, type_=None):
```
yagebu: The `type_` argument seems to be unused.

lektor/db.py
```diff
@@ -1283,6 +1288,22 @@ def _iter_datamodel_choices(datamodel_name, path, is_attachment=False):
...
+    is computed by expanding the parent’s ``slug_format`` value.
+
+    """
+    parent = getattr(record, "parent", None)
+    if parent is None:
+        return ""
+    return parent.datamodel.get_default_child_slug(record.pad, record)
+
+
+default_slug_descriptor = property(get_default_slug)
```
yagebu, line 1304: Is there a specific reason why you moved this out of the class? In particular having it defined as a property outside of the class seems odd.

yagebu (changes requested):
Hi, thanks for this PR and the detailed description of the issues you fixed. This looks pretty good, I just left two small comments.

lektor/db.py
dairiki: Background: What we're trying to do here is defer the computation of the default slug until later.  To do that we put a descriptor — an object which as a callable `.__get__` attribute — the the records `data["slug"]`.  When the slug is first accessed via [Record.__getitem__](https://github.com/lektor/lektor/blob/46b28df9f97e4b03382f7d33b8323f7343c58672/lektor/db.py#L461) the `.__get__` will be called to compute the actual value of the data field.
This "deferred data field" API just happens to look very much like Python’s [attribute descriptor API](https://docs.python.org/3/howto/descriptor.html).

I had originally coded this section something like:

```python
class DefaultSlugDescriptor(object):
    def __get__(self, record, type_=None):
        parent = getattr(record, "parent", None)
￼        if parent is None:
￼            return ""
￼        return parent.datamodel.get_default_child_slug(record.pad, record)

default_slug_descriptor = DefaultSlugDescriptor()
```

I refactored this because I thought using `property` to generate the descriptor was more succinct and a bit clearer, but the two approaches are equivalent.

[As an aside, in the above example the unused `type_` argument is included to exactly match Python’s attribute descriptor API.  I don't think lektor ever invokes `__get__` with the second argument, so it could be omitted, but I copied the style of, e.g., [MarkdownDescriptor](https://github.com/lektor/lektor/blob/46b28df9f97e4b03382f7d33b8323f7343c58672/lektor/types/formats.py#L9) here.
In reference to your previous comment, I erroneously copied the phantom `type_` argument when I refactored to using `property` to construct the descriptor.]

Now to get to your question...

We are not using `property` to create a property of class `Database`, so it's use would seem odd whether left in the class or not.  I.e. we could do something like:

```python
class Database(object):
[...]
    def get_default_slug(self, record):
        # Note: method signature changed.  Used to be get_default_slug(self, data, pad)
        [...]

    def process_data(self, data, datamodel, pad):
￼        if not data["_slug"]:
￼            data["_slug"] = property(self.get_default_slug)
￼        else:
￼            [...]
```
(Note also that the `get_default_slug`, though not explicitly marked so, has always been a static method — it doesn't depend on or mutate the state of the `Database` instance.)

Anyhow, I thought it was clearer and cleaner outside of the class.  If you like another alternative better, I'd be happy to change it.

lektor/db.py
dairiki: You’re right.  This is, I think, a remnant of a previous factoring of my code.  I will fix it.

(See my reply to your next comment for a bit more on that history.)

lektor/db.py
yagebu: Thanks for the explanation :)

> Anyhow, I thought it was clearer and cleaner outside of the class.

I agree in principle. However, I think in this instance, since moving it out of the class moves it a few hundred lines up in this rather monstrous file, it might actually be easier to keep it in the class (as a `@staticmethod` starting with an underscore). But if you'd rather have it outside the class that's fine too (maybe add an underscore), I can't really make up my mind whether I strongly prefer either.

dairiki:
Okay.  I've rebased again (due to merge conflicts introduced by #871) and deleted the spurious and unused `type_` argument.

Anything else?

yagebu (approved):
> Anything else?

Nope - thanks again :)

dairiki:
Thank you!

PR #817, opened 2020-10-08
admin: fix checkboxes widget

### LINKED ISSUE #812: Checkboxes are not working
Checkboxes cannot be checked with version 3.2.0. Checkboxes work on prior versions. I used Firefox browser version 80.0.1. They also do not work on Brave browser.

yagebu:
### Issue(s) Resolved

Fixes #812

### Description of Changes

This change from a function to an arrow function allows us to access the
containing class with `this` in the line below.

### DIFF
--- a/lektor/admin/static/js/widgets/multiWidgets.jsx
+++ b/lektor/admin/static/js/widgets/multiWidgets.jsx
@@ -43,7 +43,7 @@ export class CheckboxesInputWidget extends React.PureComponent {
     let { className, value, placeholder, type, onChange, ...otherProps } = this.props
     className = (className || '') + ' checkbox'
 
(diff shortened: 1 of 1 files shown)


xlotlu:
This should make it straight to a release, no?

runfalk:
Yeah, let's make this 3.2.1.

yagebu:
I've created a `3.2-maintenance` branch and backported this patch to it.

runfalk:
Wouldn't it make sense to just base the maintenance branch on current master? I think it's fine to include the JS version bumps and everything else. Would be nice if the CI on the maintenance branch was green as well.

yagebu:
Yeah, why not :+1: , will do.

PR #819, opened 2020-10-08
drop Python 2

### LINKED ISSUE #818: Drop Python 2 support.
I believe now after the release of Lektor 3.2 it is time to drop support for Python 2, which has reached EOL on the first day of this year.

yagebu:
### Issue(s) Resolved

Fixes #818


hugovk:
Tip: Add `python_requires='>=3.5',` to `setup.py`, it'll help pip only install this version for people with Python 3.5+, and fallback to the previous version for those still with 2.7.

runfalk:
Seems like something is really weird with both the Windows and macOS runners recently.

runfalk:
I re-ran the tests and it seems like macOS fixed itself.

The Windows tests are broken due to an installation error with ImageMagick.

I like @hugovk's suggestion of using `python_requires`. I did some digging and it's supported by `setuptool` >= 24.2.0 which was released in July 2016. That means we may break Ubuntu 16.04 LTS which is still maintained, but I think it's a reasonable risk to take. The workaround would be to manually specify the desired version.

hugovk:
The note at https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires:

* setuptools 24.2.0+ is needed by the _maintainer_ when creating sdist + wheels to include the metadata.

* pip 9+ is needed by _end users_ to read the metadata, or it's ignored (and they're in the same situation anyway).

So shouldn't break things Ubuntu 16.04 (any more than it would be). And you can recommend people upgrade pip as well.

If you want to handle pip < 9 (I'm not sure how common that is), you can could do something like "Fail early at import time" and "Fail early in setup.py" from https://python3statement.org/practicalities/

goanpeca:
Go for it! :-) Thanks for working on this @yagebu

nixjdm:
This needs a rebase now after #823.

yagebu:
Rebased. Due to all strings being changed to double quotes that was quite fun :D But it's already nice to not have to worry about formatting after such refactorings :)

yagebu:
This seems to have gathered no objections in more than a month, so I'll merge this soon.

PR #820, opened 2020-10-08
lint: fix code smells pointed out by pylint

yagebu:
On four occasions, this meant using `raise ... from ...`, which improves
tracebacks - otherwise, they'd look like there was some problem in the
exception handling code.

Also use `super()` without arguments in one place.

This is blocked by #818 since these two fixes produce invalid Python 2 code.


yagebu:
I'll close this one and include the commit in #819

PR #821, opened 2020-10-09
Add lektor package to chocolatey

### LINKED ISSUE #814: Have lektor installed via chocolatey
Chocolatey is a great package manager for windows. A simple command such as `choco install xxx` will install the necessary software and update system paths automatically.

It would be great to have lektor installed via chocolatey! I do understand the installation isn't exactly an issue right now, but it would be nice to have it this way.

- https://chocolatey.org/
- https://chocolatey.org/docs/create-packages

I'd be interested in implementing this.

nimishbongale:
### Issue Resolved

Fixes #814 by @nimishbongale

### Related Issues / Links

- https://chocolatey.org/packages/lektor/1.0.0

#### Things achieved
- Lektor can now be installed via `chocolatey` 
`choco install lektor`
It has passed the necessary automated tests as well.
- Pulls latest version (currently installs `3.2.0`)

#### Things pending
- Some fields within the `lektor.nuspec` file need to be filled in accordance to the maintainers directions.
- Scripts to auto deploy to chocolatey.
- Update website to recommend download via chocolatey.

#### Points to note
- Unless there is a breaking change to `installer.py` the deployment to chocolatey would be as simple as a change in version number in `lektor.nuspec` and a push to chocolatey.
- I have tested the files, and they seem to work just fine. The `lektor` package on chocolatey should now be publicly available to everyone! 🚀


chocolatey/lektor.nuspec
```diff
@@ -0,0 +1,34 @@
...
+    <!--<mailingListUrl></mailingListUrl>-->
+    <bugTrackerUrl>https://github.com/lektor/lektor/issues</bugTrackerUrl>
+    <tags>python pipenv cms</tags>
+    <summary>lektor</summary>
+    <description>The lektor static file content management system</description>
+    <!-- <releaseNotes>__REPLACE_OR_REMOVE__MarkDown_Okay</releaseNotes> -->
+
+    <dependencies>
+      <dependency id="python"/>
+      <dependency id="imagemagick"/>
```
runfalk: We should also add ffmpeg to here

chocolatey/ReadMe.md
```diff
@@ -0,0 +1,132 @@
+﻿## Summary
```
runfalk: I assume this file is auto-generated by chocolatey?

If so I think it should be condensed to only include the necessary steps for building the Lektor package. If there is any more information of interest this could also go here.

chocolatey/tools/installer.py
```diff
@@ -0,0 +1,388 @@
+#!/usr/bin/env python
```
runfalk, line 1: This seems to an alsost exact copy of the lektor installer. Even the Linux and macOS installer bits remain. What changes have been made? If there are little or no changes I think we should depend on the same file.

chocolatey/lektor.nuspec
```diff
@@ -0,0 +1,34 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>lektor</id>
+    <version>3.2.0</version>
+    <packageSourceUrl>https://github.com/lektor/lektor</packageSourceUrl>
+    <owners>Nimish Bongale</owners>
```
runfalk: I'm happy with you taking credit in some way (like the description), but I think the `owners` field should be something like _The Lektor Team_.

chocolatey/tools/installer.py
```diff
@@ -0,0 +1,388 @@
...
+def install_lektor(lib_dir):
+    create_virtualenv(lib_dir)
+
+    pip = get_pip(lib_dir)
+
+    args = [pip, "install"]
+    if IS_WIN:
+        # avoid fail due to PEP 517 on windows
+        args.append("--prefer-binary")
+    args.extend(["--upgrade", "Lektor"])
```
runfalk, line 227: We should lock the Lektor version to the package version.

chocolatey/lektor.nuspec
```diff
@@ -0,0 +1,34 @@
...
+
+    <title>lektor</title>
+    <authors>lektor</authors>
+    <projectUrl>https://www.getlektor.com/</projectUrl>
+    <!--<iconUrl>http://cdn.rawgit.com/__REPLACE_YOUR_REPO__/master/icons/lektor.png</iconUrl>-->
+    <!-- <copyright>Year Software Vendor</copyright> -->
+    <licenseUrl>https://github.com/lektor/lektor/blob/master/LICENSE</licenseUrl>
```
runfalk: https://www.getlektor.com/license/ <-- this is probably the more user friendly URL.

Or does this actually include the license itself in the installer? If so you should probably use the raw URL: https://raw.githubusercontent.com/lektor/lektor/master/LICENSE

chocolatey/lektor.nuspec
```diff
@@ -0,0 +1,34 @@
...
+  <metadata>
+    <id>lektor</id>
+    <version>3.2.0</version>
+    <packageSourceUrl>https://github.com/lektor/lektor</packageSourceUrl>
+    <owners>Nimish Bongale</owners>
+
+    <title>lektor</title>
+    <authors>lektor</authors>
+    <projectUrl>https://www.getlektor.com/</projectUrl>
+    <!--<iconUrl>http://cdn.rawgit.com/__REPLACE_YOUR_REPO__/master/icons/lektor.png</iconUrl>-->
```
runfalk: This should use the Egg on feet logo: https://www.getlektor.com/logo.png (just the egg part I guess)

chocolatey/tools/chocolateyuninstall.ps1
```diff
@@ -0,0 +1,6 @@
+﻿$app = Get-WmiObject -Class Win32_Product | Where-Object {
```
runfalk, line 1: Will this properly remove the virtualenv and everything?

runfalk (changes requested):
I think this is a good start.

Still missing version pinning and CI.

chocolatey/lektor.nuspec
nimishbongale: It seemed to work even without it, but I'll go ahead and add it no issues.

chocolatey/ReadMe.md
nimishbongale: It is indeed auto-generated by chocolatey. I didn't know what to add in there so yes, I'll do the needful now.

chocolatey/tools/installer.py
nimishbongale: So here's the change:

I've removed the part where the installer asks whether the lektor needs to be installed or not (Y/n)

This is because `chocolatey` provides for this step by default, and we don't have to ask the user again for it.

chocolatey/lektor.nuspec
nimishbongale: Yep! This is again autogenerated by chocolatey. Since I created the package on the site, it by default adds me as the owner. I'm happy to change it as per your direction.

chocolatey/tools/installer.py
nimishbongale: Understood. Again, by default chocolatey thinks its version `1.0.0`. I'll have to bumped to `3.2.0`.

chocolatey/lektor.nuspec
nimishbongale: Ahhh alright. The LICENSE in the installer is for a different purpose I think. Will check.

chocolatey/lektor.nuspec
nimishbongale: Cool! I'll add the logo.

chocolatey/tools/chocolateyuninstall.ps1
nimishbongale: It uninstalls lektor and deletes the associated files. I don't think it'll remove virtualenv as its a dependency. But I'll look into it!

nimishbongale:
@runfalk

Thanks for reviewing the PR! 

This is just a `v1` PR, and it was bound to have a lot of changes (because I myself wasn't sure about a lot of things)

I'll push more changes and rerequest a review. 👍

nimishbongale:
@runfalk 

I'm sorry for my lack in consistency wrt to this PR. I've made some changes, do let me know if they're okay!

dwt:
Out of interest - is there a reason this hasn't gotten merged? Having Lektor available on more package managers seems good?

runfalk:
It wasn't complete and the author dropped it before addressing the comments. The main thing being that the Lektor version was not tied to the version this installs. This would be a problem since it could lead to ambiguos bug reports. It was also missing the proper CI machinery to automatically build new releases. Given our track records of mustering releases to PyPI I'm a bit weary of adding more friction to that process (I don't want us to commit to something and then drop it).

dwt:
It seems the releases are already flowing on chocolates, so it may need to be advisable to either adopt updates there, or get them to drop it. Not sure how the package manager works, bit on homebrew the packages are community updated anyway - is this different here?

dairiki:
Public-relations-wise, there's probably more harm than good that comes from putting out something like this without the commitment to support it and keep it in sync with the main Lektor distribution.

I agree with @runfalk that unless there is someone on the release team who has the motivation and expertise to maintain a Chocolatey (or any other distribution type) it's probably better that it doesn't exist.  Or, at least, is as detached from the main project as possible.

I am not a Windows-guru and I have never once use Chocolatey, so take these observations with a grain of salt, but, other specific issues I think this PR might have:
- I doesn't look like it uninstalls very cleanly.  As pointed out in another comment, it looks like it leaves the virtualenv behind. The [Chocolatey docs][cd] say: “If your package doesn't uninstall cleanly, people will get grumpy because they'll have to manually clean up after you. Be a good human being and write an uninstaller.” This would imply that constructing the package so that it uninstalls cleanly is not automatic.
- It looks like this, as it is, can only install under `%LocalAppData%` (i.e. single-user install, rather than system-wide install.)  Chocolatey (I think) is supposed to support system-wide installs, too?
- It would be nice not to fork the stock install script, if only a few lines are different — that's just one more thing to get out of sync.  Better to introduce another environment variable or some such if another switch is needed to suppress a prompt.  A better approach might be to forego `installer.py` altogether: just package up a wheel file and install that instead?

[cd]: <https://docs.chocolatey.org/en-us/create/create-packages#uninstalling>

nimishbongale:
@runfalk @dwt @dairiki 

I would love to reopen this PR and make the necessary changes, but I'm just not sure whether it makes sense to be a part of this repository? If it does, I'd be glad.

For starters:

1. The package does not uninstall cleanly, but I have found a workaround for this lately. For this purpose, I created an `uninstaller.py` file which performs just the uninstallation, and removes the virtualenv too. Seems to work great! However, there would be code repetition in the repository itself, as the chocolatey package packs in the `installer.py` file.
2. The current lektor version on chocolatey is locked in, which means I can't actually push changes until there's a version bump on lektor itself (say to 3.2.1).
3. The CI to build it would just be additional work I believe, and I'm happy with maintaining the chocolatey package as a separate community-driven thing. If there's an inherent need for it to be included as a part of this repository, then I think there's a lot to figure out.

Thanks!

dwt:
@nimishbongale You should have a look into point releases, for example homebrew allows to push updates that are not related to upstream package updates to allow fixes in the package itself. That should be possible with chocolatey too.

nimishbongale:
> That should be possible with chocolatey too.

I did imagine it to be possible, but I wasn't able to get it done. I then posted up a question on chocolatey's gitter chat about the same, where I was told that the package was locked in and couldn't be modified. I think I'll be able to fix up the package just fine after lektor's next release!

dwt:
Maybe have a look at https://docs.chocolatey.org/en-us/create/create-packages#package-fix-version-notation - it may be that you misunderstood something.

nimishbongale:
Aha. Is it fine if I make a "date fixed" version as the article mentions? Hopefully it'll still remain to be a true representative of the current lektor version?

Thanks for the info!

runfalk:
In general Linux distributions does something like `3.2.0-1` and then `3.2.0-2`. That way you can make releases when something breaks with the packaging.

PR #823, opened 2020-10-17
Beautify code again

skorokithakis:
This is a continuation of #696, which implements #693. After review says this is okay, I will beautify the code using the configs.

Also, I want to add this to CI, but I noticed the Travis file was removed. What do we use nowadays?


yagebu:
> What do we use nowadays?

We use Github Actions now - configuration is in `.github/workflows/ci.yml`

.pre-commit-config.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - id: black
+- repo: https://github.com/asottile/reorder_python_imports
+  rev: v2.3.5
+  hooks:
+  -   id: reorder-python-imports
+- repo: https://gitlab.com/pycqa/flake8
+  rev: '3.8.4'
+  hooks:
+  - id: flake8
+    args: ["--config=setup.cfg"]
```
yagebu: Is this required? flake8 supports setup.cfg as one of its default config files.

setup.cfg
```diff
@@ -1,2 +1,6 @@
 [bdist_wheel]
 universal=1
+
+[flake8]
+exclude=wsgi.py,env/*,*/migrations/*,venv/*,.env/*,.venv/*,local_settings.py,doc/*,*/node_modules/*
```
yagebu: Is this necessary? Shouldn't `pre-commit` only run on files that are checked into git?

.pre-commit-config.yaml
skorokithakis: Hmm true, I'll remove it, thanks.

setup.cfg
skorokithakis: It shouldn't be, I just add it to all my repos. Will remove.

skorokithakis:
Is this ready to merge? I will run the actual linting after it's approved and right before we merge.

yagebu (approved):

skorokithakis:
Hmm, does anyone know why lint failed? I didn't add any exec statements.

runfalk:
Maybe the disable doesn't work when it's split across multiple lines?

https://github.com/lektor/lektor/pull/823/files#diff-75a248b337f2de305ae21d3819159e1e53ab6ca222c427a177cc759e128723a9R36-R38

skorokithakis:
Ah, that'd do it, thanks @runfalk! Will fix.

lektor/admin/modules/api.py
```diff
@@ -2,16 +2,22 @@
 import posixpath
 
 import click
-from flask import Blueprint, jsonify, request, g, current_app
-
-from lektor._compat import iteritems, itervalues
-from lektor.utils import is_valid_id
+from flask import Blueprint
```
runfalk, line 5: Which tool is causing these import splits? I don't like this idea.

skorokithakis: That's `reorder_python_imports`. What don't you like?

skorokithakis: It's done so diffs will be much simpler, you can see an added/removed line rather than hunt for a word in a long list.

runfalk: I don't like the noise it produces:

```python
# reorder_python_imports
from flask import abort
from flask import Blueprint
from flask import g
from flask import redirect
from flask import render_template
from flask import request
from flask import url_for

# black (don't know whether or not it sorts the functions alphabetically)
from flask import (
    Blueprint,
    render_template,
    abort,
    redirect, request,
    g,
    url_for,
)
```

I think the vanilla black style is much better (probably because it's similar to what I'm used to and that it's less verbose). I guess it also has the benefit of producing a smaller diff if we ever rename a file.

skorokithakis: Hmm okay, I'll revert the two commits, remove `reorder_python_imports` and redo them, though then we won't get the automatic separation of third party packages, consistent ordering, etc.

nixjdm: In other (large, unrelated) repos I've worked with, reorder-python-imports also has the effect of automatically fixing some flake8 errors, such as deduping imports, and ensuring the correct import order, which can introduce bugs, even if that's rare. Admittedly, that's less of a problem here since Lektor is small in comparison and in better shape.

I also didn't like splitting imports like this (or black, for that matter), but they've both grown on me a lot. My vote would be to keep reorder-python-imports. This single diff is a little noisy (thus the .git-blame-ignore-revs), but I think it's worth it. I now don't blink at the import lines that look like this. My 2 cents.

nixjdm: I think I typed my comment as you were typing your last one @skorokithakis . For the record, I'm ok with or without reorder-python-imports, so we can remove it if you and @runfalk want. No need to trip up this PR for that. Though I'd prefer we keep it.

skorokithakis: I would prefer to keep it too, given that I like it (which is why I added it) and how much work it would be to remove it (I'd have to redo half the PR).

My vote is to keep. This PR is ready if we do, though I don't understand why tests are failing on OS X, I think it's unrelated.

runfalk: macOS tests are flaky during to the environment setup (before anything lektor related is done). You can retrigger the tests if you want to be sure.

Though I'm still not sold on the import split I'm not gonna hold up the PR if no one else dislikes it. Maybe it's an aquired taste :laughing:

skorokithakis: I can certainly say I've acquired it and now prefer the determinism, @runfalk.

skorokithakis:
Oof, these tests just don't want to pass.

skorokithakis:
@runfalk / @nixjdm, tests won't pass, what should I do?

runfalk:
You can retrigger them by going to the run's page (https://github.com/lektor/lektor/runs/1298211911). I've retriggered them just now. If they still don't work just merge anyway I guess. It would be very weird if this introduced a platform specific bug, and if it did I guess we just have to fix it later.

skorokithakis:
@runfalk I've rerun them about five times :/ I think we should merge anyway, yeah...

skorokithakis:
This error doesn't look like it's caused by this PR, or that it's going to go away soon:

```
Error: Fetching /usr/local/Homebrew/Library/Taps/local/homebrew-openssl failed!
Fetching /usr/local/Homebrew/Library/Taps/local/homebrew-python2 failed!
fatal: invalid upstream 'origin/master'
fatal: invalid upstream 'origin/master'
```

I wonder what happened with Homebrew...

skorokithakis:
Merged, thank you all!

nixjdm:
Yeah, I was just about to say it looks like this breakage is totally unrelated to the code changes. It broke when I reran the previous tip.

PR #824, opened 2020-10-17
add tox.ini to make test execution more reproducible

yagebu:
Without tox, one either has to juggle with virtual environments or run tests/lint/coverage in the global environment. In short, with tox, life is better :)


yagebu:
This also updates the readme for our use of pre-commit and removes an outdated custom pre-commit hook

PR #825, opened 2020-10-17
add prettier and eslint

yagebu:
This is a bit like #823 but for the Javascript parts.

Somehow standard tries to do too much at once and turns out to being
inferior to combination of prettier for formatting and pure eslint for
linting.

prettier has the advantage of (like black for Python) actually
completely reformatting the file. In recent updates of stylelint, there
was always a lot of work involved to get simple stylistic things fixed.
With prettier, that's all automatic.

By using eslint directly, we can more easily enable useful plugins like
the react plugin. Also, using the "standard" linting tool instead of
standard should provide better automatic editor support.

(standard has about 250k weekly downloads on NPM, prettier and eslint
are about a factor 40 and 50 more popular in that metric).

This commits only adds config for these two tools but doesn't wire them
up to be run automatically yet.


runfalk:
I'm in favor of this. The more automation the better. I've skimmed through the changes and they look OK to me.

Before merging I'd like to make sure the validation runs as part of the CI loop.

yagebu:
> Before merging I'd like to make sure the validation runs as part of the CI loop.

I already set that up. But I'll probably change it a bit to use `pre-commit` and then merge the two lint CI runs into one.

I actually just realized that `.editorconfig` set indentation and overwrote prettier defaults for `js` (but not `jsx`) so I'll have to fix that.

yagebu:
> Before merging I'd like to make sure the validation runs as part of the CI loop.

@runfalk: see https://github.com/lektor/lektor/pull/825/checks?check_run_id=1397860764 for a run with a commit where prettier would reformat a file.

runfalk:
Hm, if it runs the pre-commit hooks. Doesn't that just reformat the code? I would have expected it to check and error if something looks off. The main thing that looks a bit odd to me is `--fix` for eslint.

yagebu:
> Doesn't that just reformat the code? I would have expected it to check and error if something looks off.

That's exactly what `pre-commit` takes care of for us (just like with `black`), it runs the hook and errors if it changed any files (or if the hook itself errored).

> The main thing that looks a bit odd to me is --fix for eslint.

eslint can automatically fix some problems

runfalk:
Yeah, I get what pre-commit and prettier does. However, the commit is already done so does the CI actually update the commit to fix the errors? It's perfectly possible for a contributor to not install the pre-commit hooks locally and make a PR that doesn't respect the code style. My question is if the CI fails if that's the case? I don't think it matters if it just fixes the thing because it can't modify the commit.

yagebu:
I think I don't quite get the point that you're trying to make here. The linters in this PR behave analogously to the linters we already have for Python, both locally and in CI. Both black and prettier reformat the files, which is how pre-commit can determine that it should fail (both locally and in CI)

> However, the commit is already done so does the CI actually update the commit to fix the errors? 

No

> It's perfectly possible for a contributor to not install the pre-commit hooks locally and make a PR that doesn't respect the code style. My question is if the CI fails if that's the case?

Yes

yagebu:
I'll merge this as I have a couple of pending fixes to the JS parts that are tedious to merge/rebase otherwise.

PR #830, opened 2020-11-10
Bugfix: allow for local rsync deployment

arabidopsis:
### Issue(s) Resolved

rsync deployment fails with `TypeError: sequence item 0: expected str instance, NoneType found`
for the following deployment:

```ini
[servers.production]
target = rsync:///home/me/static
```

### Description of Changes

Added defensive test for the existence of `ascii_host` parameter

Test added.

### DIFF
--- a/tests/test_deploy.py
+++ b/tests/test_deploy.py
@@ -324,3 +324,22 @@ def test_rsync_command_delete_no(tmpdir, mocker, env):
             "example.com:/",
         ],
     )
+
+
+def test_rsync_local(tmpdir, mocker, env):
+    output_path = tmpdir.mkdir("output")
+    publisher = RsyncPublisher(env, str(output_path))
+    target_url = url_parse("file:///path/to/directory")
+    ssh_path = tmpdir.mkdir("ssh")
+    mock_popen = mocker.patch("lektor.publisher.portable_popen")
+    publisher.get_command(target_url, str(ssh_path), credentials=None)
+    assert mock_popen.called
+    assert mock_popen.call_args[0] == (
+        [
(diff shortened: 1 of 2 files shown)


relikd:
Hi, I like the idea. I wrote a plugin for this very case a few months ago.

```
class PathPublisher(Publisher):
        cmd = [
            'rsync', '-rclzv', '--delete',
            '--exclude=.lektor', '--exclude=.git/', '--exclude=.DS_Store']

        cmd.append(self.output_path.rstrip('/\\') + '/')
        cmd.append((target_url.host or '') + target_url.path.rstrip('/') + '/')

        for line in Command(cmd):
            yield line
```

relikd:
Also, I added the `--delete` and exlude `.git` to my version because in my opinion they should be added to the official solution. But that is not my call – the team must decide.

yagebu:
Thanks @arabidopsis for this fix :) The test also looks good.

> Also, I added the --delete and exlude .git to my version because in my opinion they should be added to the official solution. 

There's already options to exclude and delete in the rsync publisher, I don't that we should explicitly set them by default.

tests/test_deploy.py
```diff
@@ -324,3 +324,22 @@ def test_rsync_command_delete_no(tmpdir, mocker, env):
             "example.com:/",
         ],
     )
+
+
+def test_rsync_local(tmpdir, mocker, env):
```
yagebu, line 329: These tests seem to be quite repetitive. I'll probably refactor them to use `pytest.mark.parametrize` in a follow-up.

yagebu (approved):

PR #842, opened 2020-11-15
Flow collapse

### LINKED ISSUE #828: Flow widget collapse button triggers Data Unsaved Warning when moving away from edit page
When a flow collapse/expand button is clicked it marks the data as 'changed' so a triggers 'Prompt'.

Since no data has changed (only the ui layout) this is confusing. See the gif below.

I have a PR I can give you to fix this if you like.

Here is a simple project you can try yourself

[Test1.zip](https://github.com/lektor/lektor/files/5505715/Test1.zip)

Visual:

flow

arabidopsis:
### Issue(s) Resolved

Fixes #828

### Related Issues / Links

Also fixes an issue where If the user creates multiple flow blocks before saving the first `Math.max([])` produces  
a newBlockId of `-Infinity`.

### Description of Changes

The problem is that the onChange method indicates both an actual contexts.lr change and *also* a UI state change
both which need a rerender. 

The *solution* -- if it is that -- is to add a second argument to the `onChange` method and a third argument to
setFieldValue to indicate this is *only* a ui state change. Since for all other uses except flowblocks this
value will be undefined  `hasPendingChanges: !uiChange` will usually be true. But not for the collapse button.

Sorry people... I didn't write any tests because I simply don't know how to for React. This is a "works for me" PR.


lektor/admin/static/js/views/EditPage.jsx
```diff
@@ -100,11 +100,11 @@ class EditPage extends RecordComponent {
     });
   }
 
-  setFieldValue(field, value) {
+  setFieldValue(field, value, uiChange) {
```
yagebu: We could make the default a bit more explicit by having this as `uiChange = false`. This would also make the comment below unnecessary.

lektor/admin/static/js/widgets.jsx
```diff
@@ -58,13 +58,14 @@ FallbackWidget.propTypes = widgetPropTypes;
...
+
 export const FieldBox = React.memo(function FieldBox(props) {
   const { field, value, placeholder, disabled } = props;
   const onChange = props.onChange
     ? props.onChange
-    : (value) => props.setFieldValue(field, value);
+    : (value, uiChange) => props.setFieldValue(field, value, uiChange);
   const className = "col-md-" + getFieldColumns(field) + " field-box";
-  let innerClassName = "field";
+  let innerClassName = `field ${field.type.name}`;
```
yagebu: This change seems to be a bit unrelated, is it wanted/necessary?

yagebu (changes requested):
Thanks for the PR, I think this is the right solution for now :)

Also, don't worry about the tests here, there's barely any for the frontend so far and this one doesn't seem like it's easily testable.

lektor/admin/static/js/widgets.jsx
arabidopsis: Sorry I have a massively hacked branch that is styling the edit box contents on field type. This just leaked.
It is unecessary for this fix.

lektor/admin/static/js/views/EditPage.jsx
arabidopsis: Changed. Still thinking in pre ES2015 days :)

lektor/admin/static/js/widgets.jsx
yagebu: No worries, can you remove it then?

arabidopsis: Done

yagebu (approved):

PR #846, opened 2020-12-03
correctly update preview page

### LINKED ISSUE #844: preview iframe does not properly track edit state
The preview iframe does not track the actual edit state as the following gif shows....

I have a PR to fix this if you like.

preview-tracking

yagebu:
This logic has always been a bit faulty and probably broke recently due
to us avoiding unnecessary re-renders. In addition to checking whether
the correct URL is loaded, we heed to check if the URL that we want to
load changed.

Fix #844


runfalk:
Looks good to me.

PR #855, opened 2020-12-19
ci: try fetching latest imagemagick release again

yagebu:


yagebu:
Well, this seems still to be broken.

Issue #793, opened 2020-07-11
`mailto:` and `ftps:` supposedly Not a valid URL


relikd:
![screen](https://user-images.githubusercontent.com/1321720/87234112-1f0f5800-c3ce-11ea-8116-98b7b9e15ad7.png)

Maybe there are more schemes that aren't supported. But `mailto:` is such a commonly used scheme that I would add it.

relikd:
Since this fix only addressed the mentioned URL schemes specifically, I want to add that there are other schemes that are still not supported with the updated version. E.g. `webcal:`, `feed:`, and probably a hundred more

relikd:
Ok, this is becoming a problem now. I can't save a page because it says the `feed:` url isnt valid. Even though I want to delete a completely unrelated part.

PR #864, opened 2021-01-19
Add support for all (and custom) URL schemes

relikd:
adds support for `webcal:` and `feed:` URLs

### Related Issues / Links

Related #793

### Description of Changes

### DIFF
--- a/lektor/admin/static/js/utils.test.ts
+++ b/lektor/admin/static/js/utils.test.ts
@@ -12,7 +12,7 @@ describe("Utils", () => {
   it("check URL validity", () => {
     ok(isValidUrl("http://example.com"));
     ok(isValidUrl("https://example.com"));
-    ok(!isValidUrl("https:file"));
+    // ok(!isValidUrl("https:file"));  // ignoring this case in favor of more generic regex
     ok(!isValidUrl("https:/example.com"));
     ok(isValidUrl("ftp://example.com"));
     ok(isValidUrl("ftps://example.com"));
@@ -20,6 +20,12 @@ describe("Utils", () => {
     ok(isValidUrl("mailto:user@example.com"));
     ok(isValidUrl("mailto:anythingreally"));
     ok(!isValidUrl("mailto:with spaces"));
+    ok(isValidUrl("z39.50r://example.com:8001/database?45"));
+    ok(isValidUrl("svn+ssh://example.com"));
+    ok(isValidUrl("feed:example.com/rss"));
+    ok(isValidUrl("webcal:example.com/calendar"));
+    ok(isValidUrl("ms-help://section/path/file.htm"));
+    ok(!isValidUrl("anyscheme:/oneslash"));
   });
 
   it("trim strings of slashes and colons", () => {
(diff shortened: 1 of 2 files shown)


yagebu (changes requested):
Thanks for the PR! Like the checklist suggests, this should have a test.

Instead of iteratively adding more and more schemes, I think it would be preferable to just accept all schemes.

yagebu (approved):
Nice, thanks :)

relikd:
damn it. Who would have thought that this can still become an issue. Should we disable the URL validity altogether?

screen

dairiki:
I am no expert on the working of the admin UI, but might it be "good enough" if, in lieu of validating via js, the input element for the URL widget were changed to `<input type="url">` (thus relying on the browser to do any validation)?

At this point, exactly what sorts of invalid URLs are we aiming to flag here?

yagebu:
> damn it. Who would have thought that this can still become an issue. Should we disable the URL validity altogether?

The current checker is quite lax and what you've input there simply is not an URL (which would look like `scheme:...`). Why use an URL field and not just a plain string field if you don't want URLs?

> I am no expert on the working of the admin UI, but might it be "good enough" if, in lieu of validating via js, the input element for the URL widget were changed to <input type="url"> (thus relying on the browser to do any validation)?

That would be an option but would probably take some work (to show the error similarly to the other inputs etc.). A simpler solution could be to use `new URL(...)` from JS and check if that works. But since the main complaint so far has been that this is too strict, I'm not too keen to spend time to make this (currently lax) check stricter.

relikd:
> Why use an URL field and not just a plain string field if you don't want URLs?

That is what I ended up doing. I guess context is relevant here. My template uses the url field as a html anchor href value. So it technically really is a URL, but a same-site relative link. And yes, using a string is fine, but it does not feel right. Because the URL type conveys additional information about the field, namely that the user is supposed to enter a valid URL(-path) and not just random characters. Additionally, the icon to the right of the field also signals its purpose (link-icon vs. paragraph-icon).

yagebu:
With `addon_label` you can also modify the icon that's displayed to the right of the input without having to change the field type.

dairiki:
If we're voting, I would vote that [relative URLs](https://datatracker.ietf.org/doc/html/rfc3986#section-4.2) should be allowed in `url` fields.  It is a common use case.  Relative URLs are URLs, and, in our case, we have the good fortune to always have an unambiguous base URL (that of page containing the `url` field) with which to interpret them.

This is an *admin* UI for a static website.  We're not concerned here with preventing HTML injection or whatever. For the most part, we should take the admin at their word that what they've typed in is what they meant.

If relative URLs are allowed, `lektor.utils.Url` (the type presented to templates for a `'url` field) could be enhanced so that it is aware of the base URL of the current page (the page that contains the `url` field.)  This would allow it to implement, e.g., an `.abspath` and perhaps an `.absurl` property.

----

If anyone really wants more specificity as to what sorts of URLs are allowed, options could be added to the `url` field to specify further constraints. E.g.
- A boolean option do disallow relative URLs
- An option (string-list) to specify allowed schemes

Ideally (I think) these constraints would be enforced both client-side by the admin UI and python-side.

PR #871, opened 2021-02-06
Reduce import cycles and linting

yagebu:
In a lot of places, imports where done within methods and functions instead of at the top level. This makes static analysis harder and might lead to import failures at run time in the worst case. Moving imports to the top level highlights any problematic cycles. I've split up some of the existing modules like `lektor.environment` to allow the imports to happen without any cycles.

I've also enabled some more pylint rules and fixed current offending pieces of code. All the changes in this PR are trivial refactors and moving of code.


dairiki:
Forgive the following opinion of this grumpy old fart:

Please consider that when these style-only-refactors-that-touch-every-bit-of-code are made it causes friction in a number of ways.  Outstanding branches, patches, pull-requests etc. no longer apply cleanly.  It makes it much harder to follow the history of the code (e.g. git blame doesn't work so well, and the commit log is just noisy.)

When changes like this are committed every few months (see #823), these disadvantages are compounded.

Lektor is in production use by many.  Some have their own custom patches that they maintain.  Stability is a feature as well.

dairiki:
The shuffling of the lektor.types package breaks [lektor-rst](https://github.com/fschulze/lektor-rst) (as well as probably any other plugins which implement their own lektor data types) where it does an ``from lektor.types import Type``.

Similarly one of my own local private plugins is broken where it does a `from lektor.environment import Expression`.  (Same for `from lektor.environment import FormatExpression`.)

yagebu:
I wouldn't really call most of the changes here "style-only-refactors-that-touch-every-bit-of-code" (parts definitely are). Import cycles have led to bugs in the past and generally make this code hard to reason about. And without tools like `black`, the code would be much harder to maintain.

> Lektor is in production use by many. Some have their own custom patches that they maintain. Stability is a feature as well.

I hear you. I don't think it's currently really feasible to both keep Lektor stable by fixing bugs without some churn in the code base. Are these custom patches things that might benefit the wider community? If so, please do submit them as PRs so that they can be maintained in sync with the rest of Lektor. If not, maybe they can be separated out more as separate packages instead of patches so that they don't rely on the code of Lektor line-for-line?

>The shuffling of the lektor.types package breaks lektor-rst (as well as probably any other plugins which implement their own lektor data types) where it does an from lektor.types import Type.

Good catch. I'll go over this code and ensure that these objects are reexported.

PR #876, opened 2021-02-07
Ctrl+e/Meta+e keyboard shortcuts for the edit page.

yagebu:
There's already a couple of keyboard shortcuts, and having one to switch between the edit page seems useful. Control+e seems to be a rather unproblematic choice. This PR supersedes #579 to implement this.

In the process I've refactored the current code to handle keyboard shortcuts to make it easier to implement more in the future. I ended up adding a new component for an `<a>` that should be associated with a keyboard shortcut, so the refactorings to change two components from classes to function components are somewhat unrelated since no use of hooks in them was actually needed.

### DIFF
--- a/lektor/admin/static/js/components/LinkWithHotkey.tsx
+++ b/lektor/admin/static/js/components/LinkWithHotkey.tsx
@@ -1,0 +1,55 @@
+import React, { MutableRefObject, ReactNode, useEffect, useRef } from "react";
+import { NavLink } from "react-router-dom";
+import { getKey, KeyboardShortcut, keyboardShortcutHandler } from "../utils";
+
+/**
+ * React hook to add a global keyboard shortcut for the given
+ * key for the lifetime of the component.
+ */
+function useKeyboardShortcut(
+  key: KeyboardShortcut,
+  action: (ev: KeyboardEvent) => void
+): void {
+  useEffect(() => {
+    const handler = keyboardShortcutHandler(key, action);
+    window.addEventListener("keydown", handler);
+    return () => window.removeEventListener("keydown", handler);
+  }, []);
+}
+
+/**
+ * Add a global keyboard shortcut for the given key and simulate
+ * a click on the ref'ed element.
+ */
+function useKeyboardShortcutRef<T extends HTMLElement>(
+  key: KeyboardShortcut
+): MutableRefObject<T | null> {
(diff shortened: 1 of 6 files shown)


dwt:
Thanks for picking up my slack!

lektor/admin/static/js/components/LinkWithKey.tsx
```diff
@@ -0,0 +1,55 @@
...
+  key: KeyboardShortcut
+): MutableRefObject<T | null> {
+  const el = useRef<T | null>(null);
+  useKeyboardShortcut(key, () => {
+    el.current?.click();
+  });
+  return el;
+}
+
+export default function LinkWithKey(props: {
```
runfalk: Couldn't we just add an optional `keyboardShortcut` optional prop to `Link` (seems to be `LektorLink` on master)? This seems to be a verbatim copy of that one pretty much.

If you'd like to keep this as a separate function, may I suggest changing the name to `LinkWithHotkey` to make it clearer that we're talking about a keyboard shortcut?

I think it'd also be nice if the keyboard shortcut was displayed on hover or something, to aid discoverability.

runfalk (commented):
Nice one. I only had one minor comment.

lektor/admin/static/js/components/LinkWithKey.tsx
yagebu: > Couldn't we just add an optional keyboardShortcut optional prop to Link (seems to be LektorLink on master)? This seems to be a verbatim copy of that one pretty much.

There's two reasons for having this as a separate component:

1. While this currently only has one additional prop, we might want to require additional props for this component at some later point like additional information to put into the `title` attribute. Typing this would be cumbersome, so I think it's better to have a little duplication here.
2. Since we cannot conditionally call hooks, we'd either wastefully attach the global event listener for every `Link` or we'd need some more logic inside `useEffect` to only conditionally attach/detach the event listener (which would also have to consider the case that the component just switched from having a keyboard shortcut to not having one or vice-versa).

I'll rename the component to `LinkWithHotkey` :+1:

> I think it'd also be nice if the keyboard shortcut was displayed on hover or something, to aid discoverability.

That's already implemented via the `title` attribute.

runfalk: Sweet, I'm happy with that.

PR #877, opened 2021-02-11
Fixes installing lokal plugins with extras_require

### LINKED ISSUE #865: Wrong way to install dependencies of plugins
Hi there,

I've got the problem, that I can't install a local version of the Lektor-atom plugin, as it has an `extras_require`of `test`.

This leads to a `egg.info/requires.txt` of 

```
MarkupSafe
feedgenerator

[test]
lxml
pytest
```

Which of course is no valid requirements.txt file that pip can use, which then leads to an explosion when this is installed [here](https://github.com/lektor/lektor/blob/7132afbeaa64429dd1367dcb719573792a8b2f43/lektor/packages.py#L150).

I am not entirely sure what the correct way to consume this information is, but I suspect that one should go about using the `pkg_resources` package. Quick googling has provided [this stackoverflow](https://stackoverflow.com/a/50193639/4572750) which looks promising.

dwt:
The problem this tries to solve is that local plugin packages with `extras_require` fail to install. I had this problem with [lektor-atom](https://github.com/lektor/lektor-atom/) where I actually stumbled upon this issue here https://github.com/lektor/lektor-atom/issues/32

### Issue(s) Resolved

Fixes #865 

### Description of Changes

The pull request converts the requires.txt file from the egg_info call into a requirements.txt file, by stripping out everything from the first `[extras_require]` section in that file. It also documents why all of that is neccessary, as it would seem a much more logical choice to just call `pip install --target $folder --editable $plugin` and be done with it - but sadly that doesn't work because pip has a bug https://github.com/pypa/pip/issues/4390

### DIFF
--- a/lektor/packages.py
+++ b/lektor/packages.py
@@ -103,6 +103,13 @@ def download_and_install_package(
 
 def install_local_package(package_root, path):
     """This installs a local dependency of a package."""
+
+    # Becaus of these bugs:
+    # - pip https://github.com/pypa/pip/issues/4390
+    # - setuptools https://github.com/pypa/setuptools/issues/392
+    # we cannot just call `pip install --target $folder --editable $package`.
+    # Hence the workaround of first installing only the package and then it's dependencies
+
     # XXX: windows
     env = dict(os.environ)
     env["PYTHONPATH"] = package_root
@@ -143,14 +150,28 @@ def install_local_package(package_root, path):
         dirs = os.listdir(tmp)
         if rv != 0 or len(dirs) != 1:
             raise RuntimeError("Failed to create egg info for local package.")
-        requires = os.path.join(tmp, dirs[0], "requires.txt")
+        requires_path = os.path.join(tmp, dirs[0], "requires.txt")
 
-        # We have dependencies, install them!
-        if os.path.isfile(requires):
-            download_and_install_package(package_root, requirements_file=requires)
+        if os.path.isfile(requires_path):
+            # We have dependencies, install them!
+            requirements_path = \
+                requiriements_txt_from_requires_file_in_same_directory(requires_path)
+            download_and_install_package(package_root, requirements_file=requirements_path)
+
     finally:
         shutil.rmtree(tmp)
 
+def requiriements_txt_from_requires_file_in_same_directory(requires_path):
+    "requirex.txt can contain [extras_require] sections wich pip doesn't understand"
+    requirements_path = os.path.join(os.path.dirname(requires_path), "requirements.txt")
+    with open(requirements_path, "w") as requirements, open(requires_path, 'r') as requires:
+        for line in requires.readlines():
+            # extra requires section starts here -> not valid requirements.txt syntax
+            if line.strip().startswith('['):
+                break
+            requirements.write(line)
+
+    return requirements_path
 
 def get_package_info(path):
     """Returns the name of a package at a path."""
--- a/lektor/packages.py
+++ b/lektor/packages.py
@@ -166,7 +166,8 @@ def install_local_package(package_root, path):
 
 
 def requiriements_txt_from_requires_file_in_same_directory(requires_path):
-    "requirex.txt can contain [extras_require] sections wich pip doesn't understand"
+    """Create a sanitized copy of `requires.txt`."""
+    # requires.txt can contain [extras_require] sections wich pip doesn't understand
     requirements_path = os.path.join(os.path.dirname(requires_path), "requirements.txt")
     with open(requirements_path, "w") as requirements, open(
         requires_path, "r"
--- a/lektor/packages.py
+++ b/lektor/packages.py
@@ -101,6 +101,26 @@ def download_and_install_package(
         raise RuntimeError("Failed to install dependency package.")
 
 
+def download_and_install_packages(package_root, packages_with_versions):
+    """This downloads and installs a specific version of a package."""
+    # XXX: windows
+    env = dict(os.environ)
+
+    args = [
+        sys.executable,
+        "-m",
+        "pip",
+        "install",
+        "--target",
+        package_root,
+    ]
+
+    args.extend(packages_with_versions)
+    rv = portable_popen(args, env=env).wait()
+    if rv != 0:
+        raise RuntimeError("Failed to install dependency package.")
+
+
 def install_local_package(package_root, path):
     """This installs a local dependency of a package."""
 
@@ -123,10 +143,9 @@ def install_local_package(package_root, path):
     editable_install_without_dependencies(package_root, path)
 
     # using a for loop syntax here so the generator can clean up the tempdir after the loop
-    for requirements_path in requirements_from_unfinished_editable_install_at_path(
-        path
-    ):
-        download_and_install_package(package_root, requirements_file=requirements_path)
+    requirements = requirements_from_unfinished_editable_install_at_path(path)
+    if requirements is not None:
+        download_and_install_packages(package_root, requirements)
 
 
 def editable_install_without_dependencies(package_root, path):
@@ -176,27 +195,20 @@ def requirements_from_unfinished_editable_install_at_path(path):
 
         if os.path.isfile(requires_path):
             # We have dependencies, install them!
-            requirements_path = requirements_txt_from_requires_file(requires_path)
-            yield requirements_path
-
+            return requirements_from_requires_file(requires_path)
     finally:
         shutil.rmtree(tmp)
 
 
-def requirements_txt_from_requires_file(requires_path):
+def requirements_from_requires_file(requires_path):
     """Create a sanitized copy of `requires.txt`."""
     # requires.txt can contain [extras_require] sections wich pip doesn't understand
-    requirements_path = os.path.join(os.path.dirname(requires_path), "requirements.txt")
-    with open(requirements_path, "w") as requirements, open(
-        requires_path, "r"
-    ) as requires:
+    with open(requires_path, "r") as requires:
         section_name, extracted_requirements = next(
             pkg_resources.split_sections(requires)
         )
         if section_name is None:
-            requirements.write("\n".join(extracted_requirements))
-
-    return requirements_path
+            return extracted_requirements
 
 
 def get_package_info(path):
--- a/lektor/packages.py
+++ b/lektor/packages.py
@@ -190,11 +190,11 @@ def requirements_txt_from_requires_file(requires_path):
     with open(requirements_path, "w") as requirements, open(
         requires_path, "r"
     ) as requires:
-        for line in requires.readlines():
-            # extra requires section starts here -> not valid requirements.txt syntax
-            if line.strip().startswith("["):
-                break
-            requirements.write(line)
+        section_name, extracted_requirements = next(
+            pkg_resources.split_sections(requires)
+        )
+        if section_name is None:
+            requirements.write("\n".join(extracted_requirements))
 
     return requirements_path
 
--- a/lektor/packages.py
+++ b/lektor/packages.py
@@ -332,10 +332,9 @@ def update_cache(package_root, remote_packages, local_package_path, refresh=Fals
                     package_root, os.path.join(local_package_path, package[1:])
                 )
             else:
-                download_and_install_packages(
-                    package_root,
-                    ["%s%s%s" % (package, version and "==" or "", version or "")],
-                )
+                requirement_spec = f"{package}=={version}" if version else package
+                download_and_install_packages(package_root, [requirement_spec])
+
         write_manifest(manifest_file, all_packages)
 
 
--- a/lektor/packages.py
+++ b/lektor/packages.py
@@ -75,34 +75,8 @@ def remove_package_from_project(project, name):
     return None
 
 
-def download_and_install_package(
-    package_root, package=None, version=None, requirements_file=None
-):
-    """This downloads and installs a specific version of a package."""
-    # XXX: windows
-    env = dict(os.environ)
-
-    args = [
-        sys.executable,
-        "-m",
-        "pip",
-        "install",
-        "--target",
-        package_root,
-    ]
-
-    if package is not None:
-        args.append("%s%s%s" % (package, version and "==" or "", version or ""))
-    if requirements_file is not None:
-        args.extend(("-r", requirements_file))
-
-    rv = portable_popen(args, env=env).wait()
-    if rv != 0:
-        raise RuntimeError("Failed to install dependency package.")
-
-
 def download_and_install_packages(package_root, packages_with_versions):
-    """This downloads and installs a specific version of a package."""
+    """This downloads and installs a specific version of packages."""
     # XXX: windows
     env = dict(os.environ)
 
@@ -352,7 +326,10 @@ def update_cache(package_root, remote_packages, local_package_path, refresh=Fals
                     package_root, os.path.join(local_package_path, package[1:])
                 )
             else:
-                download_and_install_package(package_root, package, version)
+                download_and_install_packages(
+                    package_root,
+                    ["%s%s%s" % (package, version and "==" or "", version or "")],
+                )
         write_manifest(manifest_file, all_packages)
 
 
--- a/lektor/packages.py
+++ b/lektor/packages.py
@@ -104,7 +104,7 @@ def download_and_install_package(
 def install_local_package(package_root, path):
     """This installs a local dependency of a package."""
 
-    # Becaus of these bugs:
+    # Because of these bugs:
     # - pip https://github.com/pypa/pip/issues/4390
     # - setuptools https://github.com/pypa/setuptools/issues/392
     # we cannot just call `pip install --target $folder --editable $package`.
--- a/lektor/packages.py
+++ b/lektor/packages.py
@@ -144,7 +144,6 @@ def install_local_package(package_root, path):
 
     editable_install_without_dependencies(package_root, path)
 
-    # using a for loop syntax here so the generator can clean up the tempdir after the loop
     requirements = requirements_from_unfinished_editable_install_at_path(path)
     if requirements is not None:
         download_and_install_packages(package_root, requirements)
--- a/lektor/packages.py
+++ b/lektor/packages.py
@@ -4,6 +4,7 @@ import shutil
 import site
 import sys
 import tempfile
+from contextlib import contextmanager
 from subprocess import PIPE
 
 import click
@@ -131,6 +132,15 @@ def install_local_package(package_root, path):
     if rv != 0:
         raise RuntimeError("Failed to install local package")
 
+    with requirements_from_unfinished_editable_install_at_path(
+        path
+    ) as requirements_path:
+
+        download_and_install_package(package_root, requirements_file=requirements_path)
+
+
+@contextmanager
+def requirements_from_unfinished_editable_install_at_path(path):
     # Step 2: generate the egg info into a temp folder to find the
     # requirements.
     tmp = tempfile.mkdtemp()
@@ -154,18 +164,13 @@ def install_local_package(package_root, path):
 
         if os.path.isfile(requires_path):
             # We have dependencies, install them!
-            requirements_path = requiriements_txt_from_requires_file_in_same_directory(
-                requires_path
-            )
-            download_and_install_package(
-                package_root, requirements_file=requirements_path
-            )
-
+            requirements_path = requiriements_txt_from_requires_file(requires_path)
+            yield requirements_path
     finally:
         shutil.rmtree(tmp)
 
 
-def requiriements_txt_from_requires_file_in_same_directory(requires_path):
+def requiriements_txt_from_requires_file(requires_path):
     """Create a sanitized copy of `requires.txt`."""
     # requires.txt can contain [extras_require] sections wich pip doesn't understand
     requirements_path = os.path.join(os.path.dirname(requires_path), "requirements.txt")
--- a/lektor/packages.py
+++ b/lektor/packages.py
@@ -4,7 +4,6 @@ import shutil
 import site
 import sys
 import tempfile
-from contextlib import contextmanager
 from subprocess import PIPE
 
 import click
@@ -111,6 +110,15 @@ def install_local_package(package_root, path):
     # we cannot just call `pip install --target $folder --editable $package`.
     # Hence the workaround of first installing only the package and then it's dependencies
 
+    editable_install_without_dependencies(package_root, path)
+
+    for requirements_path in requirements_from_unfinished_editable_install_at_path(
+        path
+    ):
+        download_and_install_package(package_root, requirements_file=requirements_path)
+
+
+def editable_install_without_dependencies(package_root, path):
     # XXX: windows
     env = dict(os.environ)
     env["PYTHONPATH"] = package_root
@@ -132,14 +140,7 @@ def install_local_package(package_root, path):
     if rv != 0:
         raise RuntimeError("Failed to install local package")
 
-    with requirements_from_unfinished_editable_install_at_path(
-        path
-    ) as requirements_path:
 
-        download_and_install_package(package_root, requirements_file=requirements_path)
-
-
-@contextmanager
 def requirements_from_unfinished_editable_install_at_path(path):
     # Step 2: generate the egg info into a temp folder to find the
     # requirements.
@@ -166,6 +167,7 @@ def requirements_from_unfinished_editable_install_at_path(path):
             # We have dependencies, install them!
             requirements_path = requiriements_txt_from_requires_file(requires_path)
             yield requirements_path
+
     finally:
         shutil.rmtree(tmp)
 
--- a/lektor/packages.py
+++ b/lektor/packages.py
@@ -75,7 +75,7 @@ def remove_package_from_project(project, name):
     return None
 
 
-def download_and_install_packages(package_root, packages_with_versions):
+def download_and_install_packages(package_root, requirement_specifiers):
     """This downloads and installs a specific version of packages."""
     # XXX: windows
     env = dict(os.environ)
@@ -89,7 +89,7 @@ def download_and_install_packages(package_root, packages_with_versions):
         package_root,
     ]
 
-    args.extend(packages_with_versions)
+    args.extend(requirement_specifiers)
     rv = portable_popen(args, env=env).wait()
     if rv != 0:
         raise RuntimeError("Failed to install dependency package.")
--- a/lektor/packages.py
+++ b/lektor/packages.py
@@ -165,14 +165,14 @@ def requirements_from_unfinished_editable_install_at_path(path):
 
         if os.path.isfile(requires_path):
             # We have dependencies, install them!
-            requirements_path = requiriements_txt_from_requires_file(requires_path)
+            requirements_path = requirements_txt_from_requires_file(requires_path)
             yield requirements_path
 
     finally:
         shutil.rmtree(tmp)
 
 
-def requiriements_txt_from_requires_file(requires_path):
+def requirements_txt_from_requires_file(requires_path):
     """Create a sanitized copy of `requires.txt`."""
     # requires.txt can contain [extras_require] sections wich pip doesn't understand
     requirements_path = os.path.join(os.path.dirname(requires_path), "requirements.txt")
--- a/lektor/packages.py
+++ b/lektor/packages.py
@@ -110,8 +110,19 @@ def install_local_package(package_root, path):
     # we cannot just call `pip install --target $folder --editable $package`.
     # Hence the workaround of first installing only the package and then it's dependencies
 
+    # Because pip can resolve dependencies differently when it is called with them individually,
+    # it is important to call it with all of them together. Also
+
+    # requires.txt is specified here:
+    # https://setuptools.readthedocs.io/en/latest/deprecated/python_eggs.html#requires-txt
+    # requirementx.txt is specified here:
+    # https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format
+    # What can be part of the dependencies is specified here:
+    # https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#id4
+
     editable_install_without_dependencies(package_root, path)
 
+    # using a for loop syntax here so the generator can clean up the tempdir after the loop
     for requirements_path in requirements_from_unfinished_editable_install_at_path(
         path
     ):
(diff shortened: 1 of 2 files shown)


lektor/packages.py
```diff
@@ -103,6 +103,13 @@ def download_and_install_package(
 
 def install_local_package(package_root, path):
     """This installs a local dependency of a package."""
+
+    # Becaus of these bugs:
```
webknjaz:
```suggestion
    # Because of these bugs:
```

lektor/packages.py
```diff
@@ -143,15 +150,36 @@ def install_local_package(package_root, path):
...
 
-        # We have dependencies, install them!
-        if os.path.isfile(requires):
-            download_and_install_package(package_root, requirements_file=requires)
     finally:
         shutil.rmtree(tmp)
 
 
+def requiriements_txt_from_requires_file_in_same_directory(requires_path):
+    "requirex.txt can contain [extras_require] sections wich pip doesn't understand"
```
webknjaz:
```suggestion
    """Create a sanitized copy of `requires.txt`."""
    # requires.txt can contain [extras_require] sections wich pip doesn't understand
```

tests/test_packages.py
```diff
@@ -0,0 +1,99 @@
...
+
+import pytest
+
+from lektor import packages
+
+
+@pytest.fixture
+def temp_dir():
+    with tempfile.TemporaryDirectory() as temp_dir:
+        yield pathlib.Path(temp_dir)
```
webknjaz: Pytest already has a `tmpdir` fixture built-in. No need to reinvent it. That fixture is considered deprecated and you should use `tmp_path` instead (which returns `pathlib.Path()` objects).

dwt: Oh, I didn't know that existed - thanks for the hint.

dwt:
@webknjaz That should be all of the points you brought up - I would be happy for a second round of review.

lektor/packages.py
```diff
@@ -143,15 +150,37 @@ def install_local_package(package_root, path):
         dirs = os.listdir(tmp)
         if rv != 0 or len(dirs) != 1:
             raise RuntimeError("Failed to create egg info for local package.")
-        requires = os.path.join(tmp, dirs[0], "requires.txt")
+        requires_path = os.path.join(tmp, dirs[0], "requires.txt")
+
+        if os.path.isfile(requires_path):
+            # We have dependencies, install them!
+            requirements_path = requiriements_txt_from_requires_file_in_same_directory(
```
dairiki: Spelling aside, I wonder whether this long function name is really helping readability.

(The "in the same directory", I think, is an implementation detail and can be omitted from the function name with no loss of value.  The caller of this function really has no reason to care where the requirements.txt file is created.)

The essence of what this function does is: extract the `install_requires` from the egg-info `requires.txt`, and write them to `requirements.txt` suitable for passing to pip.

Note that we are lucky — and have been taking advantage of the fact — that the format of setuptool's `requires.txt` is similar to that of pip’s `requirements.txt`.  Really, their formats are defined by two different specifications for two different (though related) purposes.  `Requires.txt` is part of the egg-info metadata — its format is specified by setuptools (which is who defined the egg format). `Requirements.txt` is really mostly a way of passing arguments to pip — its format is specified by pip.

The reason the `requires.txt` was used as a `requirements.txt` in the first place was that is was easy.  Now that it's not quite as easy, maybe the use of a `requirements.txt` is not needed at all — just pass the requirements we'd like installed to pip as command-line arguments.

It might be cleaner to refactor a bit higher into `install_local_package` to make it clearer what is being done.  Perhaps:

```python
def install_local_package(package_root, path):
    """This installs a local dependency of a package."""
    # XXX: windows
    env = dict(os.environ)
    env["PYTHONPATH"] = package_root

    # Step 1: generate egg info and link us into the target folder.
    [...] (unchanged)

    # Step 2: install additional requirements
    for requirement in get_package_install_requires(path):
        # NB: this bastardizes the `package` argument of `download_and_install_package`
        # as we're passing a *requirement specifier* rather than just a package name.
        # I think it will work, but might want to adjust the signature of `download_and_install_package`
        # to provide a way to make this more explicit.  (While at it, perhaps provide a way to provide
        # multiple requirement specifiers in one go?)
        download_and_install_package(package_root, requirement)
```

The `get_package_install_requires` (probably) runs `setup.py egg_info` (as was done previously by `install_local_package`) and returns a sequence containing the initial lines — up to the first section marker ­of the resulting `requires.txt`.

Okay, I've rambled now, but, I think, have arrived at a worthwhile place:

The only reason `setup.py egg_info` is being run in `install_local_package` is to get at the packages `install_requires`.  Abstracting that whole process to a new function makes that logic clear.  Also, it abstracts implementation details — `install_local_package` doesn't care how the `install_requires` are obtained.

yagebu:
I understand that limitations of `pip` force us to do some of this by hand, but wouldn't it be better to use `pkg_resources` to read the egg-info? This rewriting between two different formats specified by two different tools seems rather fragile.

webknjaz:
> but wouldn't it be better to use `pkg_resources`

Even better: use `importlib_metadata`

dairiki:
@webknjaz I was unaware of the emergence of `importlib.metadata`.  Good to know!

There may well be a way to make them work — if so, that would be the way to go — however, I just played a bit with both `pkg_resources` and `importlib.metadata` and could not get either of them to work with a "half-installed" (--no-deps) package, like we are working with here.  (When run for a --no-deps installed package with missing dependencies, `pkg_resources.require(<pkg>)` raises `DistributionNotFound`, `metadata(<pkg>)` raises 'PackageNotFoundError`.)

(The original author of this code, @mitsuhiko, was, of course, well aware of the existence and limitations of `pkg_resources` when he hacked this up.  I had just assumed there was a good reason as to why he didn't go that way.)

lektor/packages.py
dwt: I've chosen this way of implementing this requires.txt -> requirements.txt cleanup step in a way that tries to be minimally invasive to the existing source code.

That being said, I too think that extracting the whole step including the `egg_info` generation would be much cleaner. As I now know that you're willing to take that change, I'll see what I can come up with.

dwt:
I think that instead of trying to make the hack in a more estethically pleasing way, it would be better to fix the actual upstream bugs so that the whole installation can be reduced to `pip install --editable --target $folder $package` and get rid of the hack completely. Until then I don't quite see much value in investing a whole lot of time into this code.

lektor/packages.py
dwt: @dairiki I've pushed a change that tries to get the spirit of what you said - what do you think about it?

lektor/packages.py
```diff
@@ -103,6 +103,22 @@ def download_and_install_package(
...
+
+    # Because of these bugs:
+    # - pip https://github.com/pypa/pip/issues/4390
+    # - setuptools https://github.com/pypa/setuptools/issues/392
+    # we cannot just call `pip install --target $folder --editable $package`.
+    # Hence the workaround of first installing only the package and then it's dependencies
+
+    editable_install_without_dependencies(package_root, path)
+
+    for requirements_path in requirements_from_unfinished_editable_install_at_path(
```
yagebu: I think using a for-loop here is a slight abuse of notation.

tests/test_packages.py
```diff
@@ -0,0 +1,111 @@
...
+        setup(
+            name='dependency',
+        )
+        """,
+    )
+
+    install_dir = tmp_path / "target"
+    install_dir.mkdir()
+
+    packages.install_local_package(install_dir.as_posix(), plugin_dir.as_posix())
```
yagebu, line 33: We're on Python 3.6 now, so passing these as Paths should work

lektor/packages.py
```diff
@@ -143,15 +161,33 @@ def install_local_package(package_root, path):
...
+            yield requirements_path
 
-        # We have dependencies, install them!
-        if os.path.isfile(requires):
-            download_and_install_package(package_root, requirements_file=requires)
     finally:
         shutil.rmtree(tmp)
 
 
+def requiriements_txt_from_requires_file(requires_path):
```
yagebu: typo

yagebu (changes requested):
I think it would be preferable if this code could do without the writing of `requirements.txt` - instead just read the requirements into a list and install that list of packages. That would make it easier to get rid of the for+yield hack and also make this part unit-testable.

dairiki:
In working on #881, I've discovered that installing packages into Lektor's package cache is pretty slow.  This means that we should minimize the number package installs we do in the tests.  Running your four tests on my reasonably fast machine takes about 7 seconds.

After PR #881, which cleans up the slow `test_plugins.py`, the complete test suite — 304 tests in all — takes about 12 seconds to run.  (That doesn't include your new tests.) So the overhead to run your four new tests is significant.

This is just something to keep in mind.  Writing code to make it unit-testable, and then relying on unit testing to cover all the various edge cases can really speed up the test run time.

lektor/packages.py
dwt: fixed

tests/test_packages.py
dwt: The original code works with string paths - so I didn't want to change that so as to not create a difference in how the code runs under tests vs. in real life. I can change it if it is important to you - but I wouldn't advise it.

lektor/packages.py
dwt: That is true, but I do need a way to clean up the temporary files afterwards. I could use a context manager for that, but then the empty case (i.e. no requirements) gets more complicated and needs another in band signal to not call installation there.

Not quite sure what the best way forward here would be. I could pull out the whole tempdir into the surrounding function and handle it there, that would clean it up - but also pollute the outer function with a concern it doesn't really need to know about.

dwt:
> I think it would be preferable if this code could do without the writing of `requirements.txt` - instead just read the requirements into a list and install that list of packages. That would make it easier to get rid of the for+yield hack and also make this part unit-testable.

I thought about that, and discarded it because I think that this would be a bad idea. Sorry, I should probably write this down as a comment in the source. The reason is this: pip needs to know about all the other packages that it is installing, so it can resolve dependencies correctly. If it doesn't know about the other packages, it can install incompatible versions, and especially versions that are different than what the plugin would install when it is installed un-editable, i.e. directly from pypi.

I think that is a hard no-go. That being said, I think the real push here should be to get the upstream bugs fixed, so that we can do `pip install --editable  $package_path --target $folder` and `pip install $package_name_from_pip --target $folder` to make the two cases the same.

Also, breaking the requirements down into individual items to be passed to pip on the command line requires more rewriting as requirements.txt and (to my knowledge probably requires.txt too) allow quite some additional syntax to specify constraints on the installed packages that the command line does not easily allow.

So… I think we need to stick with the requirements.txt, and / or go fix the upstream bugs (sooner).

dwt:
> In working on #881, I've discovered that installing packages into Lektor's package cache is pretty slow. This means that we should minimize the number package installs we do in the tests. Running your four tests on my reasonably fast machine takes about 7 seconds.
> 
> After PR #881, which cleans up the slow `test_plugins.py`, the complete test suite — 304 tests in all — takes about 12 seconds to run. (That doesn't include your new tests.) So the overhead to run your four new tests is significant.
> 
> This is just something to keep in mind. Writing code to make it unit-testable, and then relying on unit testing to cover all the various edge cases can really speed up the test run time.

That's a good argument. I could rewrite the tests to not actually call the install, but just verify that it is called correctly. That being said, I would like to argue that the test-suite should have a layered approach. Some integration tests should test the whole roundtrip, to make sure that we get notified about interface changes, or changes in the way arguments are understood, while most of the test-suite then can stub out the actual installation.

Which leaves the question: where should the round-trips be, and perhaps should they marked in a special way so pytest executes them after the other tests, or even in it's own test run?

My guess would be that the `packages.py` module could be the right place to do the roundtrip tests, and then stub that module in other unit tests to speed them up.

What do you think about that?

yagebu:
> The reason is this: pip needs to know about all the other packages that it is installing, so it can resolve dependencies correctly. If it doesn't know about the other packages, it can install incompatible versions, and especially versions that are different than what the plugin would install when it is installed un-editable, i.e. directly from pypi.

>Also, breaking the requirements down into individual items to be passed to pip on the command line requires more rewriting as requirements.txt and (to my knowledge probably requires.txt too) allow quite some additional syntax to specify constraints on the installed packages that the command line does not easily allow.

`pip` takes multiple on the command line just fine and also deals with PEP440 requirement specifiers, see https://pip.pypa.io/en/stable/reference/pip_install/#argument-handling, so I don't think these points are really problems with my suggested approach.

But maybe there's an even simpler approach that allows us to get rid of that whole egg-info parsing ugliness. Currently the approach is the following: 1) do an editable install the package without dependencies (because `--editable` and `--target` can't be combined) , 2) run setup.py egg_info 3) rewrite the requires from the egg-info and 4) install the dependencies from that. How about the following instead: 1) do a normal install of the package 2) do an editable install of the package without dependencies. That way, we install the package twice (so we trade one call to egg_info for an install of the plugin), but for a local plugin I'd not expect that shouldn't be a big performance hit. This would also allow this code to work for packages with an alternative build system.

Regarding the tests, I think it would be enough to have one test running a real pip install and then we can have the remainder just be unit tests (or some integration tests with the command calling stubbed out).

dairiki:
> How about the following instead: 1) do a normal install of the package 2) do an editable install of the package without dependencies.

That would be ideal if it can be made to work.  I've just played around with it a bit, without luck.  The issue is that the second `--editable` install doesn't uninstall the normally installed version of the distribution.  Both the normal and --editable version end up installed (and, sadly, it appears it is the normal version that python loads when the package is imported.)  Manually uninstalling it is problematic, too, as `pip uninstall` doesn't appear to support a `--target` argument.

----

In other news, I noticed there is a [documented function](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#parsing-utilities), `pkg_resources.split_sections` which can be used to parse `requires.txt`.
```
with open("requires.txt") as f:
    requirements = next(pkg_resources.split_sections(f))[1]
```
will work to get the list of requirements from just the first section (i.e. dropping the extras_require sections). (And it has the bonus of filtering out comments and blank lines.)

dwt:
@dairiki While I do think that splitting out the requirements with `pkg_resources.split_sections()` is probably preferable to parsing the file myself, so I will change that.

However, I would really like to reflect on the fact, that this discussion feels quite a bit like bike-shedding.

There is a clear bug here, the pull request addresses it - and I think it would be worth merging that to get the bug fixed - and then starting the work on getting things even more better after that.

Regarding `pkg_resources.split_requirements()`: It is probably better to use that function, but it is also not super simple, for example, it will skip the default section if there are only `extras_require` and it needs still more extra hand holding if a package has no requirements.

dwt:
I have now tried implementing the `split_sections()` way and have found that both it and the previous way of manually parsing the requires.txt file doesn't seem to be able to cope with more rare requires syntax, as setuptools seems to reformat a requirement of the form `watching_testrunner;python_version >= "3.6"` into this multi line monstrosity (of which I cannot find any documentation)

```ini

[:python_version >= "3.6"]
watching_testrunner
```
And which of course is neither compatible with pip requirements.txt as well as pip command line requirements.

Also going with handing in multiple  requirements to pip on the command line requires a new function, as the interface of the old one doesn't really suite itself to being extend to this.

So all in all I don't think this is worth it.

Also I would really like to stop the bike shedding and move on from this, so that I can finally get to the upstream libraries and fix the actual bug that causes all these workarounds.

dwt:
Here's hope that https://github.com/pypa/pip/pull/9636 gets through soon and all of this becomes obsolete in a few years when users can use that new pip version.

dwt:
Still, I would like to get this pull request through in some shape or form and not depend on the pip one, as no one knows how long that will take to propagate. So how do we continue here?

I would recommend to go back to a simple cleaning of requires.txt to generate requirements.txt.

Reasoning: AFAIK it is far more common to have `extra_requires` than environment markers, especially since setuptools deprecated test_requries, and suggested to use `extra_requires.test` as a replacement.

I think that this would make Lektor more usable until the pip pull request or an alternative gets through and we can get rid of the whole complication in lektor.

What do you guys think?

dairiki:
@dwt What's your short-term motivator for this?  Is it just that lektor-atom is broken?

If so, what needs to be done to fix lektor-atom?  Does anyone know who can make a release?  With lektor-atom appearing to be part of the lektor project, and with lektor-atom currently breaking the CI tests for lektor-website, this is a bad look for the project and should be addressed.

Some immediate local fixes (which I'm sure you realize) are to create a private fork of lektor-atom, delete the extras_requires, and include that as a git submodule in the `packages` directory of your site.  Alternatively, you could (temporarily, until this is resolved) remove lektor-atom from your `.lektorproject`, install `lektor` and `lektor-atom@master` to a virtualenv, and run lektor from that.

Another alternative which might be worth considering is to disuse lektor-atom altogether.  It's really pretty straightforward to write a jinja template to construct a simple atom feed directly — no plugin necessary.  (I can share an example template, if you'd like.)

dwt:
> @dwt What's your short-term motivator for this? Is it just that lektor-atom is broken?

Hm, maybe yes? I already have a local solution (the one you mentioned below) since quite some time. I would like to think that this is a bit of a more general problem as `test_requires` is deprecated and I really think that plugins ought to have a way to specify their testing-requirements in the setup.py file. I know you could just drop a requirements-test.txt in the project, but `extra_requires` are the recommended solution so I think we should support it until upstream fixes can get merged. (Which might take quite some time, as the discussion there shows)

> If so, what needs to be done to fix lektor-atom? Does anyone know who can make a release? With lektor-atom appearing to be part of the lektor project, and with lektor-atom currently breaking the CI tests for lektor-website, this is a bad look for the project and should be addressed.

That would be a good Idea regardless. If neccessary, I have worked on lektor-atom in the past and would be willing to take care of it, if given the neccessary rights. Not sure how you guys organise here, so maybe that's a no no?

All that being said - I think the pull request (maybe one or two commits back) still

- Improves the documentation dramatically why the code is structured the way it is
- Which bugs it depends upon to to get fixed before it can be removed
- Works a little bit better than before
- Without adding a lot of complexity.

I have to say, I think that this would be worth a merge.

dairiki:
This is mostly off-topic, but...

> but extra_requires are the recommended solution

Who says?  There is debate about that. Myself, I don't think [tests] extras_requres are the way of the future.

The `setup.py`, these days, doesn't specify *how* to run tests, so it doesn't make much sense to me that that should be where test dependencies are specified.

Back in the old (`distutils`) days, there was `python setup.py test` which made it easy to run a specific `unittest` test suite. `Tests_require` specified the dependencies that should be installed in order to run that test suite. When that was in use, that made sense, since setup.py contained the complete specification as to how to run the python tests for the project. That is now [deprecated][], with the advice “Users looking for a generic test entry point independent of test runner are encouraged to use tox.”

Tox seems to currently be the de facto method of running python tests these days.  It supports the concept of multiple isolated test environments — each of which can require a different set of dependencies.  As an example of how this is useful, one can easily imagine that a Lektor plugin package might well want to run integration tests against multiple different versions of Lektor.  In that case, there is no one list of test requirements.

My preference is to put as much of the python test configuration as possible — that includes testing dependencies, as well as how to actually run the tests — in `tox.ini`.  (The `tox` configuration can alternatively be put into `pyproject.toml` or `setup.cfg` should you want to minimize the number of config files. Personally, I like having a `tox.ini`, as it serves as a fairly obvious marker that the project uses tox for testing.)

[deprecated]: <https://setuptools.readthedocs.io/en/latest/userguide/commands.html#test-build-package-and-run-a-unittest-suite>

Note that the [setuptools docs][], while they do make mention of the possibility of a `[tests]` or `[docs]` extra, do explicitly call out “two common use cases” for `extras_require`:

1. When a specific `console_scripts` entry point depends on extra dependencies.
2. To allow another distribution to require the extras in their own dependencies.

A `[test]` extras does not fall under either of those two use cases.  I view extras_require (and the rest of `setup.py`) as specifying packaging metadata, rather than development metadata.  A user of the distribution does not care what’s required to run the tests — most often the tests are not included in the (wheel) packaging, so the user can not run the tests anyway — so the test dependencies do really belong in package metadata.


[setuptools docs]: <https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies>

dairiki:
As for the PR...

Keep in mind that I am not a project member and do not have merge/push rights, so this is just an opinion from the peanut gallery.

Also, keep in mind that, as already demonstrated, I am old and grumpy.  I don't mean to be overly critical.

Two points.

First, for some reason that I can't fully put into words, the idea of reading `requires.txt`, pulling the desired dependencies from that, writing them to `requirements.txt` and then passing that to pip sticks in my craw.  Writing of `requirements.txt` seems to be an extra step that serves no purpose.  I'd really rather see the requirements parsed from `requires.txt` and then those passed as command-line arguments to pip.

Personally, I would change the signature of `download_and_install_package` from
```python
def download_and_install_package(
    package_root, package=None, version=None, requirements_file=None
):
```
to
```python
def download_and_install_package(package_root, packages):
```
The only reason it currently supports passing `requirements_file` was to directly pass it the `requires.txt` from the egg-info.

Also moving the logic for forming the full requirement spec from `package` and `version` further up in the call chain will help if, at some time in the future, someone wants to put in the work to support more flexible ways of specifying version requirements in `.lektorproject`.  E.g. it would be nice to be able to do something like:
```ini
[packages]
lektor-plugin = >=1.0,<2
```

Second, the idea of adding code to "fix" something that mostly works, but only kind of by luck, with something that works a bit better, but still is known to be not fully correct or compatible does not give me warm fuzzy feelings.  Particularly when there are work-arounds which are (perhaps debately) not overly onerous.

dwt:
Updated code according to the comments from @dairiki 

How are the @maintainers feeling about merging this? Is there still some more work required to make the test-suite faster? I somehow consider this something that can pretty much go completely away as soon as we can switch to `pip install --target $folder [--editable] $package_name_or_path_to_package`, as that would remove almost all of the workarounds. Till then, I kind of want the tests as a reminder that all the hackery a) still works and b) is a pain to maintain and should rather be removed ASAP.

What do you think?

lektor/packages.py
dwt: For loop is gone in the current iteration.

lektor/packages.py
```diff
@@ -75,10 +75,8 @@ def remove_package_from_project(project, name):
     return None
 
 
-def download_and_install_package(
-    package_root, package=None, version=None, requirements_file=None
-):
-    """This downloads and installs a specific version of a package."""
+def download_and_install_packages(package_root, packages_with_versions):
```
dairiki: I'd suggest the name `requirements` (or perhaps `requirement_specifiers`), rather than `packages_with_versions` since that what pip [calls them](https://pip.pypa.io/en/stable/reference/pip_install/#requirement-specifiers).

lektor/packages.py
```diff
@@ -292,7 +332,10 @@ def update_cache(package_root, remote_packages, local_package_path, refresh=Fals
                     package_root, os.path.join(local_package_path, package[1:])
                 )
             else:
-                download_and_install_package(package_root, package, version)
+                download_and_install_packages(
+                    package_root,
+                    ["%s%s%s" % (package, version and "==" or "", version or "")],
```
dairiki: I know you just moved this code from where it used to be, but I'd rather see this as:
```python
requirement_spec = f"{package}=={version}" if version else package
download_and_install_packages(package_root, [requirement_spec])
```

(Perhaps a full `if` ... `else` ... to compute `requirement_spec` would be even more readable.)

lektor/packages.py
```diff
@@ -143,14 +166,31 @@ def install_local_package(package_root, path):
...
+            # We have dependencies, install them!
+            return requirements_from_requires_file(requires_path)
     finally:
         shutil.rmtree(tmp)
 
+    # no dependencies to install
+    return None
+
+
+def requirements_from_requires_file(requires_path):
```
dairiki, line 181: (Is there a good reason not in in-line this function within `requirements_from_unfinished_editable_install_at_path`?)

lektor/packages.py
```diff
@@ -292,7 +332,10 @@ def update_cache(package_root, remote_packages, local_package_path, refresh=Fals
                     package_root, os.path.join(local_package_path, package[1:])
                 )
             else:
-                download_and_install_package(package_root, package, version)
+                download_and_install_packages(
+                    package_root,
+                    ["%s%s%s" % (package, version and "==" or "", version or "")],
+                )
```
dairiki: (I wonder if there's a reason not to accumulate all of the non-local packages and install them in a single pip invocation, rather than installing them one at a time.)

dairiki (commented):
FWIW, I like the current version much better.  Of course, I did have to leave a few comments...

Finally, given pypa/pip#9636 (great work!) is in the works, is it worth worrying about this at all, or would it be better to wait until we can just rely on pip to do the right thing?  (What's the chance Lektor will make a release before pip does?)

yagebu (commented):
I'd prefer to not merge this, it still seems like a rather hackish stop-gap solution.

I believe that the underlying problem that we should address instead is our use of editable installs. I think they're too dangerous to do without explicit use action, since they pollute the python import path with whole packages directories. Their only advantage for Lektor is keeping track of changes, but that only works for code and not other packages changes, so I think doing normal install with some more explicit change tracking would be better. I'll open a (draft) PR removing our use of editable installs.

lektor/packages.py
dwt: [x] done

lektor/packages.py
dwt: I think a full if:else doesn't add much readability. Still I would really like to keep this patch somewhat minimal, given that it might become obsolete soon anyway.

lektor/packages.py
dwt: I think the code reads and explains better what it does this way - the function is mostly self contained and doesn't need the context of the other function where it is used to be understood.

lektor/packages.py
dwt: I would say that is something for a different pull request. This one should keep it's focus on fixing one thing first.

dwt:
> FWIW, I like the current version much better. Of course, I did have to leave a few comments...

I hope those are addressed adequately in my commits and comments.

> Finally, given [pypa/pip#9636](https://github.com/pypa/pip/pull/9636) (great work!) is in the works, is it worth worrying about this at all, or would it be better to wait until we can just rely on pip to do the right thing? (What's the chance Lektor will make a release before pip does?)

I think that it is not quite as simple as waiting for a new pip release, as that would also mean that we have to add a dependency on that pip version and throw an error if the user has an older version installed.

I would do that in a separate (future) pull request, once pip is released and we had time to actually test it.

dwt:
> I believe that the underlying problem that we should address instead is our use of editable installs. I think they're too dangerous to do without explicit use action, since they pollute the python import path with whole packages directories. Their only advantage for Lektor is keeping track of changes, but that only works for code and not other packages changes, so I think doing normal install with some more explicit change tracking would be better. I'll open a (draft) PR removing our use of editable installs.

Well, I don't think there is a good way to avoid editable installs of plugins you are working on completely. I don't know the history of why lektor did it's own editable installs, but I guess it could have something to do with the App GUI story that was present in the past? In that scenario, Lektor didn't really expose the virtualenv / place where the packages where exposed and needed to manage editable installs itself?

Some arguments that come to mind:
- Someone working on a plugin probably knows how to install Lektor himself into a virtualenv
- Someone distributing private plugins as submodules or in tree code as part of a site probably doesn't want collaborators to need to know how to install plugins himself (but for those cases a static install might be good enough?

As for something completely different: It might be possible to `pip install $plugin; pip uninstall $plugin --no-deps` to get jus the dependencies installed and then just add the `$LEKTOR_PROJECT/packages` folder to the python path?

webknjaz:
Here's a wild thought: what if instead of requiring a plugin to be editable-installed, Lektor could just have a local plugin discovery mechanism that injects paths into `$PYTHONPATH`/`sys.path`?

This would make it importable and as for the deps, it's possible to monkey-patch `setuptools.setup`+`sys.argv` and invoke `setup.py` via something like `runpy.run_module('setup.py', name='__main__')`. Then the patched function would just collect the deps from the `install_requires` arg.

On the second thought, it could be easier to just require the plugin authors to use `setup.cfg` (which means requiring some minimum version of setuptools) — this would make parsing the data so much easier and would establish a better standard for describing the modules.

Another extreme would be relying on PEP517 to get the deps (there are hooks for this IIRC) but I think this doesn't allow one to extract the entrypoints as easily.

--------------------

@dairiki
> This is mostly off-topic, but...
>
> > but extra_requires are the recommended solution
>
> Who says? There is debate about that. Myself, I don't think [tests] extras_requres are the way of the future.

A part of the PyPA feels like having `tests` extras is a good idea. IIRC you can find a discussion about this on https://discuss.python.org. But then, no tools support installing extras-only, w/o the package itself. I'm in the camp of not using an extra for this too: test requirements should describe the env and therefore they should be strictly pinned unlike the open-ended runtime requirement specs, I find pip-tools to be a great solution for this problem.

dairiki:
> I believe that the underlying problem that we should address instead is our use of editable installs. I think they're too dangerous to do without explicit use action, since they pollute the python import path with whole packages directories. 

Personally, since we're installing code from a single project's `packages` subdirectory into a private package library which is used only for that Lektor project, I don't see "pollution" as an issue.

The [docs make it clear](https://www.getlektor.com/docs/plugins/dev/) that some of the original motivation was to make it easy to work on the project-local plugins with changes to the code taking effect immediately (at least when running with `LEKTOR_DEV=1`.)

> Here's a wild thought: what if instead of requiring a plugin to be editable-installed, Lektor could just have a local plugin discovery mechanism that injects paths into $PYTHONPATH/sys.path?

> Another extreme would be relying on PEP517 to get the deps (there are hooks for this IIRC) but I think this doesn't allow one to extract the entrypoints as easily.

Without having thought about it too much, this seems like a great idea.  [Build a .dist-info](https://www.python.org/dev/peps/pep-0517/#prepare-metadata-for-build-wheel) in the source directory and add the directory to sys.path.
(That is all beyond the scope of this PR, though.)

dwt:
@maintainers: I'd like to get this moving again.

Some points I would like to make:
- This pull request fixes a problem with existing released plugins
- It makes the code structure better, and more amenable to future expansion and fixes
- It also prepares to an eventual simplification when my fixes to pip are released
- It adds comments that explain why the code contains the hacks that it does
- It adds test cases for all these issues

I don't mind that much of this code gets removed in the long run, but I think it is an extremely potent base from which to start further refactorings from.

So I would like to get this merged and then continue working on further approaches. Because incremental improvement always wins.

dairiki:
Closing in favor of #1065.

PR #879, opened 2021-02-11
Fix exception message to make it a bit more helpful.

### LINKED ISSUE #875: Is there a reason Lektor is so strict about plugin distribution names?
This pertains to the relationship between plugin name and package name which is enforced by [lektor.pluginsystem.load_plugins()](https://github.com/lektor/lektor/blob/ce4d96be0af2703059109b2f1b7c491e53280351/lektor/pluginsystem.py#L115).

## Motivating Story

I have  a package called [excerpt-html](https://github.com/dairiki/excerpt-html), which, essentially contains one python function (not surprisingly named `excerpt_html`) whose purpose is to compute excerpts (or "summaries") of HTML text fragments.
I thought it would be nice to include a simple Lektor plugin in the package. The plugin would register a single jinja filter (also called `excerpt_html`) which would allow excerpts to be computed within jinja templates.

So I write a lektor plugin, and add something like this to `setup.py`:
```python
entry_points={
    'lektor.plugins': [
        'excerpt-html = excerpt_html.lektor:ExcerptHtmlPlugin',
    ]
}
```

When I go to use this, this error message obtains:
```
lektor/pluginsystem.py:123: in initialize_plugins
E RuntimeError: Mismatching entry point name.  Found "excerpt-html" but expected "-html" for package "excerpt-html".
```

## Points and Questions

First off, note that the error message is rubbish.  The problem is not the name of the entry point, but rather (for the most part) the distribution name of my package.  There seems to be no way to pass muster unless the distribution name starts with `lektor-`. (The message claims that Lektor expects the plugin to be named `-html`, but it’s clear that it won’t be happy with that name either.)

Second, is there really a good reason to require a specific relationship between distribution name and plugin name?  Why shouldn't I be able to register a plugin named `excerpt-html` in my `excerpt-html` distribution?

It seems silly to have to create a whole new distribution (`lektor-excerpt-html`) just to package the (essentially three-lines of) Lektor plugin code.  Do I have any better options?

dairiki:
This fixes the exception message emitted when a Lektor plugin is detected from a distribution with a non-conforming name.

See lektor/lektor#875 for further discussion.

### Issue(s) Resolved

The current message is incorrect and non-elucidating in many cases.

Fixes #875

### Description of Changes

### DIFF
--- a/lektor/pluginsystem.py
+++ b/lektor/pluginsystem.py
@@ -115,9 +115,10 @@ def load_plugins():
         match_name = "lektor-" + ep.name.lower()
         if match_name != ep.dist.project_name.lower():
             raise RuntimeError(
-                "Mismatching entry point name.  Found "
-                '"%s" but expected "%s" for package "%s".'
-                % (ep.name, ep.dist.project_name[7:], ep.dist.project_name)
+                "Disallowed distribution name: distribution name for "
+                "plugin {ep.name!r} must be {match_name!r}.".format(
+                    ep=ep, match_name=match_name
+                )
             )
         rv[ep.name] = ep.load()
     return rv
--- a/lektor/pluginsystem.py
+++ b/lektor/pluginsystem.py
@@ -116,9 +116,7 @@ def load_plugins():
         if match_name != ep.dist.project_name.lower():
             raise RuntimeError(
                 "Disallowed distribution name: distribution name for "
-                "plugin {ep.name!r} must be {match_name!r}.".format(
-                    ep=ep, match_name=match_name
-                )
+                f"plugin {ep.name!r} must be {match_name!r}."
             )
         rv[ep.name] = ep.load()
     return rv
--- a/tests/test_plugins.py
+++ b/tests/test_plugins.py
@@ -2,6 +2,7 @@ import os
 import shutil
 import textwrap
 
+import pkg_resources
 import pytest
 
 from lektor.cli import cli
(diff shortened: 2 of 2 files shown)


dairiki:
As an aside, note that the use of [`pkg_resources.__{get,set}state__`](https://github.com/lektor/lektor/blob/5c1f829710eb187e346e5ff48ef3a6efc8e1163a/tests/test_plugins.py#L261) in the new unit test points to a way that the other tests in this file could be cleaned up (ref [this comment](https://github.com/lektor/lektor/blob/5c1f829710eb187e346e5ff48ef3a6efc8e1163a/tests/test_plugins.py#L106).)
If there is interest, I could work up a PR to do that.

Note that while `__getstate__` and `__setstate__` do not appear to be a documented part of the `pkg_resource` API, they are used in `setuptools`' [own tests](https://github.com/pypa/setuptools/blob/b6bbe236ed0689f50b5148f1172510b975687e62/setuptools/tests/contexts.py#L82), and appear to have been a stable feature [since 2011](https://github.com/pypa/setuptools/commit/d76ad0a7d55448992833c36ecbc1899ae0e17683).

yagebu (approved):
Thanks for the PR and providing this fix even though you disagree with the behaviour of Lektor here :)

This LGTM and gets bonus points for using f-strings ;)

> As an aside, note that the use of pkg_resources.__{get,set}state__ in the new unit test points to a way that the other tests in this file could be cleaned up (ref this comment.)
If there is interest, I could work up a PR to do that.

Yes, sure!

Issue #878, opened 2021-02-11
Consider dropping python3.5 support as it is end of life since some time


dwt:
3.6 is the current oldest version that still receives patches. AFAIK

runfalk:
Python 3.5 is actually dropped as per the latest release notes: https://www.getlektor.com/blog/2020/8/lektor-32-released/

dairiki:
It appears that 3.5 is still listed in the CI test matrix.

https://github.com/lektor/lektor/blob/194b023c3020b5cc4b7f03c4b52ba9134b0da165/.github/workflows/ci.yml#L219

runfalk:
True, that's an artifact of that 3.2 is still supporting Python 3.5. I assume we'll remove it in the near future.

PR #880, opened 2021-02-13
require Python 3.6 and remove Python 3.5 from CI

yagebu:
Ref #878 

That we'd drop Python 3.5 has already been announced with the release of Lektor 3.2, so this PR removes it from the CI matrix and adds the requirement for 3.6.

[I saw that 3.5 compat caused some extra work in #877 so I think it's time to make this change]


runfalk:
Sounds good to me.

PR #881, opened 2021-02-16
Refactor tests/test_plugins.py

dairiki:
Some test "cleanup", as promised in [#879](https://github.com/lektor/lektor/pull/879#issuecomment-778657494).

This is a work in progress.  So far several changes have been made.

- Improve test-to-test isolation in terms of changes of state to `pkg_resources`, `sys.path`, and `sys.modules`.

- Unify several different fixtures whose purpose was to set up test sites with a plugin to a single fixture.

- Disuse the indirect pytest fixture pattern.

- Use a temporary directory for Lektor’s cache when running tests.  (Previously, tests used and left remains behind in `~/.cache/lektor/{packages,builds}`.)


dairiki:
I think I would like to refactor `test_plugins.py` even more.

Currently, most of the tests in the file are integration tests.  They essentially run the `lektor` command line command to ensure the `extra_flags` can be passed to plugins from the command line.  That's fine, as far as that goes, but these tests are run quite a number of times to further check that a number of events are emitted when `lektor build` and `lektor clean` is run.  That could be tested in a single run of each of those commands.  (Running the full command is slow — currently it takes 30 seconds to run just `test_plugins` on my reasonably high-end computer.)

Meanwhile, there is quite a bit of `lektor.pluginsystem` which is not covered at all by tests.  (Coverage is 64%.)

I propose to add some real unit tests, in order to increase the test coverage of `lektor.pluginsystem`, as well as reduce the number of integration tests run, so as to speed up the tests.

Comments?  Questions?

yagebu:
Yes, that sounds like a very good idea! In my quick testing, these tests also took nearly 30s and about two thirds of the time of all of Lektor's tests, which is really too slow. Your outlined plan sounds like it could really cut that down while actually improving these tests :)

nixjdm:
Great work, and thank you! Good find on `pkg_resources.__getstate__` / `__setstate__`. Being used in their tests since 2011 does make that seem plenty stable enough to me.

I'd also be happy to see a lot of this become unit instead of integration tests.

dairiki:
I've just refactored again.  This converts almost all tests to true unit tests.

One integration test is preserved to check that extra_flags make it from the command line to the plugin.

It turns out that the primary cause of slowness of the previous tests is in setting up Lektor’s private package cache.
Installing even a simple distribution takes on the order of a second.  The current tests build a new private package cache, installing the test plugin distribution, for each test — so about a second per test.

Of the tests in this PR, there is only one test ([this one][]) that builds a package cache.  It is slow and it requires a working internet connection (to PyPI) in order to install a plugin.  I don't like it, but couldn't figure out another way to get to 100% code-line coverage for `lektor.pluginsystem`.  Perhaps that test could be omitted (and we can live without 100% coverage).  Or it could be pytest-marked so that it is easy to skip (or so that it is skipped by default?)

[this one]: <https://github.com/lektor/lektor/blob/ba3e9a6ff0817f3489ee585532129c4803269c36/tests/test_pluginsystem.py#L155>


----

The second commit in this PR completely overwrites the changes made in the first one.  If you like it, let me know before you merge it.  I'll squash the two commits to get rid of the first commit altogether.

yagebu:
> Perhaps that test could be omitted (and we can live without 100% coverage). Or it could be pytest-marked so that it is easy to skip (or so that it is skipped by default?)

Your call - I'd be fine with omitting it (if we keep it, it shouldn't be skipped IMHO).

dairiki:
> Your call - I'd be fine with omitting it (if we keep it, it shouldn't be skipped IMHO).

Okay,  I'll leave it in for now.  If it causes problems down the road it can be deleted (or conditionally skipped.)

dairiki:
I think this PR is ready to be merged.  Let me know if it needs anything else.

tests/conftest.py
```diff
@@ -13,6 +17,79 @@
...
+def temporary_lektor_cache(tmp_path_factory):
+    """Get Lektor to use a temporary cache directory.
+
+    This prevents the tests from leaving scats behind in the
+    user’s real cache directory.
+
+    """
+    cache_dir = tmp_path_factory.mktemp("lektor_cache")
+
+    # NB: lektor.utils.get_cache_dir() does not make it easy to configure
```
yagebu: I think it's fine that we need to use monkeypatch to change the cache directory. I wouldn't want to add an environment variable that we'd just expect to use internally for tests - if you have other uses for it, that's another story of course. In any case, I think this comment could be shortened a bit, it currently sounds like an open TODO.

yagebu (changes requested):
Just a minor comment about the monkeypatching of the cache directory, otherwise it LGTM

tests/conftest.py
dairiki: Okay, I'll just delete that part of the comment.

It seems like it would be polite to provide a general way to configure the location of Lektor’s cache directory, but I don't have a particular use-case in mind.  ... so ... never mind!

yagebu (approved):
Thanks again :) This PR gets bonus points for completely refactoring `test_plugins` but keeping the line count at exactly 298.

dairiki:
> Thanks again :) This PR gets bonus points for completely refactoring `test_plugins` but keeping the line count at exactly 298.

That's pretty funny!  I hadn't noticed that.  It was totally unintentional.

PR #882, opened 2021-02-18
Fix reference cycle in Environment

dairiki:
There is a reference cycle in instances of `Environment` which prevents them from being quickly garbage collected.

Fixing it is probably not of high importance, since the garbage collector will eventually find discarded instances.
In this case, however, the fix is easy, and since other parts of the Lektor code appear to go to some trouble to
hold weakrefs to the environment instances, I figure it might as well be fixed.

The reference cycle is as follows:  In [`Environment.__init__`](https://github.com/lektor/lektor/blob/25ba7c7df2646078ae6667781e51444ee305bf8e/lektor/environment/__init__.py#L121), the bound method `self.select_jinja_autoescape` is passed to the constructor for `CustomJinjaEnvironment`.  Bound methods hold a reference to the object they are bound to. The resulting jinja environment is saved in `self.jinja_env`, completing the cycle.

There appears to be no reason that `Environment.select_jinja_autoescape` can not be made a static method.  Making it so fixes the cycle.

### DIFF
--- a/lektor/environment/__init__.py
+++ b/lektor/environment/__init__.py
@@ -260,7 +260,8 @@ class Environment:
         )
         return values
 
-    def select_jinja_autoescape(self, filename):
+    @staticmethod
+    def select_jinja_autoescape(filename):
         if filename is None:
             return False
         return filename.endswith((".html", ".htm", ".xml", ".xhtml"))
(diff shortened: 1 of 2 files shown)


yagebu (approved):
Thanks!

PR #884, opened 2021-02-18
Upgrade Bootstrap from v3 to v4

### LINKED ISSUE #648: Date picker overflows admin page width
It seems that the date picker widget in it's current form always occupies a set minimum width. Therefore, it can overflow the admin page body in some cases, as shown in the image blow:

date-overflow
(image taken from the the blog page template from the example project)

I was only able to "fix" this by decreasing the font size of the date picker, which is not an ideal fix.

This is for screen widths between 1024px and 1365px.

yagebu:
We're still currently using the long outdated version v3 of Bootstrap. This PR upgrades Bootstrap to the current stable v4 (v5 is in development). This gets rid of a `npm audit` warning due to a dependency on an old jQuery and switches our CSS preprocessor from LESS to SCSS.

I've tried to keep the visual changes to a minimum. There's a difference in some details, but in most of these cases it seems to fix some broken things like the bug linked to below. The only bigger change that I made, which is also well visible in the screenshots below, is the removal of Lektor's custom grid and screen sizes. I'd expect Bootstrap to get these defaults right better than us... Also, the current admin UI doesn't really work well on mobile, which is something I'm planning to change, and I expect that to be easier if we don't modify Bootstrap too heavily.

Fixes #648

## Before
Screenshot_2021-02-18 Lektor Admin(1)

## After
Screenshot_2021-02-18 Lektor Admin


dairiki:
Very nice!


I haven't looked at this very hard, but I do have one comment:

Is there an easy way to make those margins narrower?  (We can use screen space we can get, right?)

Switching [two][1] [instances][2] of `.container` to `.container-fluid` seems to mostly work.  (There might be a few other places that also need tweaking.  I haven't looked very hard.)

Maybe tweaking `$container-max-widths` (and/or `$grid-breakpoints`) would work, too.

[1]: <https://github.com/yagebu/lektor/blob/b9a79db16aec7b5c6cad55427fd676911f9a01a1/lektor/admin/static/js/views/App.tsx#L44>
[2]: <https://github.com/yagebu/lektor/blob/b9a79db16aec7b5c6cad55427fd676911f9a01a1/lektor/admin/static/js/header/BreadCrumbs.tsx#L97>

runfalk:
Very nice indeed! I have a tiny observation from your before and after screenshots.

lektor

I can see that the menu is also quite a bit closer to the "content", but I think that looks fine too. I skimmed the changes and think it looks OK, but the `/* this really should not go here, it's the wrong place but oh well */` sticks out a little :smile:.

yagebu:
> (We can use screen space we can get, right?)

I'm not so sure about that, having the content be too wide is also not ideal IMHO. I prefer the new slimmer variant and think that in the previous one, the textarea for the body field was unhelpfully wide. This might also just be a personal preference, but as I stated in the PR, I think sticking to Bootstraps default is a solid choice.

> Very nice indeed! I have a tiny observation from your before and after screenshots.

I partially fixed that (on the right side). Since the Bootstrap "breadcrumbs" component has some default padding I didn't modify it on the left. More fine-tuning of this can be done later - there's more issues on smaller screens, but I want to do that in a later PR.

> I skimmed the changes and think it looks OK, but the /* this really should not go here, it's the wrong place but oh well */ sticks out a little 

That was just copy-pasted over, so don't shoot the messenger :smile:

dairiki:
>  I'm not so sure about that, having the content be too wide is also not ideal IMHO.

@yagebu  I see no reason to think that bootstrap defaults are necessarily the best choice in this case. This is a GUI, not a regular webpage, so normal web-design criteria don't necessarily apply.   (How would you feel if your editor or your xterm or your file manager added extra margins to its window?)

In my case, I often have page models with a lot of small bits of metadata.  (Often these are event pages, and the model includes all the bits needed to construct the google-structured data for it: startDate, endDate, eventStatus, previousStartDate, &c.)  It's nice to have these (related) fields displayed in a minimal number of rows on the editor screen, rather than one or two per row.  (It's good to be able to see all the bits at one time without having to scroll to get to them all.)  So I like having more width to play with.

Also, having as wide an area as possible in which to render the page preview (which may, of course, and rightfully, have its own page margins) is a plus.

Again, it's a GUI.  If it's too wide, the user can always resize their browser window.

If textarea-too-wide is really a problem, alternative solutions might be:
- add max-width (or some other similar style-related) model field option
- Put a drag-handle on the text-area, allowing the user to resize it.  (I wouldn't mind this for other reasons.  Currently the textarea seems to resize to contain the entire content.  If the content is very large, this is not necessarily ideal.  E.g. if I really only want to edit the metadata at the bottom of the edit page, having to scroll past the huge textarea is a pain.  Even if I am editing part of a big textarea, I really only need to be able to see the section of it that I'm editing.)

yagebu:
@dairiki  I've restored the previous maximum width (+- a few pixels to get some round numbers). I probably still think it's too wide but I'll not make that change in this PR.

PR #886, opened 2021-02-19
Fall back to PollingObserver if creation of default Observer fails.

### LINKED ISSUE #861: Use 'watchgod' module instead of 'watchdog'
I'm just starting to use Lektor but i noticed that it uses the `watchdog` module and one of the first comments that i saw at the [quickstart tutorial](https://www.getlektor.com/docs/quickstart/) was:
> When I run lektor server I get OSError: inotify watch limit reached

I ran into that myself and thus i know what it is and that it's related to `watchdog` but a lot of beginners probably don't know. Instead of using separate approaches for each OS to watch for file-changes, i think that using the [watchgod](https://github.com/samuelcolvin/watchgod) module could be a better and simpler solution.

Maybe it's a good idea to at least check the module out and see if it fits better than the overly complex watchdog implementation.

dairiki:
The `watchdog` distribution provides several different types of "Observers" that can be used to watch for file changes. On Linux, the default observer is `InotifyObserver`. There are other observers which are used on different OSes. There is also a `PollingObserver`, which  — while not as efficient as the other observers — should work anywhere.

Sometimes creation of the `InotifyObserver` fails.  This can be caused by inotify resource limits being exhausted by other processes running on the users system, for example.

This change makes it so that if creation of the default `Observer` fails, Lektor will fall back to using a `PollingObserver` (rather than just crapping out.)

### Issue(s) Resolved

Fixes #861

### Concerns

I suspect that is still possible that exceptions due to inotify limits may still be possible after the Observer has been created, while it is in use.  If that is indeed a possibility, this does not fix that.

### Description of Changes

### DIFF
--- a/lektor/watcher.py
+++ b/lektor/watcher.py
@@ -93,16 +93,10 @@ class BasicWatcher:
         return True
 
     def __iter__(self):
-        # Alias this since we may need it during interpreter shutdown
-        queue_Empty = queue.Empty
-
         while 1:
-            try:
-                item = self.event_handler.queue.get(timeout=1)
-                if self.is_interesting(*item):
-                    yield item
-            except queue_Empty:
-                pass
+            time_, event_type, path = self.event_handler.queue.get()
+            if self.is_interesting(time_, event_type, path):
+                yield time_, event_type, path
 
 
 class Watcher(BasicWatcher):
--- a/lektor/watcher.py
+++ b/lektor/watcher.py
@@ -11,9 +11,6 @@ from watchdog.observers.polling import PollingObserver
 
 from lektor.utils import get_cache_dir
 
-# Alias this as this can be called during interpreter shutdown
-_Empty = queue.Empty
-
 
 class EventHandler(FileSystemEventHandler):
     def __init__(self):
@@ -76,12 +73,15 @@ class BasicWatcher:
         return True
 
     def __iter__(self):
+        # Alias this since we may need it during interpreter shutdown
+        queue_Empty = queue.Empty
+
         while 1:
             try:
                 item = self.event_handler.queue.get(timeout=1)
                 if self.is_interesting(*item):
                     yield item
-            except _Empty:
+            except queue_Empty:
                 pass
 
 
--- a/lektor/watcher.py
+++ b/lektor/watcher.py
@@ -1,6 +1,8 @@
 import os
 import queue
 import time
+from collections import OrderedDict
+from itertools import zip_longest
 
 import click
 from watchdog.events import DirModifiedEvent
@@ -25,6 +27,16 @@ class EventHandler(FileSystemEventHandler):
             self.queue.put((time.time(), event.event_type, path))
 
 
+def _fullname(cls):
+    """Return the full name of a class (including the module name)."""
+    return f"{cls.__module__}.{cls.__qualname__}"
+
+
+def _unique_everseen(seq):
+    """Return the unique elements in sequence, preserving order."""
+    return OrderedDict.fromkeys(seq).keys()
+
+
 class BasicWatcher:
     def __init__(self, paths, observer_classes=(Observer, PollingObserver)):
         self.event_handler = EventHandler()
@@ -33,19 +45,21 @@ class BasicWatcher:
         self.observer = None
 
     def start(self):
-        untried = set(self.observer_classes)
-        for observer_class in self.observer_classes:
+        observer_classes = list(_unique_everseen(self.observer_classes))
+        for observer_class, next_observer_class in zip_longest(
+            observer_classes, observer_classes[1:]
+        ):
             try:
                 self._start_observer(observer_class)
+                return
             except Exception as exc:
-                untried.discard(observer_class)
-                if len(untried) == 0:
+                if next_observer_class is None:
                     raise
                 click.secho(
-                    f"Creation of {observer_class.__module__}.{observer_class.__name__} "
-                    f"failed with exception:\n  {exc.__class__.__name__}: {exc!s}\n"
-                    "This may be due to a configuration or other issue with "
-                    "your system.\nFalling back to polling for file modifications.",
+                    f"Creation of {_fullname(observer_class)} failed with exception:\n"
+                    f"  {exc.__class__.__name__}: {exc!s}\n"
+                    "This may be due to a configuration or other issue with your system.\n"
+                    f"Falling back to {_fullname(next_observer_class)}.",
                     fg="red",
                     bold=True,
                 )
--- a/lektor/watcher.py
+++ b/lektor/watcher.py
@@ -2,10 +2,12 @@ import os
 import queue
 import time
 
+import click
 from watchdog.events import DirModifiedEvent
 from watchdog.events import FileMovedEvent
 from watchdog.events import FileSystemEventHandler
 from watchdog.observers import Observer
+from watchdog.observers.polling import PollingObserver
 
 from lektor.utils import get_cache_dir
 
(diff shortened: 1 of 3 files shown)


dairiki:
Did I mention that I *hate* these enforced pre-commit hooks, in particular the blacking and reorder-python-imports?

This is failing the Lint CI test due to reorder-python-imports, however, reorder-python-import passes locally for me on my dev machine.  I'm not exactly sure what is going on, but I suspect it's unhappy with the location of the `import py` line.

I'll try pushing a trial-and-error commit to see if that might be what it is...

dairiki:
Well, moving the `import py` seems to have fixed the pre-commit failure.

Note that now this fails pre-commit on my dev machine.  I don't know why reorder-python-imports behaves differently in the two environments.

yagebu:
Does this really error when the Observer is created? In https://github.com/gorakhargosh/watchdog/issues/673 it seems like this errors once the observer is started.

dairiki:
@yagebu Erff.  You're probably right.  I will investigate further.

dairiki:
@yagebu You were absolutely right.  Good catch.

I've just force-pushed another try.

I've now tested this by running `sudo sysctl fs.inotify.max_user_watches=0` and making sure that `lektor server` and `lektor build --watch` still run (and that the big, red "OSError: [Errno 28] inotify watch limit reached" warning is emitted).

A question:

The `callback` argument of [lektor.watcher.BasicWatcher](https://github.com/lektor/lektor/blob/f30b62827b11fe90217d8d8e5eea3b6433f35b18/lektor/watcher.py#L38) (and `lektor.watcher.EventWatcher`), though it has been there since the initial commit in this repo, appears never to have been used.

May I just delete it?  This would get rid of a bit of code.

If you like the rest of this, don't merge it quite yet.  I'll add a few more docstrings and tests...

dairiki:
I have no idea why pylint doen't like my last commit.  Its complaints appear to have nothing to do with any code I touched — and it all passed 19 hours ago before the small commit.

dairiki:
Apparently, the change to the newly-released pylint 2.7.0 is what triggered (at least most of) these new errors.
Really, the version of pylint used for CI and tox testing should be pinned so that we don't have to aim for a moving target.

yagebu:
> The callback argument of lektor.watcher.BasicWatcher (and lektor.watcher.EventWatcher), though it has been there since the initial commit in this repo, appears never to have been used.

> May I just delete it? This would get rid of a bit of code.

Yes, sounds good. I'd be surprised if someone external used that code.

lektor/watcher.py
```diff
@@ -2,59 +2,86 @@
...
+        # Alias this since we may need it during interpreter shutdown
+        queue_Empty = queue.Empty
+
         while 1:
             try:
                 item = self.event_handler.queue.get(timeout=1)
                 if self.is_interesting(*item):
                     yield item
-            except _Empty:
+            except queue_Empty:
```
yagebu: Maybe do a `from queue import Empty`, then we can do without the alias (and then the comment could be moved to the except, to ensure that it is not changed back)

yagebu (commented):
Left just one minor comment, will give it a full review later this week.

lektor/watcher.py
runfalk: `from queue import Empty as QueueEmpty` maybe?

dairiki: > Maybe do a `from queue import Empty`, then we can do without the alias

Well (as you know) it's essentially the same thing.  My feeling is that using the assignment to create the reference to `queue.Empty` better indicates that we're doing it for a different reason than just to import it into the namespace.

> `from queue import Empty as QueueEmpty` maybe?

I like that better, though again, the purpose of what we're doing is not the same as a normal `import`, so using an assignment better expresses that notion. (I'm fairly sure that is why @mitsuhiko wrote it as an assignment rather than import in the first place.)


My change just moves that assignment from module to function scope, since we really only need to hold the reference in that one function.

Also note that, if this is converted to a function-scoped `import`, that would currently require a `# pylint: ignore=import-outside-toplevel`.  That, in my book, is a strong argument against doing that.

I'd rather leave it as is.  My second choice would be to put it back the way it was before I touched it.  Third choice would be a module-level `from queue import Empty as QueueEmpty` — but that would definitely require a comment with the `import` to warn as to why that is being done (rather than just referring to `queue.Empty`.)

dairiki: @yagebu @runfalk  Tell me if/how you want this adjusted.

(I was going to rebase to PR so that the CI lint test passes, but I'm holding off on that because I'm not sure whether these code comments will survive such a rebase.)

yagebu: My initial thought was to convert it to a module-level import, but you're right that having it in the function (as an assignment) is more explicit and "safer" from accidental changes in the future. So I'm also for leaving it as is now :)

lektor/watcher.py
```diff
@@ -2,59 +2,86 @@
...
+        self.event_handler = EventHandler()
+        self.paths = paths
+        self.observer_classes = observer_classes
+        self.observer = None
+
+    def start(self):
+        untried = set(self.observer_classes)
+        for observer_class in self.observer_classes:
+            try:
+                self._start_observer(observer_class)
```
yagebu, line 57: There should be a return below this line, the code currently starts all the observers in sequence

lektor/watcher.py
```diff
@@ -2,59 +2,86 @@
...
+            except Exception as exc:
+                untried.discard(observer_class)
+                if len(untried) == 0:
+                    raise
+                click.secho(
+                    f"Creation of {observer_class.__module__}.{observer_class.__name__} "
+                    f"failed with exception:\n  {exc.__class__.__name__}: {exc!s}\n"
+                    "This may be due to a configuration or other issue with "
+                    "your system.\nFalling back to polling for file modifications.",
```
yagebu: This message is only correct with the default observer classes. No big deal, but with the iterator approach outlined above, you could also put the name of the next class that we are about to try here.

lektor/watcher.py
```diff
@@ -2,59 +2,86 @@
...
-            self.observer.schedule(self.event_handler, path, recursive=True)
-        self.observer.setDaemon(True)
+    def __init__(self, paths, observer_classes=(Observer, PollingObserver)):
+        self.event_handler = EventHandler()
+        self.paths = paths
+        self.observer_classes = observer_classes
+        self.observer = None
+
+    def start(self):
+        untried = set(self.observer_classes)
```
yagebu: This set seems to be mainly used to check whether there's still an observer left in the except clause or whether we need to re-raise. The fact that the set removes duplicates also has some weird effects in that a failing observer class would be tried multiple times if given multiple times at the start or middle of the list of observer classes, but it would only be tried once if given multiple times at the end of that list.

Maybe with an iterator, this could be handled in the same loop (two other nice side effects: fails loudly for empty observer_classes and gives us access to the next class in the except handler). 

```suggestion
        observers = iter(self.observer_classes)
        observer_class = next(observers)
        while observer_class: [...]
```

and then a `observer_class = next(observers, None)` in the except.

yagebu (changes requested):

lektor/watcher.py
dairiki: Well, that's embarrassing.  Thank you.

dairiki: I've now added a check to prevent starting more than one observer.  (It also guards against someone calling `BasicWatcher.start()` more than once.)

lektor/watcher.py
dairiki: I've adjusted the message.  See if you like the latest version

lektor/watcher.py
dairiki: The real (though esoteric) case that this was concerned with was to prevent issuing a spurious "falling back to polling" message when creating the default `Observer` fails on systems where `Observer` is `PollingObserver`.

I've refactored the start-observer loop.  See if you like.

lektor/watcher.py
```diff
@@ -1,60 +1,103 @@
...
-            self.observer.schedule(self.event_handler, path, recursive=True)
-        self.observer.setDaemon(True)
+    def __init__(self, paths, observer_classes=(Observer, PollingObserver)):
+        self.event_handler = EventHandler()
+        self.paths = paths
+        self.observer_classes = observer_classes
+        self.observer = None
+
+    def start(self):
+        observer_classes = list(_unique_everseen(self.observer_classes))
```
yagebu, line 52: Could you put a comment here to explain why you need to remove duplicates? I'd expect a function that I pass a sequence of classes to try to try these classes in sequence - if I pass something in twice, I'd expect it to be tried twice.

Alternatively this could also be made explicit by conditionally defining the default observers to avoid a duplicate PollingObserver.

yagebu (changes requested):
Just one minor comment, otherwise this LGTM :)

lektor/watcher.py
```diff
@@ -1,60 +1,103 @@
...
 
     def __iter__(self):
-        if self.event_handler.queue is None:
-            raise RuntimeError("watcher used with callback")
+        # Alias this since we may need it during interpreter shutdown
+        queue_Empty = queue.Empty
+
         while 1:
             try:
                 item = self.event_handler.queue.get(timeout=1)
```
dairiki: Does anyone have an idea as to why the `timeout` is specified here?

Since timeouts (signaled by `queue.Empty`) are just ignored, it seems to me the timeout could be eliminated (so that the `queue.get` blocks forever).

yagebu: Sounds good.

yagebu (approved):

dairiki:
I think this is ready to merge.

If you'd like me to rebase and/or squash/cleanup the commit history (or anything else), let me know.

yagebu:
> If you'd like me to rebase and/or squash/cleanup the commit history (or anything else), let me know.

The commit history didn't look too bad, so I merged it as is :)

PR #887, opened 2021-02-20
lint: enable no-self-use

yagebu:
Add disables in classes that are expected to be subclassed and turn the
methods into static methods in the other places.

### DIFF
--- a/lektor/assets.py
+++ b/lektor/assets.py
@@ -74,6 +74,7 @@ class Asset(SourceObject):
         return iter(())
 
     def get_child(self, name, from_url=False):
+        # pylint: disable=no-self-use
         return None
 
     def resolve_url_path(self, url_path):
--- a/lektor/build_programs.py
+++ b/lektor/build_programs.py
@@ -138,6 +138,7 @@ class BuildProgram:
         building.  An individual build never recurses down to this, but
         a `build_all` will use this.
         """
+        # pylint: disable=no-self-use
         return iter(())
 
 
--- a/lektor/builder.py
+++ b/lektor/builder.py
@@ -343,7 +343,8 @@ class BuildState:
         )
         return cur.fetchone() is not None
 
-    def _get_artifact_config_hash(self, cur, artifact_name):
+    @staticmethod
+    def _get_artifact_config_hash(cur, artifact_name):
         """Returns the artifact's config hash."""
         cur.execute(
             """
@@ -1159,7 +1160,8 @@ class Builder:
                 return ctx
         return None
 
-    def update_source_info(self, prog, build_state):
+    @staticmethod
+    def update_source_info(prog, build_state):
         """Updates a single source info based on a program.  This is done
         automatically as part of a build.
         """
--- a/lektor/context.py
+++ b/lektor/context.py
@@ -138,7 +138,8 @@ class Context:
     def push(self):
         _ctx_stack.push(self)
 
-    def pop(self):
+    @staticmethod
+    def pop():
         _ctx_stack.pop()
 
     def __enter__(self):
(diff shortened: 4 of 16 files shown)


dwt:
Out of interest - what is the advantage to create so many static methods throughout the codebase? In my experience that makes the code base more complicated to work with without adding many benefits.

So, what have you found that this changes helps you with so much?

yagebu:
> Out of interest - what is the advantage to create so many static methods throughout the codebase? In my experience that makes the code base more complicated to work with without adding many benefits.

> So, what have you found that this changes helps you with so much?

I don't quite understand what you mean by complicated, I find they do quite the opposite since they make it more explicit whether a method is really a method or actually just a function/static method attached to the class. And in contrast to a language like Javascript, where static methods can't be called on class instances, I don't really see any disadvantages to static methods in Python.

As an example #882 would have been prevented by this rule.

dwt:
thanks

runfalk:
I agree with @yagebu here. Looks good!

dairiki:
And here comes the old grumpy guy...

I don't like this, for pretty much the same reasons I don't like black.

Python, in contrast to most brace-based languages (i.e. most other commonly-used languages) has a clean and *expressive* syntax.  One of the things that attracted me to python (many years ago) was that it read like pseudo-code.  When one starts legislating every bit of how one should format one's code, one loses that expressivity.

(As an aside, here is a [blog post on black](https://luminousmen.com/post/my-unpopular-opinion-about-black-code-formatter) which makes, I think, many good points. (I do not know the author of the post.))

As another aside, there are real downsides to adding commits to the project history which are solely about what is essentially code formatting.  Sometimes these are called for, but they should be avoided, as much as possible, especially for established projects.

In this case, specifically, how much do [these changes](https://github.com/lektor/lektor/pull/887/files) really improve readability?

Sometimes methods are static for structural or for logical reasons, sometimes they are logically just regular methods (that is: it is intended that they will only be invoked on an instance of the class) but they happen not to refer to `self`.  (By "structural reason", I mean that the program will fail if it isn't a static method, e.g. see #882.)

There is value in being able to mark the ones that have a good reason to be static with `@staticmethod`, and leaving the ones that just don't happen to reference `self` as regular methods.  What if tomorrow, or next year, someone comes along and wants to reference `self` from that method?  If `@staticmethod` means "there is a real reason for this to be a static method" rather than just "this method does not currently reference `self`", then it is a helpful reminder that our future self should think twice about what they are doing.

Most of the time, for the methods which weren't explicitly marked `@staticmethod`, why do we care whether they access `self` or not?  If we do care, we can look to see whether they do.  (Most of the methods touched by this PR are one-liners, making this particularly trivial.)

Some of this can be worked-around (as this PR does) by leaving some of the methods as is but adding `# pylint: disable=no-self-use`.  If one feels the need to add a bunch of exceptions to the rule, why make the rule in the first place?  My feeling is this just adds noise to the code base.  (Again, most of these are one-liner methods, where it is quite obvious that, in case it matters, they do not reference `self`.)

In summary, and specifically speaking to this case, I think there is value in having `@staticmethod` mean "there is a reason this is a static method", rather than solely meaning "this method does not reference `self`."

dairiki:
To expound on how #882 relates to this.

In that case, there is really no logical reason why `select_jinja_autoescape` should be a static method.  It would be perfectly reasonable to imagine that the determination of which template files should be autoescaped might depend on a setting in the site config file or some such (and thus depend on `self`.)  In the future, someone may indeed want to implement something like that.

There is a structural reason that it should be a static method — that is the program fails (in this case in a rather unimportant way) if `select_jinja_autoescape` is not a static function.

Prior to this PR, having the `@staticmethod` mark on it might cause anyone who is "enhancing" this method in the future to think twice as to why it was marked a static method.  After the PR, `@staticmethod` mostly means "this method does not currently reference `self`" (a fact which is easily determined, anyway, by just looking at the code) and loses its value as a signal that there might be a more significant reason for it to be a static method.

dwt:
@dairiki Thanks for making my case much more eloquently than I would have wanted to take the time for.

PR #896, opened 2021-02-28
Fix for unclosed file ResourceWarnings, plus more

dairiki:
I set out to fix the "ResourceWarning: unclosed file" which were being displayed on the console when using the rsync publisher under `lektor server`.  That led to a couple of rabbit holes.

### Issue(s) Resolved

This fixes `lektor.publisher.Command` so that (when used by `RsyncPublisher`) the stdin/stdout pipes from the rsync command get properly closed.  This fixes the warnings I set out to fix.

While fixing the unclosed files in `lektor.publisher.Command`, I discovered a possible deadlock in the code in `Command.__iter__` responsible for reading stdout and stderr pipes from the subprocess.  (If the subprocess writes a partial line, .readline() can block. This could possibly result in deadlock, with the reader blocked in readline() waiting for a "\n", and the subprocess waiting for the other pipe to drain before it gets to writing that "\n".

This PR adds a test and fix for that.

### Further questions

Is there really a good reason to keep separate stdout and stderr pipes from `Command`s subprocess?
Can we live with stdout and stderr both going to a single pipe (i.e. `stderr=subprocess.STDOUT`)?
If so, the code gets much simpler.

Are there other usages of `flask.json` that could be dropped in favor of plain old `json`?

### Description of Changes

- Close stdout and stderr pipes when `lektor.publisher.Command` completes
- Fix possible deadlock in `lektor.publisher.Command`


tests/test_publisher.py
```diff
@@ -0,0 +1,98 @@
...
+
+    # This is essentially how RsyncPublisher runs rsync.
+    with Command(["echo"], env={}) as client:
+        for _ in client:
+            pass
+    del client
+
+    for warning in recwarn.list:
+        print(warning)
+    assert len(recwarn) == 0
```
runfalk: `recwarn.list` is annotated to be a list. If we do `assert recwarn.list == []` I think we could skip the for loop.

tests/test_publisher.py
```diff
@@ -0,0 +1,98 @@
...
+        pipes = [os.pipe() for n in range(2)]
+        readers = Streams(open(r, "rb") for r, w in pipes)
+        writers = Streams(open(w, "wb", buffering=0) for r, w in pipes)
+        try:
+            yield readers, writers
+        finally:
+            writers.close()
+            readers.close()
+
+    def test(self, join_pipes, pipes):
```
runfalk: In general I prefer the tests to be flat. I think these could be `test_join_pipes_ordering` (if that is what it tests) and `test_join_pipes_prevents_partial_line_deadlock`. I know `pytest` supports both methods but I consider not grouping by class to be a bit more standard.

tests/test_publisher.py
```diff
@@ -0,0 +1,98 @@
...
+        for _ in client:
+            pass
+    del client
+
+    for warning in recwarn.list:
+        print(warning)
+    assert len(recwarn) == 0
+
+
+def iter_in_thread(it):
```
runfalk: It would be great with a comment why this function is needed.

tests/test_publisher.py
```diff
@@ -0,0 +1,98 @@
...
+from lektor.publisher import _join_pipes_using_threads
+from lektor.publisher import Command
+
+
+def test_Command_triggers_no_warnings(recwarn):
+    # This excercises the issue where publishing via rsync resulted
+    # in ResourceWarnings about unclosed streams.
+
+    # This is essentially how RsyncPublisher runs rsync.
+    with Command(["echo"], env={}) as client:
```
runfalk: I don't think passing `env={}`  does anything.

tests/test_publisher.py
```diff
@@ -0,0 +1,98 @@
...
+        for _ in client:
+            pass
+    del client
```
runfalk: Is the `del` call necessary here? I don't think Python gives any guarantees about when the destructor runs.

runfalk (commented):
Very nice and clean refactor of join_pipes. I added a few comments and questions.

runfalk:
I don't know about replacing `flask.json` with `json` in general. Flask's version seems to come preconfigured for serializing a few types that I think stdlib doesn't do (https://github.com/pallets/flask/blob/master/src/flask/json/__init__.py#L37). I think replacing it in general is better as its own PR.

As long as we don't risk breaking anything I don't see why we shouldn't replace it.

tests/test_publisher.py
dairiki: Probably true.  That's a remnant from when I was trying to elicit the unclosed file warnings within the test.  (See next comment.)  I will delete.

tests/test_publisher.py
dairiki: The unclosed stream ResourceWarnings don't get emitted until the streams are garbage collected.  That happens when the Popen is collected, which, in turn, happens with the Command is collected.
Without the `del`, that doesn't happen until the test is completed, in which case, the ResourceWarnings (should there be any) don't get caught in the recwarn.
I will add a comment.

tests/test_publisher.py
dairiki: The print loop is just to display any warnings on stdout to help with diagnostics should the test fail.  (The `assert len(recwarn) == 0` handles the actual check of the test result.)

I can delete the loop if you'd like.  It just that then, when the test fails you get no indication of what warnings were caught (just how many.)

dairiki:
To reiterate, I wonder if the join_pipes code is really necessary at all.

It appears that the only thing the captured output of the `Command`'s subprocess is to yield it, line by line, from the Publisher.publish method.  The point of this is to display the output to the user — either by streaming it through the web admin API, or just by printing on the console.

`Command` currently uses two pipes — one for stdout, one for stderr — from the child process.  It wants to be able to return any output,  line by line, on either stream.  That's where the join_pipes complication becomes necessary.

Since we always combine stderr and stdout before feeding the output back to the user, why not just use one pipe?  (Dup stdout to stderr using the `stderr=subprocess.STDOUT` parameter to `Popen`.)  Then we only need to read from the one pipe.  There are no deadlock issues, so no need for select or threads to do the reading.

The output might be a little different this way, in that there's the possibility that partial lines written to stdout can get mixed up with partial lines written to stderr (and vice versa), but I can't see how that's really an issue.  No output will get lost — just shuffled/scrambled a bit, in the worst case.  Most commands write either to stdout or stderr, not both, so, in most cases, it makes no difference at all.

tests/test_publisher.py
runfalk: That was my point. If you do `assert recwarn.list == []` instead of the length check pytest will print the extra elements in the array.

runfalk:
I haven't really taken a deep dive into this yet and don't know much about the original implementation, but it seems like you are indeed correct in your observation. I don't see why we couldn't get rid of `_join_lines`.

tests/test_publisher.py
dairiki: That only sort of works.  The repr of the WarningMessage objects is not very useful, while the stringification of them is.
```
>>> pytest tests/test_publisher.py -vv
============================= test session starts ==============================
platform linux -- Python 3.8.5, pytest-6.2.2, py-1.10.0, pluggy-0.13.1 -- /home/dairiki/git/web/lektor-packages/lektor/py38/bin/python
cachedir: .pytest_cache
rootdir: /home/dairiki/git/web/lektor-packages/lektor, configfile: pytest.ini
plugins: cov-2.11.1, mock-3.5.1, click-1.0.2
collected 1 item                                                               

tests/test_publisher.py::test_Command_triggers_no_warnings FAILED        [100%]

=================================== FAILURES ===================================
______________________ test_Command_triggers_no_warnings _______________________

recwarn = WarningsRecorder(record=True)

    def test_Command_triggers_no_warnings(recwarn):
        # This excercises the issue where publishing via rsync resulted
        # in ResourceWarnings about unclosed streams.
    
        # This is essentially how RsyncPublisher runs rsync.
        with Command(["echo"], env={}) as client:
            for _ in client:
                pass
        del client
    
>       assert recwarn.list == []
E       assert [<warnings.WarningMessage object at 0x7f9d3488c130>,\n <warnings.WarningMessage object at 0x7f9d3488c1c0>] == []
E         Left contains 2 more items, first extra item: <warnings.WarningMessage object at 0x7f9d3488c130>
E         Full diff:
E           [
E         -  ,
E         +  <warnings.WarningMessage object at 0x7f9d3488c130>,
E         +  <warnings.WarningMessage object at 0x7f9d3488c1c0>,
E           ]

tests/test_publisher.py:14: AssertionError
```

vs

```
>>> pytest tests/test_publisher.py -vv
============================= test session starts ==============================
platform linux -- Python 3.8.5, pytest-6.2.2, py-1.10.0, pluggy-0.13.1 -- /home/dairiki/git/web/lektor-packages/lektor/py38/bin/python
cachedir: .pytest_cache
rootdir: /home/dairiki/git/web/lektor-packages/lektor, configfile: pytest.ini
plugins: cov-2.11.1, mock-3.5.1, click-1.0.2
collected 1 item                                                               

tests/test_publisher.py::test_Command_triggers_no_warnings FAILED        [100%]

=================================== FAILURES ===================================
______________________ test_Command_triggers_no_warnings _______________________

recwarn = WarningsRecorder(record=True)

    def test_Command_triggers_no_warnings(recwarn):
        # This excercises the issue where publishing via rsync resulted
        # in ResourceWarnings about unclosed streams.
    
        # This is essentially how RsyncPublisher runs rsync.
        with Command(["echo"], env={}) as client:
            for _ in client:
                pass
        del client
    
        for warning in recwarn.list:
            print(warning)
>       assert len(recwarn) == 0
E       assert 2 == 0
E         +2
E         -0

tests/test_publisher.py:16: AssertionError
----------------------------- Captured stdout call -----------------------------
{message : ResourceWarning('unclosed file <_io.BufferedReader name=11>'), category : 'ResourceWarning', filename : '/home/dairiki/git/web/lektor-packages/lektor/tests/test_publisher.py', lineno : 12, line : None}
{message : ResourceWarning('unclosed file <_io.BufferedReader name=13>'), category : 'ResourceWarning', filename : '/home/dairiki/git/web/lektor-packages/lektor/tests/test_publisher.py', lineno : 12, line : None}
```

dairiki:
All right, I think I've addressed most or all of the comments.  (Mostly about adding comments/docstrings to the tests.)

Then I deleted all the _join_pipes stuff (which deleted most of the new comments and docstrings.)  (But that's all still in the last commit, so easily undone.)

dairiki:
I do have one question — I'm sure someone else knows about this better than I.

Currently, the command output is always decoded using UTF-8.  Would it make more sense to use the system preferred encoding?  Or is that a whole other can of worms?  (E.g. problems when the preferred encoding is set to something stupid like ASCII.  See <https://click.palletsprojects.com/en/7.x/python3/#python-3-surrogate-handling>.)

tests/test_publisher.py
runfalk: Ok, leave it then.

tests/test_publisher.py
```diff
@@ -9,11 +11,9 @@ def test_Command_triggers_no_warnings(recwarn):
     with Command(["echo"]) as client:
         for _ in client:
             pass
-    # Delete our reference so that the Command instance gets garbage
-    # collected here. Otherwise, gc will not happen until after the
-    # test completes and warnings emitted during gc will not be captured
-    # by the recwarn fixture.
-    del client
+    # Check for warnings during garbage collection, too
+    client = weakref.ref(client)
```
runfalk: Maybe you want to explicitly run garbage collection here? https://docs.python.org/3/library/gc.html?highlight=gc#gc.collect

dairiki: I don't think it is necessary.  Objects are finalized as soon as their reference count gets to zero.   Explicitly triggering garbage collection should only be necessary when there are reference cycles involved.  As `gc.collect()` performs a full mark-and-sweep collection of all live objects (I think), it can take a bit of time to run. If there's no reason to do it, I think it's best to avoid.

I just added the weakref sanity check just to make sure that the instance is actually being finalized when we think it is.

runfalk: Yes, this is true for CPython (but I don't think it's true for some other implementations like PyPy). It'd make the test a little bit more explicit about what we're testing and it would be future proof if we decide to officially support other interpreters. It's nothing major though.

dairiki: You're right — [PyPy does not](https://doc.pypy.org/en/latest/cpython_differences.html#differences-related-to-garbage-collection-strategies) rely on reference counting to trigger finalization.

I've updated the test again.  See if you like it.

runfalk: Looks good!

dairiki:
If we're happy switching to the single-pipe solution, let me know — I'll squash commits and clean-up so that this can be merged.
(If we'd rather keep the separate stdin/stdout pipes, let me know, too, and I'll revert to the join_pipes solution.)

dairiki:
> Currently, the command output is always decoded using UTF-8. Would it make more sense to use the system preferred encoding? 

It seems that starting with python 3.7 it is relatively safe to use the system configured default encoding.  Prior to that, if the system was configured (most likely, inadvertently) to use ASCII, this was likely to cause trouble.  Py37+ will now force the use of UTF-8 in such cases.  See [PEP 538](https://www.python.org/dev/peps/pep-0538/).

Previously, lektor always forced the use of UTF-8 for reading output from subcommands.  I've just pushed a commit that changes this to use the system default encoding for py37+.

yagebu (approved):
This looks great, thanks :)

PR #897, opened 2021-03-01
Fix two i18n buglets/typos

### LINKED ISSUE #895: bug in _build_parent_path function
https://github.com/lektor/lektor/blob/5a2895c1b683c1df98beed2ad970f8ebdab26a75/lektor/sourcesearch.py#L61

In this function, the info object can be None, but `info['path']` is used before checking ! This happened in one of my plugins. 

```
def _build_parent_path(path, mapping, alt, lang):
    rv = []
    for parent in _iter_parents(path):
        info = _find_best_info(mapping.get(parent) or [], alt, lang)
        id = _id_from_path(info['path'])
        if info is None:
            title = id or '(Index)'
        else:
            title = info.get('title')
        rv.append({
            'id': id,
            'path': parent,
            'title': title
        })
    return rv
```

See https://stackoverflow.com/questions/66389221/a-lektor-website-unable-to-build-twice-in-a-row for my own case, but it should'nt really matter as the function seems buggy whatever my plugin :-)

Here is a fix we could use ?

```
def _build_parent_path(path, mapping, alt, lang):
    rv = []
    for parent in _iter_parents(path):
        info = _find_best_info(mapping.get(parent) or [], alt, lang)
        if info is None:
            id=''
            title = '(Index)'
        else:
            id = _id_from_path(info['path'])
            title = info.get('title')
        rv.append({
            'id': id,
            'path': parent,
            'title': title
        })
    return rv
```
 
What do you guys think ?

dairiki:
In looking into #895, two apparent typos were discovered.  Both of these result in edge-case bugs.  This fixes them.

### Issue(s) Resolved

Fixes #895

It's clear from inspection that there's something not quite right about the code around [line 61 & 62][so61] of `sourceobj.py` — the `if` branch is not reachable, since if `info is None`, the exception reported in #895 will occur before the if statement is executed.
(After further digging, I believe that it *should* never come to be that `info` is `None`. That is was happening was due to another buglet. More on that later.)

Note `mapping` is a dict which maps from `path` to a list of `info`s.  Inspection of [`_mapping_from_cursor`][mfc] makes it clear that all the `info`s in `mapping.get(parent)` (if any) will have `info["path"] == parent`.  The fix here is to use `parent` in place of `info["path"]`.

[so61]: <https://github.com/lektor/lektor/blob/5a2895c1b683c1df98beed2ad970f8ebdab26a75/lektor/sourcesearch.py#L61>
[mfc]: <https://github.com/lektor/lektor/blob/5a2895c1b683c1df98beed2ad970f8ebdab26a75/lektor/sourcesearch.py#L28>

### Related Issues / Links

There's another typo in [`Record.get_record_label_i18n`][grl] in `db.py`.  The comment says "Fill in english if missing", but the code current does "fill in english if there are no other labels in any language".  In this case, the comment is right, the code is wrong.

Other [comments][] and [code][] spell out that `"en"` is the hard-coded fallback for internationalizations. There should always be an `en` entry in the i18n mappings.

[grl]: <https://github.com/lektor/lektor/blob/5a2895c1b683c1df98beed2ad970f8ebdab26a75/lektor/db.py#L413>
[comments]: <https://github.com/lektor/lektor/blob/5a2895c1b683c1df98beed2ad970f8ebdab26a75/lektor/i18n.py#L61>
[code]: <https://github.com/lektor/lektor/blob/5a2895c1b683c1df98beed2ad970f8ebdab26a75/lektor/admin/static/js/i18n.tsx#L46>

You can exercise this bug as follows:
1. Create a new quickstart Lektor project
2. Edit `models/page.ini` and change it so that it has a localized label, but no default label defined.  E.g.
```ini
[model]
name = Page
label[de] = Auf Deutsch: {{ this.title }}
[fields.title]
...
```
3. Run the admin GUI, setting the language to something that is neither English nor the one the label is localized for.
```sh
lektor --language fr server
```
4. Note that the page titles are screwy/blank in various places.
Screenshot from 2021-03-01 09-07-26

### DIFF
--- a/lektor/db.py
+++ b/lektor/db.py
@@ -410,7 +410,7 @@ class Record(SourceObject):
                 label = self.get_fallback_record_label(lang)
             rv[lang] = label
         # Fill in english if missing
-        if not rv:
+        if "en" not in rv:
             rv["en"] = self.get_fallback_record_label("en")
         return rv
 
--- a/lektor/sourcesearch.py
+++ b/lektor/sourcesearch.py
@@ -58,7 +58,7 @@ def _build_parent_path(path, mapping, alt, lang):
     rv = []
     for parent in _iter_parents(path):
         info = _find_best_info(mapping.get(parent) or [], alt, lang)
-        id = _id_from_path(info["path"])
+        id = _id_from_path(parent)
         if info is None:
             title = id or "(Index)"
         else:


dairiki:
I suspect that the CI test failure is a transient github issue(?) There is an extra, duplicate Mac py3.6 job that failed with "_GitHub Actions has encountered an internal error when running your job_" after about an hour — and there's no log output.

Perhaps someone with sufficient permissions could try re-running the workflow.

lektor/db.py
```diff
@@ -410,7 +410,7 @@ def get_record_label_i18n(self):
                 label = self.get_fallback_record_label(lang)
             rv[lang] = label
         # Fill in english if missing
-        if not rv:
+        if "en" not in rv:
```
runfalk, line 413: Wouldn't this add an `en` language, even for sites that don't have an `en` translation?

runfalk:
I've retriggered the jobs. GitHub actions had an outage today, maybe it was that (https://www.githubstatus.com/incidents/xn0sd2x4nd7f).

yagebu (approved):
Thanks :)

PR #898, opened 2021-03-01
Use python's json.load instead of flask.json.load

dairiki:
### Issue(s) Resolved

`Flask.json.load` has a buglet which causes python to issue "unclosed file" ResourceWarnings.  The first commit in this PR has a test which demonstrates this.

It appears that the only reason `flask.json.load` was used in `lektor.i18n` was for its ability to read from binary files.  Python's `json.load` supports reading from binary files since python 3.6.
(Note that there are two other places where Lektor uses `json.load` — in `buildfailures.py` and `databags.py`.  In both of those cases it is already using Python's `json.load`.)

This PR replaces the use of `flask.json.load` with `json.load`.  This eliminates the unclosed file warnings.

### Still to be investigated

Can/should Lektor's other uses of `flask.json` be replaced by `json`?

### DIFF
--- a/lektor/i18n.py
+++ b/lektor/i18n.py
@@ -1,6 +1,5 @@
+import json
 import os
-
-from flask import json
 
 
 translations_path = os.path.join(
@@ -13,8 +12,6 @@ KNOWN_LANGUAGES = list(
 
 translations = {}
 for _lang in KNOWN_LANGUAGES:
-    # We use flask.json here which can deal with bytes unlike the stdlib
-    # json module which barfs on bytes on 3.x
     with open(os.path.join(translations_path, _lang + ".json"), "rb") as f:
         translations[_lang] = json.load(f)
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -1,8 +1,8 @@
+import json
 import os
 from operator import itemgetter
 
 import pytest
-from flask import json
 
 from lektor.admin import WebAdmin
 from lektor.admin.utils import eventstream
--- a/tests/test_i18n.py
+++ b/tests/test_i18n.py
@@ -1,0 +1,10 @@
+import importlib
+
+import lektor.i18n
+
+
+def test_loading_i18n_triggers_no_warnings(recwarn):
+    importlib.reload(lektor.i18n)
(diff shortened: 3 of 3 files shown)


dairiki:
## Remaining usages of `flask.json`

### lektor/admin/utils.py

`Flask.json.dumps` is used in the `lektor.admin.utils.eventstream` decorator.  Currently, the `eventstream` decorator is only used to stream the results of `Publisher.publish` (a generator that yields output lines from the publishing process as individual `str`s) back to the admin GUI when publishing the site.

Since currently this usage of `dumps` is only being called upon to serialize `str`s (and `None`), none of the [extra magic][1] of  `flask.json.JSONEncoder` is being used.  This usage could be replaced by stock `json.dumps`.

On the other hand, who knows what the `eventstream` decorator may be used for in the future?  And we are using this within a `flask` app.  My feeling is to leave this as it is.

[1]: <https://github.com/pallets/flask/blob/82d69cd06c7ff738774c5ab7ef48bc6251f5c014/src/flask/json/__init__.py#L43>

### lektor/admin/modules/api.py

`Flask.json.jsonify` is used here to serialize JSON responses from API methods.  `Jsonify` has lots of flask-magic (primarily, it wraps the result in a response object), so this should be left as is.

### tests/test_api.py

`Flask.json.loads` is being used to parse the output of the `/admin/api/recordinfo` admin API method.  `Flask.json.loads` has extra magic to support deserializing `bytes`, as well as to support configuring the use of custom `JSONDecoder`s within a particular flask app or request, however, none of that magic is being used here.  This can be replaced by stock Python `json.loads`.

dairiki:
As noted in my previous comment, I suggest we leave the usage of `flask.json.dumps` (and `flask.json.jsonify`) within the admin flask app as they are.

All other usages of `flask.json` have now (I think) been converted to using plain old `json`.

yagebu (approved):
LGTM. Agree with mostly moving to stdlib json outside of the flask app but sticking with flask.json in a Flask context.

If we ever run into problems with flask.json in the Flask app, adding a dependency on `simplejson`, which is used instead of the stdlib one by Flask if available, might solve some issues.

PR #904, opened 2021-03-20
Avoid editable installs

yagebu:
This would fix (among others) #865 and is an alternative to PR #877.

Some problems due to our use of editable installs:
- Packages with extra_requires fail to install (#865) - this might be fixed at some point in pip
- Due to the same pip bug, these editable installs currently end up polluting the global Python `sys.path`
- Packages that use something other than setuptools won't install - this probably won't ever be fixed, PEP517 has no notion of editable installs.

The main advantage of editable installs currently is that one gets instant reloads on code changes when running lektor server in dev mode. However, since we only do editable once, updates to the dependencies or entry points (really anything but the code) of a local package would never get picked up. So we'd really have to do reinstalls on changes to the package.

What this PR does instead of doing only an (editable) install once, is to compute a checksum of local packages and to write that checksum to the manifest on install. That way, we are sure to do a reinstall on any package change but not more often than that.

Once #886 is merged, I'll add the packages path to the watcher to ensure that changes to the packages are picked up and reinstalls can be triggered in dev mode.

### DIFF
--- a/lektor/packages.py
+++ b/lektor/packages.py
@@ -3,7 +3,6 @@ import os
 import shutil
 import site
 import sys
-import tempfile
 from subprocess import PIPE
 
 import click
@@ -75,9 +74,7 @@ def remove_package_from_project(project, name):
     return None
 
 
-def download_and_install_package(
-    package_root, package=None, version=None, requirements_file=None
-):
+def download_and_install_package(package_root, package=None, version=None):
     """This downloads and installs a specific version of a package."""
     # XXX: windows
     env = dict(os.environ)
@@ -93,8 +90,6 @@ def download_and_install_package(
 
     if package is not None:
         args.append("%s%s%s" % (package, version and "==" or "", version or ""))
-    if requirements_file is not None:
-        args.extend(("-r", requirements_file))
 
     rv = portable_popen(args, env=env).wait()
     if rv != 0:
@@ -114,42 +109,14 @@ def install_local_package(package_root, path):
             "-m",
             "pip",
             "install",
-            "--editable",
             path,
-            "--install-option=--install-dir=%s" % package_root,
-            "--no-deps",
+            "--target",
+            package_root,
         ],
         env=env,
     ).wait()
     if rv != 0:
         raise RuntimeError("Failed to install local package")
-
-    # Step 2: generate the egg info into a temp folder to find the
-    # requirements.
-    tmp = tempfile.mkdtemp()
-    try:
-        rv = portable_popen(
-            [
-                sys.executable,
-                "setup.py",
-                "--quiet",
-                "egg_info",
-                "--quiet",
-                "--egg-base",
-                tmp,
-            ],
-            cwd=path,
-        ).wait()
-        dirs = os.listdir(tmp)
-        if rv != 0 or len(dirs) != 1:
-            raise RuntimeError("Failed to create egg info for local package.")
-        requires = os.path.join(tmp, dirs[0], "requires.txt")
-
-        # We have dependencies, install them!
-        if os.path.isfile(requires):
-            download_and_install_package(package_root, requirements_file=requires)
-    finally:
-        shutil.rmtree(tmp)
 
 
 def get_package_info(path):
--- a/lektor/packages.py
+++ b/lektor/packages.py
@@ -182,9 +182,9 @@ def write_manifest(filename, packages):
     with open(filename, "w", encoding="utf-8") as f:
         for package, version in sorted(packages.items()):
             if package[:1] == "@":
-                f.write("%s\n" % package)
+                f.write(f"{package}\n")
             else:
-                f.write("%s=%s\n" % (package, version))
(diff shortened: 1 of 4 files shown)


lektor/packages.py
```diff
@@ -241,6 +178,12 @@ def list_local_packages(path):
     return rv
 
 
+def hash_directory(path) -> str:
+    """Compute a checksum for a directory."""
+    file_info = FileInfo(None, path)
```
dairiki: FileInfo.checksum may not be exactly what is wanted here.  (FileInfo appears to only compute a checksum over the files in the top-level of a directory.  It does not recurse into subdirectories.)

dairiki:
Here are some random, not completely formed, thoughts.

First, note that there is progress in modernizing support for editable installs.  [PEP 660](https://www.python.org/dev/peps/pep-0660/) [has been accepted](https://discuss.python.org/t/pronouncement-on-peps-660-and-662-editable-installs/9450).

An alternative view of what is causing the pain here is that it is not the use of editable installs that is causing issues.  (These have and continue to be well supported, e.g. by pip running in a normal virtualenv.)  What is causing the trouble is Lektor's custom management (achieved through the use of `pip install`s `--target` option) of a separate python library directory.  To me the operation of the `--target` option is not really fully specified, and seems to be the cause of all these problems.  (Pip has an issue tag [dedicated to the topic](https://github.com/pypa/pip/issues?q=is%3Aissue+is%3Aopen+%22--target%22+label%3A%22C%3A+target%22).)

Were we starting from scratch, it seems the natural way to go would be to encourage one virtualenv per Lektor project.  That virtualenv would include the Lektor installation as well as plugins.  (All of this custom package management is required to support running a shared Lektor installation with various, not necessarily mutually compatible, sets of installed plugins.)

I don't have the solution, but it would sure be nice to get Lektor out of the package-management business altogether.

yagebu:
> I don't have the solution, but it would sure be nice to get Lektor out of the package-management business altogether.

I couldn't agree more.

This is PR is just intended as a makeshift solution to fix some of the existing bugs.

dairiki:
I'm just not sure the things addressed here a really "bugs" (rather than new features).  As I understand it, packages in the `packages` subdirectory are really intended for project-local code, with the package in the structure created by `lektor dev new-plugin`.  I don't think it was ever intended that one should be able to drop any distribution source tree into `packages` and have it work.

As such, supporting non-setuptools based distributions in `packages` is surely out of scope.  At least, not without some roadmap as to how Lektor will handle packaging/installation of plugins in the future.

Supporting `extra_requires` also seems to me like a feature addition rather than a bug fix.  I don't really have anything against it, but this seems like a pretty elaborate work-around.  (Editable installs, are really the canonical way to install packages in a way that is responsive to edits.)

Spit-balling follows:

Would it be horrible to deprecate Lektor's plugin-installation features?  At it's simplest, this would move away with automatic installation of stuff in the `packages` directory and the `[packages]` section of the project file, encouraging users instead to install plugins in whatever virtualenv is hosting their project's Lektor installation.  Perhaps we could come up with a little sugar to help with that process — I haven't thought it through very completely.

A way to disable/enable the use of specific plugins on a per-project basis (i.e. via the project file) would probably need to be added back.  Thus different projects running off the same virtualenv could still use different sets of plugins.

yagebu:
I'm with you that the `--target` option is the bigger culprit, but of the two problematic option it is the one we can't drop without creating an even bigger mess (installing packages into users current environment without their knowledge) or basically taking on re-implementing the complexity of pipenv or poetry to build a package manager into Lektor that can deal with the horrible mess that is Python packaging.

>Would it be horrible to deprecate Lektor's plugin-installation features? At it's simplest, this would move away with automatic installation of stuff in the packages directory and the [packages] section of the project file, encouraging users instead to install plugins in whatever virtualenv is hosting their project's Lektor installation. Perhaps we could come up with a little sugar to help with that process — I haven't thought it through very completely.

Yes, please. The main reason for this PR is that I just had too much of my time wasted because some broken local package in lektor-website or the like broke my development Python environment for all Lektor projects. I just want to fix that horrible bug, then deprecate this whole mess and rip it out in a couple of years.

I'd suggest we recommend pipenv instead for people who want that "little sugar". I've recently moved all my (production) Lektor projects to pipenv and that has worked quite well. But that's really a topic for a new issue...

> A way to disable/enable the use of specific plugins on a per-project basis (i.e. via the project file) would probably need to be added back. Thus different projects running off the same virtualenv could still use different sets of plugins.

This doesn't even work currently due to the editable installs ending up as system-wide or virtualenv-wide installs. In that case we should just recommend per-project virtualenvs or "pipenvs"/Pipfiles.

frostming:
I am against introducing Pipenv given how expensive it would be to lock dependencies.

I recommend we use a real virtualenv for the package caches and install plugins into it. In this way, we can get rid of the hack of editable install.

yagebu:
#1065 will be a better fix for the underlying problems

Issue #909, opened 2021-05-24
Unable to install d/t Werkzeug / Flask / Lektor Dependency issue (temp fixed by rolling back flask.)


ViolanteCodes:
I tried installing Lektor using the step by step guide on the homepage, both using the cURL method and the pip method and was unable to run either Lektor quickstart or Lektor server. I received the following error during installation

ERROR: flask 2.0.1 has requirement Werkzeug>=2.0, but you'll have werkzeug 0.16.1 which is incompatible.
Installing collected packages: Werkzeug, Lektor
  Attempting uninstall: Werkzeug
    Found existing installation: Werkzeug 2.0.1
    Uninstalling Werkzeug-2.0.1:
      Successfully uninstalled Werkzeug-2.0.1
Successfully installed Lektor-3.2.0 Werkzeug-0.16.1

When I tried to run lektor quickstart or lektor server, I got this error:

ImportError: cannot import name 'ContextVar' from 'werkzeug.local'

I tried pip upgrading werkzeug, which did nothing. I deleted the venv and started over; same issue. I then checked the flask update history and noticed there was a recent major update in early may, so i tried uninstalling flask and then pip install Flask==1.1.2.

At that point, I was successfully able to run Lektor quickstart. I just wanted to wanted to point this out as it looks like the Flask/Werkzeug changes were very recent.

bbbart:
same here. Steps I just used to get to a working installation were:
 - pipenv install "flask==1.1.4"
 - pipenv install lektor

PR #911, opened 2021-05-30
Port to Werkzeug 2.x

### LINKED ISSUE #910: Port to Werkzeug 2.x
Werkzeug 2.x should be mostly compatible, so it would be nice to have it ported.

ismaell:
Make Lektor compatible with Werkzeug 2.x.

### Issue(s) Resolved

* Closes: https://github.com/lektor/lektor/issues/910

### Related Issues / Links

* https://github.com/lektor/lektor/issues/909

### Description of Changes

* Replaces `werkzeug.posixemulation.rename` with `os.replace`
* Changes dep to `Werkzeug<3`

### DIFF
--- a/lektor/builder.py
+++ b/lektor/builder.py
@@ -11,7 +11,6 @@ from contextlib import contextmanager
 from itertools import chain
 
 import click
-from werkzeug.posixemulation import rename
 
 from lektor.build_programs import builtin_build_programs
 from lektor.buildfailures import FailureController
@@ -929,7 +928,7 @@ class Artifact:
                 op(con)
 
             if self._new_artifact_file is not None:
-                rename(self._new_artifact_file, self.dst_filename)
+                os.replace(self._new_artifact_file, self.dst_filename)
                 self._new_artifact_file = None
 
             if con is not None:
--- a/lektor/utils.py
+++ b/lektor/utils.py
@@ -25,7 +25,6 @@ from markupsafe import Markup
 from slugify import slugify as _slugify
 from werkzeug import urls
 from werkzeug.http import http_date
-from werkzeug.posixemulation import rename
 from werkzeug.urls import url_parse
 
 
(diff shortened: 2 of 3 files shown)


lektor/builder.py
```diff
@@ -11,7 +11,10 @@
 from itertools import chain
 
 import click
-from werkzeug.posixemulation import rename
+try:
+    from os import replace as rename
+except ImportError:
```
dairiki: Given that Lektor no longer supported python < 3.6, `os.replace` should always be available. (I think?)

ismaell: Oh, I didn't realize that, then I'll remove the fallback.

SrijanSriv: Hi! I believe this will install Werkzeug 2.x instead of 0.16. Is there a way to do it on my system manually? I'm running Ubuntu 20.04 by the way.
Thanks!

SrijanSriv:
Hi! I believe this will install Werkzeug 2.x instead of 0.16. Is there a way to do it on my system manually? I'm running Ubuntu 20.04 by the way.
It gets installed through `curl -sf https://www.getlektor.com/installer.py | python3` and immediately gives the error of how Werkzeug is not updated enough for Flask(I'm sorry I don't remember its version) and I can't do anything with lektor from there.
Thanks!

ismaell:
@SrijanSriv you would need to download the source code, apply the patch from this PR, then install from there...

SrijanSriv:
Thanks a ton! I have had issues about this for a while now. One more thing, the lektor from this development version runs only through venv right? I have the faulty lektor installed from the above mentioned link(which I can't delete for some reason?) and I don't want it to interfere with a sudden error.
Is there a way to delete the faulty lektor?

ismaell:
> lektor from this development version runs only through venv right?

You can also install it "normally" using `setup.py`...

> I have the faulty lektor installed from the above mentioned link(which I can't delete for some reason?) and I don't want it to interfere with a sudden error.
> Is there a way to delete the faulty lektor?

I think it uses `pip`, so:
```
pip uninstall lektor
```

andoresuperesu:
- Also Closes: #914

yagebu (approved):
Thanks! :)

PR #919, opened 2021-08-10
Cache directory can be set using the "LEKTOR_CACHE" environment variable

paternal:
Corresponding documentation pull request: https://github.com/lektor/lektor-website/pull/323

### Description of Changes

If environment variable `LEKTOR_CACHE` is set, use it as the cache directory. Example:

~~~sh
# Uses default ~/.cache/lektor directory
$ lektor build
…

# Uses custom ~/foo/bar/cache/lektor to store cache
$ export LEKTOR_CACHE="~/foo/bar/cache/lektor"
$ lektor build
…
~~~

### What problem does it solve?

The documentation explains [how to keep cache between build with Travis-CI](https://www.getlektor.com/docs/deployment/travisci/). But this is not possible on Gitlab, since one cannot refer to the home directory in the `.gitlab-ci.yaml` ([example with nodejs](https://docs.gitlab.com/ee/ci/caching/#cache-nodejs-dependencies)). With this feature, this is now possible (see page `glpages` in the corresponding pull request for the documentation: https://github.com/lektor/lektor-website/pull/323).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -10,6 +10,7 @@ UNRELEASED - in development
 - Fix the toggling of flow widgets in the admin UI to not mark the content as changed.
 - Fix the handling of the URLs when opening the admin UI from the pencil button.
 - Allow rsync deployment to a local path.
+- Allow cache directory to be set using the environment variable LEKTOR_CACHE.
 - Relax the checking of URL fields and allow FTP and mailto URLs.
 - Correctly track the page URL on the admin preview page.
 - Rename a CSS class in the admin UI so that it is not blocked by the EasyList FR adblock list.
--- a/lektor/utils.py
+++ b/lektor/utils.py
@@ -694,6 +694,9 @@ def get_app_dir():
 
 
 def get_cache_dir():
+    if "LEKTOR_CACHE" in os.environ:
+        return os.environ.get("LEKTOR_CACHE")
+
     if is_windows:
         folder = os.environ.get("LOCALAPPDATA")
         if folder is None:


dairiki:
Note that there are [already environment variables][1] one can use to influence the placement of the cache directory.

E.g. on Linux, one can set [`XDG_CACHE_HOME`][xdg] to specify an alternate location for everything that, by default, goes in `~/.cache`.  (Note that XDG_CACHE_HOME affects the placement of caches for all XDG-compliant applications, e.g. it affects where `pip` will put its cache.)

Specifically, XDG_CACHE_HOME will solve your problem with caching on Gitlab's CI/CD system.
 
[1]: <https://github.com/lektor/lektor/blob/f273c6918b5f1043a6b09980de0a82f758033929/lektor/utils.py#L697>
[xdg]: <https://wiki.archlinux.org/title/XDG_Base_Directory>

paternal:
> Note that there are [already environment variables](https://github.com/lektor/lektor/blob/f273c6918b5f1043a6b09980de0a82f758033929/lektor/utils.py#L697) one can use to influence the placement of the cache directory.

Why didn't I notice that? It is only a few lines below the lines I added… It looks like my PR is useless, then. Sorry for the noise

paternal:
I have created a new pull request for the documentation, to document the XDG_CACHE_HOME variable.
https://github.com/lektor/lektor-website/pull/324

yagebu:
I'll close this PR, since XDG_CACHE_HOME probably works well enough already as detailed above.

PR #920, opened 2021-08-13
Don't build universal wheel

dairiki:
Since we no longer support Python 2, building a universal wheel is unnecessary and confusing.


yagebu (approved):
Thanks :)

PR #921, opened 2021-08-13
Address DeprecationWarnings from Jinja2

dairiki:
The `autoescape` and `with` extensions have been enabled by default since Jinja2 2.9.  Since we currently require Jinja2>=2.11, there is no need to specify these explicitly.  (Doing so now elicits a deprecation warning.)

As of Jinja2>=3.0, `contextfilter` has been deprecated in favor of the new `pass_context` decorator.

It would be possible to check for `pass_context` and fall back to using `contextfilter` if it is not available, however this PR currently just updates our version pins to require Jinja2>=3.  (If this is a problem, let me know.)

### DIFF
--- a/lektor/environment/__init__.py
+++ b/lektor/environment/__init__.py
@@ -120,7 +120,7 @@ class Environment:
 
         self.jinja_env = CustomJinjaEnvironment(
             autoescape=self.select_jinja_autoescape,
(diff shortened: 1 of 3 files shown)


yagebu (approved):
Thanks!

Jinja2 v3 is quite new but since it makes our life easier here I think that the version bump is fine :)

PR #922, opened 2021-08-13
Use shutil.get_terminal_size instead of click.get_terminal_size`

dairiki:
`Shutil.get_terminal_size` has been supported since Python 3.3. Since we no longer support Python 2, there's no reason not to use it directly.

`Click.get_terminal_size` has been deprecated (and elicits deprecation warnings) since `click==8.0.0`


yagebu (approved):
Thank you :)

PR #924, opened 2021-08-14
Pin pytest-click<1

dairiki:
This is a PR for the 3.2-maintenance branch.

I'm not sure whether anyone still cares about the 3.2 branch, but just in case...

Pytest-click>=1.0 does not support python 2.7, so this new pin is needed.


yagebu (approved):
Thanks for paying attention to this branch :)

PR #927, opened 2021-08-22
CI updates and simplifications

yagebu:


dairiki:
Generally, +1 to the unification of the test configs between Linux, Windows, MacOS.

I notice that we're not testing under py3.8 at all.  Is that deliberate?

Also, the "smoke tests" seem an unnecessary complication to me.  Does their short-circuiting really provide a useful benefit?

yagebu:
> I notice that we're not testing under py3.8 at all. Is that deliberate?

I don't think it's deliberate. IMHO, testing on 3.6 and 3.9 would be enough and testing all intermediate versions is just a waste of resources.

> Also, the "smoke tests" seem an unnecessary complication to me. Does their short-circuiting really provide a useful benefit?

Some of the other maintainers were in favour of them. I also find them unnecessary and I'd be happy to remove them.

dairiki:
I do think it's worth testing on all supported python versions.  There are significant changes in the standard library API (in terms of supported functions and classes, arguments, deprecations, etc.) between versions.

It seems to be pretty standard practice that CI tests get run under each python version explicitly mentioned in the distribution'sv trove classifiers.

And the Linux python tests are relatively speedy (at least, compared to the Windows and MacOS tests.) Adding one or two more isn't really going to affect the overall test running time significantly. (I don't think that we need to test on all python versions on all OSes.)

PR #931, opened 2021-09-02
Make dist and unify CI with local tests

yagebu:
This adds a `dist`  make target which will build both a source distribution and a wheel to automate most of the distribution-creating part of the release process.

I'll also try to unify local tests (with tox) and the tests that are run in CI (to now also use tox), so that reproducing CI failures should be more reproducible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,14 @@ lint:
 .PHONY: test
 test: lint test-python test-js
 
+.PHONY: test-all
+# Run tests on all supported Python versions.
+test-all: lint test-js
+	tox -e py36
+	tox -e py37
+	tox -e py38
+	tox -e py39
+
 # This creates source distribution and a wheel.
 dist: build-js setup.cfg setup.py MANIFEST.in
 	python setup.py sdist bdist_wheel
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ test-python:
 # Run tests on the Frontend code.
 test-js: lektor/admin/node_modules
 	@echo "---> running javascript tests"
+	@cd lektor/admin; npx tsc
 	@cd lektor/admin; npm test
 
 .PHONY: lint
@@ -31,11 +32,9 @@ test: lint test-python test-js
 
 .PHONY: test-all
 # Run tests on all supported Python versions.
-test-all: lint test-js
-	tox -e py36
-	tox -e py37
-	tox -e py38
-	tox -e py39
+test-all: test-js
+	pre-commit run -a
+	tox
 
 # This creates source distribution and a wheel.
 dist: build-js setup.cfg setup.py MANIFEST.in
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 all: build-js
 
+.PHONY: build-js
 build-js: lektor/admin/node_modules
 	@echo "---> building static files"
 	@cd lektor/admin; npm run webpack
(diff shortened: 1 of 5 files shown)


dairiki:
Note that I did some very similar work on refactoring the CI tests awhile ago.  See my [refactor.ci-testing](https://github.com/dairiki/lektor/tree/refactor.ci-testing) branch.  The test speedups weren't quite as blatant as I was hoping for, so I let it drop...   Here's the [diff](https://github.com/lektor/lektor/compare/32c9ed51f01a22f67c808a8c033df780fc42857d...dairiki:refactor.ci-testing) showing changes made in that branch.

Note that I came up with a more complicated [tox.ini](https://github.com/dairiki/lektor/blob/refactor.ci-testing/tox.ini).  The biggest difference is that it will locally test under all supported python versions.  (Personally, I think this is a good thing.)

If you'd like to more easily compare, I can update my work (rebase to current master) and open a PR.  Let me know.

yagebu:
> Note that I came up with a more complicated tox.ini. The biggest difference is that it will locally test under all supported python versions. (Personally, I think this is a good thing.)

I'll add that here as well :) For orchestrating the frontend tests and lints, I prefer the Makefile, which I consider to be more of a more cross-language tool than tox.

yagebu:
> The test speedups weren't quite as blatant as I was hoping for, so I let it drop... 

The main "speedup" I've achieved here is to only run the "expensive" tests once for windows and macos (with Python 3.9) and run the tests on Python 3.6 without the ones with the ffmpeg and imagemagick deps. That way, one gets much quicker feedback if there's some platform-specific failures on Windows or macOS.

It seems you tried some more elaborate approaches to speeding these runs up, but if the speedups were only marginal, I think I'd prefer this shorter ci.yml here. Let me know if there's anything in particular, which worked well in your config that I should include.

dairiki:
Other points from my work on tox/CI:

I was moving toward specifying the python test requirements in the tox.ini, rather than in the [test] extras_require.
(I realize this is controversial, but I feel the test requirements don't belong in extras_require.  Extras_require is part of the distribution metadata, visible to all users of PyPI.  Unless the tests are part of the source tree (i.e. included in the sdist) — which they are not — that information is useless (and confusing) to PyPI users.)

The tox.ini tries to run tests, as much as possible, identically to the way they are run in the CI tests.  E.g. the frontend tests run `npm ci` (and take some care not to blow-away any work-in-progress in the current local source tree in so doing.)

yagebu:
> I was moving toward specifying the python test requirements in the tox.ini, rather than in the [test] extras_require.

:+1:, I agree - done here as well.

> The tox.ini tries to run tests, as much as possible, identically to the way they are run in the CI tests. E.g. the frontend tests run npm ci (and take some care not to blow-away any work-in-progress in the current local source tree in so doing.)

I've now changed the node CI parts to use `make` targets directly - the impact of `npm i` instead of `npm ci` should be negligible - to align local and CI use a bit further :)

Makefile
```diff
@@ -9,34 +10,41 @@ lektor/admin/node_modules: lektor/admin/package-lock.json
...
+lint:
+	pre-commit run -a
+	tox -e lint
+
+.PHONY: test
+test: lint test-python test-js
+
+.PHONY: test-all
+# Run tests on all supported Python versions.
+test-all: lint test-js
```
dairiki: The normal way to "test all" in a python project which uses `tox` is simply to run `tox` (with no args).  (That runs all the environments listed in the `[tox]envlist` setting.)  I think having to run `make test-all` to do this is fairly counter-intuitive.

I can see arguments both ways as to whether tox should be used to run the JS tests (though I lean towards using tox for those, too) but I think `tox.ini` should be the canonical "top-level" test configuration for the python tests, at least.  (I.e. just running `tox` should run, essentially, all the tests — or at least all the python tests — that are run by CI.)

yagebu: I am aware of that. However, since make acts as a the top level test entry here (and I don't think it makes sense to test all Python versions by default), I've added it here. I can also add the Python part to tox.

PR #933, opened 2021-09-02
Use setuptools-scm to deduce distribution version and PEP-517-ify the packaging

dairiki:
Instead of hardcoding the distribution version in `setup.py`, use [setuptools-scm](https://github.com/pypa/setuptools_scm/) to deduce the version automatically (from git tags).

This will help with our move towards automatic tag-to-release CI (per discussion in #724.)

## Questions

Should we PEP-517ize, too?  (I.e. add a `pyproject.toml` and convert `setup.py` to `setup.cfg`?)
(I think we should.  I will work on that...)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,71 @@
+[metadata]
+name = Lektor
+description = A static content management system.
+long_description = file: README.md
+long_description_content_type = text/markdown
+license = BSD
+platforms = any
+author = Armin Ronacher
+author_email = armin.ronacher@active-4.com
+url = http://github.com/lektor/lektor/
+classifiers =
+    Framework :: Lektor
+    Environment :: Web Environment
+    Intended Audience :: Developers
+    License :: OSI Approved :: BSD License
+    Operating System :: OS Independent
+    Programming Language :: Python
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Topic :: Internet :: WWW/HTTP :: Dynamic Content
+    Topic :: Software Development :: Libraries :: Python Modules
+requires = setuptools_scm
+
+[options]
+packages = find:
+include_package_data = True
+zip_safe = False
+python_requires = >=3.6
+install_requires =
+    Babel
+    click>=6.0
+    EXIFRead
+    filetype>=1.0.7
+    Flask
+    inifile
+    Jinja2>=3.0
+    mistune>=0.7.0,<2
+    pip
+    python-slugify
+    requests[security]
+    setuptools
+    watchdog
+    Werkzeug<3
+setup_requires =
+    # XXX: b/c: This duplicates config in pyproject.toml and is only
+    # used when running the stub setup.py directly
+    setuptools >= 45
+    setuptools_scm >= 6
+
+[options.extras_require]
+ipython =
+    ipython
+# XXX: the [test] extras is deprecated
+# Ref: https://github.com/lektor/lektor/pull/931#issuecomment-911879538
+test =
+    pre-commit
+    pylint==2.7.0
+    pytest
+    pytest-cov
+    pytest-mock
+    pytest-click
+
+[options.entry_points]
+console_scripts = 
+    lektor = lektor.cli:main
+
 [flake8]
 ignore=E203,F403,E128,E126,E123,E121,E265,E501,N802,N803,N806,C901,D100,D102,D102,D10,W503,E731,E402,E741
--- a/setup.cfg
+++ b/setup.cfg
@@ -66,6 +66,3 @@ test =
 [options.entry_points]
 console_scripts = 
     lektor = lektor.cli:main
-
-[flake8]
-ignore=E203,F403,E128,E126,E123,E121,E265,E501,N802,N803,N806,C901,D100,D102,D102,D10,W503,E731,E402,E741
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,15 +52,6 @@ setup_requires =
 [options.extras_require]
 ipython =
     ipython
-# XXX: the [test] extras is deprecated
-# Ref: https://github.com/lektor/lektor/pull/931#issuecomment-911879538
-test =
-    pre-commit
-    pylint==2.7.0
-    pytest
-    pytest-cov
-    pytest-mock
-    pytest-click
 
 [options.entry_points]
 console_scripts = 
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,7 +22,6 @@ classifiers =
     Programming Language :: Python :: 3.9
     Topic :: Internet :: WWW/HTTP :: Dynamic Content
     Topic :: Software Development :: Libraries :: Python Modules
-requires = setuptools_scm
 
 [options]
 packages = find:
--- a/setup.py
+++ b/setup.py
@@ -1,74 +1,18 @@
-import io
-
-from setuptools import find_packages
+# Stub setup.py.
+#
+# XXX: This is deprecated in favor of PEP-517-aware tools and should probably be deleted.
+#
+# The canonical way to build an sdist and wheel is now:
+#
+#     pip install build
+#     python -m build
+#
+# NB: Recent versions of pip are PEP-517-aware so that:
+#
+#     pip install -e .
+#
+# will work fine without this stub.
+#
 from setuptools import setup
 
-with io.open("README.md", "rt", encoding="utf8") as f:
-    readme = f.read()
-
-tests_require = [
-    "pre-commit",
-    "pylint==2.7.0",
-    "pytest",
-    "pytest-cov",
-    "pytest-mock",
-    "pytest-click",
-]
-
-setup(
-    name="Lektor",
-    use_scm_version=True,
-    url="http://github.com/lektor/lektor/",
-    description="A static content management system.",
-    long_description=readme,
-    long_description_content_type="text/markdown",
-    license="BSD",
-    author="Armin Ronacher",
-    author_email="armin.ronacher@active-4.com",
-    packages=find_packages(),
-    include_package_data=True,
-    zip_safe=False,
-    platforms="any",
-    python_requires=">=3.6",
-    setup_requires=["setuptools_scm"],
-    install_requires=[
-        "Babel",
-        "click>=6.0",
-        "EXIFRead",
-        "filetype>=1.0.7",
-        "Flask",
-        "inifile",
-        "Jinja2>=3.0",
-        "mistune>=0.7.0,<2",
-        "pip",
-        "python-slugify",
-        "requests[security]",
-        "setuptools",
-        "watchdog",
-        "Werkzeug<3",
-    ],
-    tests_require=tests_require,
-    extras_require={
-        "ipython": ["ipython"],
-        "test": tests_require,
-    },
-    classifiers=[
-        "Framework :: Lektor",
-        "Environment :: Web Environment",
-        "Intended Audience :: Developers",
-        "License :: OSI Approved :: BSD License",
-        "Operating System :: OS Independent",
-        "Programming Language :: Python",
-        "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.6",
-        "Programming Language :: Python :: 3.7",
-        "Programming Language :: Python :: 3.8",
-        "Programming Language :: Python :: 3.9",
-        "Topic :: Internet :: WWW/HTTP :: Dynamic Content",
-        "Topic :: Software Development :: Libraries :: Python Modules",
-    ],
-    entry_points="""
-        [console_scripts]
-        lektor=lektor.cli:main
-    """,
-)
+setup(use_scm_version=True)
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ tests_require = [
 
 setup(
     name="Lektor",
-    version="4.0.dev",
+    use_scm_version=True,
     url="http://github.com/lektor/lektor/",
     description="A static content management system.",
     long_description=readme,
@@ -30,6 +30,7 @@ setup(
     zip_safe=False,
     platforms="any",
     python_requires=">=3.6",
+    setup_requires=["setuptools_scm"],
     install_requires=[
         "Babel",
         "click>=6.0",
--- a/tox.ini
+++ b/tox.ini
@@ -26,12 +26,11 @@ commands =
     pylint {posargs:lektor setup.py tests}
 
 [flake8]
+max-line-length = 100
 ignore =
     # E203: Whitespace before ':'
     E203,
     # E402: Module level import not at top of file
     E402,
-    # E501: Line too long
-    E501,
     # W503: Line break occurred before a binary operator
     W503
--- a/tox.ini
+++ b/tox.ini
@@ -24,3 +24,48 @@ deps =
     pytest
 commands =
     pylint {posargs:lektor setup.py tests}
+
+[flake8]
+ignore =
+    # C901: Function is too complex
+    C901,
+    # E121: Continuation line under-indented for hanging indent
+    E121,
+    # E123: Closing bracket does not match indentation of opening bracket's line
+    E123,
+    # E126: Continuation line over-indented for hanging indent
+    E126,
+    # E128: Continuation line under-indented for visual indent
+    E128,
+    # E203: Whitespace before ':'
+    E203,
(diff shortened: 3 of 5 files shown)


yagebu:
Nice, this was on my todo list as well :)

dairiki:
I've started playing around in [dairiki/lektor:feature.custom-build-backend](https://github.com/dairiki/lektor/tree/feature.custom-build-backend) with a custom PEP-517 backend which automatically ensures that the admin UI static files are (re)built when building an *sdist* or *wheel*.

I.e. with the changes in that branch, (assuming Node and [build](https://pypi.org/project/build/) are already installed), the following is sufficient to build a working distribution (no manual `make` or `npm` invocation is necessary):
```sh
git clone https://github.com/lektor/lektor.git
python -m build lektor
```

Is there interest in this?

setup.cfg
```diff
@@ -1,2 +1,71 @@
...
+    pytest
+    pytest-cov
+    pytest-mock
+    pytest-click
+
+[options.entry_points]
+console_scripts = 
+    lektor = lektor.cli:main
+
 [flake8]
```
dairiki: Should the `[flake8]` config be moved to `tox.ini`?

Flake8 will pick it up from [either place](https://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations).

It might be cleaner to keep the packaging config in `setup.cfg` and the test config in `tox.ini`.

dairiki:
On a side note: the Windows CI tests seem to be failing sporadically due to failures installing imagemagick/ffmpeg with Chocolately.  My guess is that the failures are due to intermittent lagginess from the  Chocolatey Community Package Repository.  (Shortly after one of the CI test failures I was attempting to search for a package on the repo website (e.g. https://community.chocolatey.org/packages?q=imagemagick) and the search page seemed unresponsive.)

andoresuperesu:
Could those be solved using winget? I tend to get very good installs using it, however ffmpeg isn't included in that package manager yet.

setup.cfg
yagebu: Sounds good

dairiki: I've now moved the flake8 config to tox.ini.

In so doing, I noticed that there were a lot of apparently extraneous ignored rules: some for rules from flake8 plugins which we don't currently use, some for rules which are not currently matched by our codebase (many of those are probably auto-fixed by the use of black.)

In a series of commits, I've cleaned up the unused ignores.

yagebu:
> On a side note: the Windows CI tests seem to be failing sporadically due to failures installing imagemagick/ffmpeg with Chocolately. 

I've allowed that install step to fail in a5d98588279ae18faa359cfb9c466679b8c9fe79, so those should run through now.

> I've started playing around in dairiki/lektor:feature.custom-build-backend with a custom PEP-517 backend which automatically ensures that the admin UI static files are (re)built when building an sdist or wheel.

I'm not sure that the ecosystem is ready for that yet (and it's good to have an equivalent setup.py stub still). Also, this would force us to install node on all CI builds, right? That seems to be a bit overkill considering that there's already an equivalent one-liner: `make dist`.

Also I believe for the install to now properly run in CI, we'll need to do a full checkout (with `fetch-depth: 0`, see, e.g., https://github.com/lektor/lektor-atom/commit/8d040aa4c46916662e8218a2d2beb336c832bf5d. There might be a more elegant solution at some later point... (https://github.com/pypa/setuptools_scm/issues/431)

dairiki:
> I'm not sure that the ecosystem is ready for that yet (and it's good to have an equivalent setup.py stub still).

Which ecosystem are you referring to?
We generally install from git (for dev and testing) using pip.  Pip is PEP517-aware (and works fine without a setup.py.)
As far as packaging for PyPI, the project gets to define how that gets done.  Once the wheel/sdist is published to PyPI it shouldn't matter how it was generated.

> Also, this would force us to install node on all CI builds, right? That seems to be a bit overkill considering that there's already an equivalent one-liner: `make dist`.

As my experimental branch stands, yes.

PEP-517 specifies [a mechanism](https://www.python.org/dev/peps/pep-0517/#config-settings) for passing custom options to build backends. This would allow for passing a `no-build-js=1` or similar option to the custom build backend to cause it to skip the webpack run.
In theory, I really like this since it makes it hard to build a broken wheel/sdist without explicit action by the user.  In practice, it appears that, as of yet, `pip` [provides no way](https://github.com/pypa/pip/issues/5771) to set build backend `config_settings` on the `pip install` command line.

I suppose we could pass a `no-build-js` option to the build backend as an environment variable.

Another way to go would be to run the node tests and the python tests in the same workflow job instead of having separate jobs for each.  (That would prevent "wasting" the node install just for the python test job.)

> Also I believe for the install to now properly run in CI, we'll need to do a full checkout (with `fetch-depth: 0`, see, e.g., [lektor/lektor-atom@8d040aa](https://github.com/lektor/lektor-atom/commit/8d040aa4c46916662e8218a2d2beb336c832bf5d). There might be a more elegant solution at some later point... ([pypa/setuptools_scm#431](https://github.com/pypa/setuptools_scm/issues/431))

When run in a shallow clone, `setuptool-scm` will issue a warning to stderr, and will generate a nonsense version number (something like `0.1.dev1+gXXXXXXX`, where XXXXXXX is a git commit hash) but I don't *think* that should break the CI tests.

setup.py
```diff
@@ -1,73 +1,18 @@
...
+# Stub setup.py.
+#
+# XXX: This is deprecated in favor of PEP-517-aware tools and should probably be deleted.
+#
+# The canonical way to build an sdist and wheel is now:
+#
+#     pip install build
+#     python -m build
+#
+# NB: Recent versions of pip are PEP-517-aware so that:
```
webknjaz, line 10: Exactly so it could be reasonable to just delete it and assume that the contributors are capable of upgrading their pip.

tox.ini
```diff
@@ -22,3 +22,13 @@ paths =
 # NB: (Pinned) pylint is installed as part of the lektor[test] extras
 commands =
     pylint {posargs:lektor setup.py tests}
+
+[flake8]
+max-line-length = 100
+ignore =
```
webknjaz: 1) use
```suggestion
extend-ignore =
```
to append to the default list of ignores instead of replacing it.
2) it's usually _much_ easier to find this config if it's in a dedicated file

tox.ini
```diff
@@ -22,3 +22,13 @@ paths =
 # NB: (Pinned) pylint is installed as part of the lektor[test] extras
 commands =
     pylint {posargs:lektor setup.py tests}
+
+[flake8]
+max-line-length = 100
```
webknjaz: Is this the current standard? If so, it's not very friendly accessibility-wise. It may cause poor developer experience for people who use large fonts and vertical panels with 2-3 files open side-by-side sometimes.

setup.cfg
```diff
@@ -1,2 +1,68 @@
...
+    Operating System :: OS Independent
+    Programming Language :: Python
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Topic :: Internet :: WWW/HTTP :: Dynamic Content
+    Topic :: Software Development :: Libraries :: Python Modules
+requires = setuptools_scm
```
webknjaz: is this valid?

dairiki: No.  I think it's cruft.  I'm not sure why I put it there.  Will delete.

tox.ini
dairiki: I copied it from the existing [pylintrc](https://github.com/lektor/lektor/blob/14324ab33f51d488783efd8d162f199e3c784eff/pylintrc#L5).  (The `pylintrc` should probably be deleted, since I don't think we currently use pylint.)

The current code-base has python lines of up to 91 characters in length.
It seems that `black` wraps (by default?) to a line-length of 88.  The lines in our current code-base longer than that appear to be comments and strings (which `black` is apparently okay with, I guess.)

Our `.editorconfig` sets a line-length of 88.

What should we use for a line length?

tox.ini
dairiki: > 1. use extend-ignore

+1

Thank you




> 2) it's usually _much_ easier to find this config if it's in a dedicated file

You mean put the flake8 config in `.flake8`, I assume.
Personally, I don't like project-local configs in dot files, as I tend not to notice that they are there.   But I don't have strong feelings.  I do kind of like to have all the dev/test tool config in one place — but, even if you agree with me on that, that's not always possible these days.
I don't have strong feelings either way.  Tell me where to put the flake8 config and I'll put it there.

setup.py
dairiki: If there is consensus on that, I'm more than happy to delete the setup.py.

tox.ini
yagebu: > Personally, I don't like project-local configs in dot files

:+1:

setup.py
yagebu: Please no, I still want to be able to do editable installs.

tox.ini
yagebu: > The pylintrc should probably be deleted, since I don't think we currently use pylint.

We still do use pylint.

>It seems that black wraps (by default?) to a line-length of 88. 

Yes, indeed (see https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length for why it uses a value higher than PEP8). Let's just set the lowest value that doesn't throw up any errors (so 91 I guess) in this PR since that seems to be a rather unrelated topic. In general I'd be fine with either the black default of 88 or 79 or 80 but I also assume that the 10% difference won't be too stark accessibility-wise.

yagebu:
> Which ecosystem are you referring to?
We generally install from git (for dev and testing) using pip. Pip is PEP517-aware (and works fine without a setup.py.)
As far as packaging for PyPI, the project gets to define how that gets done. Once the wheel/sdist is published to PyPI it shouldn't matter how it was generated.

pip usually comes with Python so people often don't have the latest and greatest installed, so I'd always be careful with requirements on a recent pip. I guess PEP517 is not that recent anymore but editable installs still only work with setup.py. I agree that for end-users getting the distribution via PyPI it wouldn't make a difference, but for me as a developer, not being able to do editable installs (without keeping the stub setup.py with a different behaviour) is a deal-breaker.

> In theory, I really like this since it makes it hard to build a broken wheel/sdist without explicit action by the user.

I agree. But I still want my setup.py for editable installs and I worry that having two inconsistent ways of installation is worse than the current situation.

> Another way to go would be to run the node tests and the python tests in the same workflow job instead of having separate jobs for each. (That would prevent "wasting" the node install just for the python test job.)

Yes, that would work :)

> When run in a shallow clone, setuptool-scm will issue a warning to stderr, and will generate a nonsense version number (something like 0.1.dev1+gXXXXXXX, where XXXXXXX is a git commit hash) but I don't think that should break the CI tests.

Yep, I've been running CI builds with setuptools-scm in shallow clones for years and they don't break (well I think there was some breakage at some point), but it would be nice to see that it works correctly (particularly if we want to do automatic deployments at some point). If the full clone doesn't take too long I think it would be worth it.

dairiki:
> pip usually comes with Python so people often don't have the latest and greatest installed, so I'd always be careful with requirements on a recent pip. I guess PEP517 is not that recent anymore but editable installs still only work with setup.py. I agree that for end-users getting the distribution via PyPI it wouldn't make a difference, but for me as a developer, not being able to do editable installs (without keeping the stub setup.py with a different behaviour) is a deal-breaker.

Normal installation (installing from a PyPI wheel) will, I believe, still work fine without a `setup.py`, even with older pips.  A wheel is a wheel...

Editable installs work fine without a `setup.py`, at least with recent `pip`s.  (I must admit that I don't totally understand how this works, but it does...)

tox.ini
dairiki: > We still do use pylint.

Maybe this is a discussion for another thread, but why are we using both `pylint` and `flake8`?  Isn't one or the other sufficient?

setup.py
dairiki: I don't completely understand how it works, but recent versions of `pip` will perform editable installs just fine without a `setup.py`.

yagebu:
> Editable installs work fine without a setup.py, at least with recent pips. (I must admit that I don't totally understand how this works, but it does...)

Oh wow - I tried it assuming that my Arch Linux-installed pip would be recent enough but I guess it's still the pip version that came with Python 3.9 and not recent enough.

I'd be for making the switch then I guess, but let's do that in a followup PR, ok?

tox.ini
yagebu: IMHO pylint would be enough since we're also using black so there's not much formatting stuff to complain about for flake8. It's still good to have config for both so that they work for contributors who have these linters installed as editor plugins or such.

setup.py
yagebu: If a setup.cfg but no setup.py exists then pip will basically just create a stub like this setup.py here

webknjaz:
> Editable installs work fine without a setup.py, at least with recent pips. (I must admit that I don't totally understand how this works, but it does...)

Pip special-cases setuptools to make it work. It won't work with other backends, though. The mechanism is still not standardized but there's some work in progress to improve this.

setup.py
dairiki: Leaving the removal of `setup.py` for a subsequent PR.

tox.ini
dairiki: Leaving flake8 config in `tox.ini` for now.

I've switched to using `extend-ignore`.

setup.cfg
dairiki: Deleted.    I'm not sure how that got there.  Thank you for pointing it out!

dairiki:
Windows tests are now failing.

It appears that the errors are from when `pip` tries to parse the output of `python setup.py sdist`.  I believe this issue is triggered by non-ASCII in the output.  On windows it appears to be being written in latin1 (or similar) — pip assumes it is utf-8 when parsing the output, leading to a `UnicodeDecodeError`.  I think this is triggered by lines in the output (e.g. "creating ...", "copying ...", "adding ...") having to do with our `tests/ünicöde-project` subdirectory.

I have no idea why this seems only to happen with this PR.

Perhaps this can be worked around by setting an appropriate environment variable (e.g. PYTHONIOENCODING)?
[Edited (again):] Setting `PYTHONIOENCODING=utf-8` appears to be an effective work-around.  (When I tested this the first time, I was, apparently, not actually successfully setting the environment variable in the windows test runner.)

[Further edit:] I found tox-dev/tox#1550 which reports a similar but different encoding mismatch issue. There, the work-around of pinning `tox==3.8.3` is suggested.  I've just tried that, and it appears to work as well.  (I've now gone with setting PYTHONIOENCODING rather than pinning pip.)

[More edit:] Conclusion: I believe this is a tox bug.  It seems to be exercised, in our case, at least, by using setuptools-scm when there are non-ascii file names in the distribution under Windows.  Setting PYTHONIOENCODING=utf-8 appears to be an effective work-around.

dairiki:
I have pushed a commit which removes the [test] extras_require settings.  (The test dependencies are now declared explicitly in `tox.ini`.)

I believe this was discussed previously, and that there was general buy-in, but I can not, now, find that discussion.  If there's objection to removing the test extras, I can put them back.

dairiki:
> > On a side note: the Windows CI tests seem to be failing sporadically due to failures installing imagemagick/ffmpeg with Chocolately.
> 
> I've allowed that install step to fail in [a5d9858](https://github.com/lektor/lektor/commit/a5d98588279ae18faa359cfb9c466679b8c9fe79), so those should run through now.

That does "fix" the test failures, though I notice that the chocolatey installation of imagemagick on windows seems to be failing pretty much all the time these days.  It's a little scary just to ignore those failures.  Part of the point of testing on Windows is to make sure our interaction with the OS (e.g. things like running the imagemagick subprocess) work correctly.

tox.ini
dairiki: I say we should pick one or the other between flake8 and pylint.

Also, since we are using pre-commit to enforce these sorts of things, whatever we decide to use should be included in the pre-commit config so that problems surface at the pre-commit stage rather than at the CI testing stage (or wherever).

(In any case, I think this is out-of-scope so far as this particular PR is concerned.)

yagebu (approved):
LGTM :)

> It's a little scary just to ignore those failures.

Indeed. It's also a little scary that the "standard" package manager on Windows fails so often. I'll open an issue to track this.

tox.ini
```diff
@@ -24,3 +24,11 @@ deps =
     pytest
 commands =
     pylint {posargs:lektor setup.py tests}
+
+[flake8]
+max-line-length = 91
```
webknjaz, line 29: Alternatively, `flake8-length` is more flexible

PR #939, opened 2021-09-18
Add pyupgrade pre-commit hook

yagebu:
This hook automatically updates Python files with modern syntax, e.g., f-strings and sets.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,11 @@ repos:
     hooks:
       - id: flake8
         language_version: python3
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v2.25.0
+    hooks:
+      - id: pyupgrade
+        args: ["--py36-plus"]
   - repo: local
     hooks:
       - id: eslint
--- a/lektor/assets.py
+++ b/lektor/assets.py
@@ -87,7 +87,7 @@ class Asset(SourceObject):
         return None
 
     def __repr__(self):
-        return "<%s %r>" % (
+        return "<{} {!r}>".format(
             self.__class__.__name__,
             self.artifact_name,
         )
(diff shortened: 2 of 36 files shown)


dairiki:
As you might expect, I'm not a huge fan of this.  It touches a lot of code for arguable benefit.

Mostly, though, I would really like to see a working release make it out the door before we make too many more changes.

yagebu:
> Mostly, though, I would really like to see a working release make it out the door before we make too many more changes.

I already released 3.2.2 with the most important bugfixes

.pre-commit-config.yaml
```diff
@@ -12,6 +12,11 @@ repos:
     hooks:
       - id: flake8
         language_version: python3
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v2.25.0
+    hooks:
+      - id: pyupgrade
+        args: ["--py36-plus"]
```
dairiki, line 19: Should be `--py37-plus` (though, in practice, that doesn't seem to change anything) since we no longer support python 3.6.

lektor/assets.py
```diff
@@ -87,7 +87,7 @@ def resolve_url_path(self, url_path):
         return None
 
     def __repr__(self):
-        return "<%s %r>" % (
+        return "<{} {!r}>".format(
```
dairiki, line 90: There are a number of these cases where `%`-formatted strings are converted to `.format()` formatted strings rather than f-strings.

The `pyupgrade` README does [say](https://github.com/asottile/pyupgrade#f-strings):

> note: pyupgrade is intentionally timid and will not create an f-string if it would make the expression longer or if the substitution parameters are sufficiently complicated (as this can decrease readability).

That said, a lot of these cases seem like they are short enough that they would, in fact, be cleaner as f-strings.  Should we manually convert them?

yagebu:
I'm closing this one. I still think it makes sense to have these lint rules but I think it makes more sense to use the newer, faster (not that pyupgrade wasn't fast enough already) and more powerful ruff.

PR #942, opened 2021-09-22
Remove setup.py

dairiki:
PR #933 started the process of [PEP-517][pep517]-ifying our build process, converting to setuptools declarative configuration (configuration in `setup.cfg`) and specifying build dependencies in `pyproject.toml`.  This PR completes the process by removing `setup.py`.

Notes:
- The canonical way to build a PEP-517-aware distribution is to use build.
- Recent versions of pip are PEP-517-aware and can install from source (e.g. in editable mode) without a `setup.py`.

See comments on #933 for further discussion.

[pep517]: <https://www.python.org/dev/peps/pep-0517/> (PEP 517 -- A build-system independent format for source trees)


yagebu (approved):
LGTM

PR #944, opened 2021-10-10
Support newer mistune versions in addition to older ones

evilham:
### Description of Changes

It looks like lepture has been busy and mistune 2.0.0 is about to be
released, there is already some software that depends on its alpha/rc
versions and rather than backporting that, it makes more sense to
prepare Lektor to be compatible with the newer version.

This patch aims to introduce minimal changes to achieve that while
keeping compatibility with mistune 0.8.X.

As a curiosity and the reason why I personally am interested in upstreaming these changes is being able to use [md2gemini](https://github.com/makeworld-the-better-one/md2gemini/) from Lektor with a plugin I just published today: https://pypi.org/project/lektor-gemini-capsule/ which effectively turns Lektor into a dual format (HTML and Gemini) publishing software.

With this patch it works wonders!

### DIFF
--- a/lektor/markdown.py
+++ b/lektor/markdown.py
@@ -11,7 +11,7 @@ from lektor.context import get_ctx
 
 _markdown_cache = threading.local()
 
-_old_mistune = int(mistune.__version__.split('.')[0]) < 2
+_old_mistune = int(mistune.__version__.split('.', maxsplit=1)[0]) < 2
 
 if _old_mistune:
     from mistune import Renderer as BaseRenderer
@@ -20,7 +20,7 @@ else:
 
 
 class ImprovedRenderer(BaseRenderer):
-    def link(self, link, text=None, title=None):  # pylint: disable=arguments-differ
+    def link(self, link, text=None, title=None):  # pylint: disable=arguments-differ, arguments-renamed
         if _old_mistune:
             (title, text) = (text, title)
         if self.record is not None:
@@ -33,7 +33,7 @@ class ImprovedRenderer(BaseRenderer):
         title = escape(title)
         return '<a href="%s" title="%s">%s</a>' % (link, title, text)
 
-    def image(self, src, alt="", title=None):  # pylint: disable=arguments-differ
+    def image(self, src, alt="", title=None):  # pylint: disable=arguments-differ, arguments-renamed
         if _old_mistune:
             (title, alt) = (alt, title)
         if self.record is not None:
--- a/lektor/markdown.py
+++ b/lektor/markdown.py
@@ -11,41 +11,50 @@ from lektor.context import get_ctx
 
 _markdown_cache = threading.local()
 
-_old_mistune = int(mistune.__version__.split('.', maxsplit=1)[0]) < 2
+_old_mistune = int(mistune.__version__.split(".", maxsplit=1)[0]) < 2
 
 if _old_mistune:
-    from mistune import Renderer as BaseRenderer
+
+    class ImprovedRenderer(mistune.Renderer):
+        def link(self, link, title, text):
+            return _render_link(link, text, title, record=self.record)
+
+        def image(self, src, title, text):
+            return _render_image(src, text, title, record=self.record)
+
 else:
-    from mistune import HTMLRenderer as BaseRenderer  # pylint: disable=no-name-in-module
+
+    class ImprovedRenderer(mistune.HTMLRenderer):
+        def link(self, link, text=None, title=None):
+            return _render_link(link, text, title, record=self.record)
+
+        def image(self, src, alt=None, title=None):
+            return _render_image(src, alt, title, record=self.record)
 
 
-class ImprovedRenderer(BaseRenderer):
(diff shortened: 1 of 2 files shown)


evilham:
Oh, it looks like these changes broke the linter; I'm unsure about how to fix that since the linter is running against mistune 0.8.4 (which makes sense!). Any suggestions?
(Thank you for enabling the workflow!)

yagebu:
I think in this case it's fine to just add disable comments (which look like ` # pylint: disable=no-name-in-module`).

evilham:
It looks like running `tox -e lint` on my system is not really checking the same as the workflow, sorry about the noise.
Particularly the `arguments-renamed` warning is not raised on my system and if I try to silence it it turns into an error '-.- hopefully like this it passes in the workflow.
(Packages are installed in a venv and I run it as `tox -e lint`)

lektor/markdown.py
```diff
@@ -11,9 +11,18 @@
...
+if _old_mistune:
+    from mistune import Renderer as BaseRenderer
+else:
+    from mistune import HTMLRenderer as BaseRenderer  # pylint: disable=no-name-in-module
+
+
+class ImprovedRenderer(BaseRenderer):
+    def link(self, link, text=None, title=None):  # pylint: disable=arguments-differ, arguments-renamed
+        if _old_mistune:
+            (title, text) = (text, title)
```
dairiki: This swapping of arguments makes me nervous.  I think it's true that mistune always calls .link with all positional arguments (no named arguments), but that is not clear from inspection.  `title, text = text, title` doesn't really defuse the confusion.

It would be nice to keep the call signatures of the methods identical to those of the underlying renderer.

A couple of alternative ideas:

My preference, I think (at least, right now) would be something like:
```py
if _old_mistune:
    class ImprovedRenderer(mistune.Renderer):
        def link(self, link, title, text):
            return _render_link(link, text, title, record=self.record)
        def image(self, src, title, text):
            return _render_image(src, text, title, record=self.record)
else:
    class ImprovedRenderer(mistune.HTMLRenderer):
        def link(self, link, text=None, title=None):
            return _render_link(link text, title, record=self.record)
        def image(self, src, alt="", title=None):
            return _render_image(src, alt, title, record=self.record) 
        ...

def _render_link(link, text=None, title=None, record=None):
    if record is not None:
        ...


def _render_image(src, alt="", title=None, record=None):
    ...
```

A more compact — but I think more half-assed — alternative would be something like: 
```py
class ImprovedRenderer(BaseRenderer):
    def _link(self, link, text=None, title=None):
        if self.record is not None:
            ...

    if _old_mistune:
        def link(self, link, title, text):
            return self._link(link, text, title)
    else:
        link = _link
```

dairiki: Note that [mistune 1.x version of link][1] does an
```py
if text is None:
    text = link
```
around here.  The implication is that sometimes this gets called with just a `link` (URL) and no text, in which case we want to display the URL for the text of the link.  I suspect this might be needed sometimes.

[1]: <https://github.com/lepture/mistune/blob/46a65a978b005651dcc54a340e44aef71f9f7e95/mistune/renderers.py#L137>

lektor/markdown.py
```diff
@@ -24,17 +33,19 @@ def link(self, link, title, text):
         title = escape(title)
         return '<a href="%s" title="%s">%s</a>' % (link, title, text)
 
-    def image(self, src, title, text):
+    def image(self, src, alt="", title=None):  # pylint: disable=arguments-differ, arguments-renamed
+        if _old_mistune:
+            (title, alt) = (alt, title)
```
dairiki: As noted above, swapping the semantics of named arguments like this seems dangerous.

lektor/markdown.py
```diff
@@ -56,7 +67,9 @@ def make_markdown(env):
     env.plugin_controller.emit("markdown-config", config=cfg)
     renderer = cfg.make_renderer()
     env.plugin_controller.emit("markdown-lexer-config", config=cfg, renderer=renderer)
-    return mistune.Markdown(renderer, **cfg.options)
+    if _old_mistune:
+        return mistune.Markdown(renderer, **cfg.options)
+    return mistune.Markdown(renderer)
```
dairiki: Mistune v1.x appears to support plugins.  Those would (I think) be passed to the `Markdown` constructor.
I'm not familiar enough to say, but would there be worth in adding the hooks to `MarkdownConfig` so that Lektor plugins can configure the use of Mistune plugins?

evilham: Thank you for the general review, it is very useful and makes the PR more maintainable.

I am also not super familiar with these particular bits, but you are right to point out that we were ignoring configuration data.
I found this bit of documentation (might be new, or maybe I'm fresh and noticed it today :-D) https://mistune.readthedocs.io/en/latest/api.html#mistune.create_markdown

Which seems somewhat compatible with the way the previous version was being used: https://mistune.readthedocs.io/en/v0.8.4/#options

dairiki:
This is fixed in master by #992.  Closing.

Issue #410, opened 2017-07-18
lektor build and lektor deploy do not detect changes of macros and templates


torian257x:
neither in backend, nor from command line.

I am using rsync to deploy

big show stopper

If I update a macro file or a page template file and do not visit it via browser, e.g. I do not run

`lektor server`

then go to the url

I do not get updates at all of the files

`yes | lektor clean && lektor build && lektor deploy`

doesn't work

`lektor build && lektor deploy`

doesn't work

`lektor build --output-path ./build && lektor deploy --output-path ./build`

doesn't work

what means 'doesnt work`?

doesn't work means, the new file version of the template file is not used, the old one is deployed instead.

what does work?

if I recompile my css files from the scss originals, that file change is recognized in the assets folder.

what does work 2?

if I use

`lektor server`

then visit the file that makes use of a changed template, it will update to the new version, and then running `lektor build && lektor deploy` will deploy that newly correctly created version of that file.

Is that a good workaround?
No! If I change a macro, do you want me to revisit _all_ blog posts _each_ time? this needs to be fixed

Do the admin backend cms "refresh build" work?

No! neither they work.

there is more discussion on gitter

nixjdm:
Thanks for finding and posting this @Jossnaz. Certainly sounds bad. @smitty42, tagging you if you can get to this faster than I can.

wgwz:
Can you post details about the version of Lektor you are using and how it was installed? I use roughly the same workflow, without any problems AFAIK.

torian257x:
Lektor, version 3.0.1

i followed:

https://www.getlektor.com/docs/installation/

so

`sudo apt-get install python-dev libssl-dev libffi-dev`
`sudo apt-get install imagemagick`


actually thinking about it, because it is highlighted this way, I think I first ran:

`curl -sf https://www.getlektor.com/install.sh | sh`

and then afterwards added the missing packages when things failed and tried again.

what surprises me is that css changes are detected. But template changes aren't

torian257x:
you cannot reproduce? It is lektor with python 2.7 btw on ubuntu 14.04

torian257x:
when I run

`$ lektor build | grep index.html`

the pages I need updated are not getting listed...

torian257x:
alright I can reproduce the bug like this:


I have a content 

`mypage1`

that uses a model

`mypage1-model`

`mypage1-model` content was this:


```
[model]
name = mypage1-model
label = Federation
hidden = yes


[pagination]
[...]

PR #952, opened 2021-12-08
Paginated pages should have at least one page, even when childless

dairiki:
### Issue(s) Resolved

Pages with pagination enabled that do not have children are not being generated.

Fixes at least part of what's reported on #410.

### Description of Changes

The current pagination logic for determining the number of pages of a paginated page that are required reports that zero pages are required if the page has no children.  This causes zero pages of the paginated page to be generated.

This fixes the page-counting logic so that it always returns, at minimum, a page count of one, even when there are no children.

### Tasks (still to do)

- [x] Added a unit test exercising the bug that is being fixed


yagebu (approved):
Thanks, LGTM :)

PR #954, opened 2021-12-09
Fix #946: rsync on macos does not support --delete-delay

### LINKED ISSUE #946: rsync --delete-delay is not supported on macOS
https://github.com/lektor/lektor/blob/0ce8fd42affd53c1e45a8aa25c76f1554e3bb2c8/lektor/publisher.py#L187-L188

```man
--delete                delete extraneous files from dest dirs
--delete-before         receiver deletes before transfer (default)
--delete-during         receiver deletes during xfer, not before
--delete-after          receiver deletes after transfer, not before
--delete-excluded       also delete excluded files from dest dirs
```

... and since we are on the topic of rsync: I'd like to suggest to add support for local filesystem sync (e.g., if you are working on the same machine on which the web service is running). Currently providing a `rsync:///absolut/path` will crash Lektor. (haha, I have the same scenario as #699, currently using a custom plugin for that)

... and yet another request: I would welcome the addition of a dry-run option to see what changes exist. :)

... also: add `--no-motd`? – to omit the first few lines? =)

dairiki:
This adds machinery to detect whether the locally installed `rsync` supports `--delete-delay`.
If it does not, we fall back to using `--delete-after`.

### Issue(s) Resolved

Fixes #946

### Description of Changes

- [x] Added integration tests to exercise the reported bug

### Outstanding Questions

- [ ] Currently our CI tests do not test rsync integration on Windows, since `rsync` is not installed on the Windows runner.
      Do people use the `rsync` publisher on Windows?  If so, how do you install `rsync`?  (I am out of my depth on Windows.) I've found the Cygwin version of `rsync`, but it seems to want paths on the command line in Cygwin/unix format (forward slashes) rather than Windows format.  Is there a version of `rsync` that will work with Windows-style paths?

- [ ] What if we just always used `rsync`'s `--delete-after` instead of `--delete-delay`?  Would that give anyone gas? @m-lnrs, @xlotlu ?


yagebu:
Looks good, as usual - however this adds quite some complexity for yet unknown benefit as I don't really if know if the performance difference is really relevant.

> What if we just always used rsync's --delete-after instead of --delete-delay?

`--delete-delay` is not even the most performant delete option, that would probably be `--delete` (which means `--delete-during` on modern rsync). Only deleting after additions doesn't really save us from creating inconsistent states on the target in case of error either. So I'd be fine with just changing the default to `--delete-after` or `--delete`.

Or a completely different idea altogether: If we want to beef up the rsync publisher (and make it work with ridiculously outdated versions), why not add support for arbitrary flags and options? Why create some arbitrary mapping of "Lektor rsync publisher options" to rsync command line options? Some users might want/need some of the other hundreds of rsync command line options.

xlotlu:
Hi @dairiki, I'm with @yagebu here. This looks terribly complicated (or, umm, complex?) for such a small thing.

I'd personally quick-fix it by trying for the particular exception, and altering the command line if caught. Also, if you use [`locate_executable` you get it readily cached](https://github.com/lektor/lektor/blob/f52811d869c964e6d6110697ea76a33c3fdae8ab/lektor/utils.py#L285).

It would indeed be nice to have a full-fledged rsync class, and I bet it would be useful in more ways than one, but not for such a little thing. If deemed a worthy improvement, a separate library would be the really useful thing, and without searching I'm sure there must be a good one out there, maybe good enough to warrant inclusion as a dependency.

Regards

dairiki:
@yagebu, @xlotlu 
Okay, I've rebased, deleted my probe for supported command-line options, and have gone with just using `--delete-after` instead of `--delete-delay`.

xlotlu:
> and have gone with just using `--delete-after` instead of `--delete-delay`.

@dairiki :rofl:. +1

Issue #950, opened 2021-12-06
BUG: attaching images are stored with a broken URL


andreamoro:
Still trying to play around with Lektor and I have tried to attach an image into a post to see how this is handled.
Aside from the fact I don't see any changes in the local contents.lr file, when working with the web interface, the URL is clearly wrong as it appears nested twice under the containing parent folder, thus making the file not reachable by the webserver.

To make it clear, the local path is ..../website/content/blog/test but in the web interface the image is shown under http://127.0.0.1:5000/en/blog/test/en/blog/test/screenshot-from-2021-10-31-09-05-11.png/

See below image of the local folder

![image](https://user-images.githubusercontent.com/5256583/144921151-c52d2385-be6e-49e5-8cc7-07c7f62da55b.png)

I'm not entirely sure this is due to the alternatives settings. If I delete them from the settings file, regrettably I will lose the testing files as apparently these are deleted immediately after. 
Any idea of what the problem can be?

tomgross:
Here is a minimal example based on the  boilerplate blog:  https://github.com/tomgross/test-for-lektor-issue
See http://localhost:5000/blog/first-post/ and http://localhost:5000/de/blog/first-post/

dairiki:
Just a note: an (untested) workaround, in the interim is to link to the attachment by full absolute path.  E.g. in your `markdown` field, instead of

```md
![my screenshot](screenshot-from-2021-10-31-09-05-11.png)
```

do

```md
![my screenshot](/blog/test/screenshot-from-2021-10-31-09-05-11.png)
```

I realize that this is less than ideal, but I _think_ it should work, at least.

andreamoro:
Hi Both,

Not sure I capture the nuances in the contents+de.lr @tomgross. But I guess that's the key objective I want to investigate/improve.

I believe improvements on the alternative part should be done so to:
- have alternatives having their own url property with which internal URLs can be generated
-- whether absolute or relative, this should not make any difference
- url property to be used only during the build used for the publication
-- here correct me whether my understanding is wrong, but it seems like lektor server is using the build also internally
- find a way to move the static under the alternative folders when this is in place

I appreciate this is not the best place to discuss this feature here, so my question is: do we have a place? 
It would be great joining forces rather than creating individual forks with people inventing something that might never find place in the final code. 
Do you think we can regroup somewhere? I trust you have more experience than I do with this SSG and you may know things I barely scratched during my early days.

dairiki:
[...]

PR #958, opened 2021-12-12
Attachment.url_path incorrect when alt != PRIMARY_ALT

### LINKED ISSUE #965: Weird semantics of `pad.get()` with implicit `alt`
As things are now when alternatives are in use, `pad.get("/path")` with no explicit `alt` returns a record that does not correspond to any page actually being generated.

Let's assume some specifics, to present an example.

Assume two alternatives are configured in the project file:
```ini
[alternatives.en]
name = English
primary = yes
locale = en_US

[alternatives.fr]
name = French
url_prefix = /fr/
locale = fr
```

Let's say that for a particular page, `/page1`, all three possible data files exist.  That is to say in the directory `contents/page1/`, the three files `contents.lr`, `contents+en.lr`, and `contents+fr.lr` all exist and are non-empty.

Lektor will build two versions of this page: one for `alt="en"` and one for `alt="fr"`.  The English page uses the contents
of `contents+en.lr`, merged with fallbacks from `contents.lr`. The French page uses `contents+fr.lr` overlayed onto `contents.lr`.

Currently, the record returned by `pad.get("/page1")` reflects only the data from the `contents.lr` file.  _This does not correspond
to any of the actual output pages being generated_. 

This does not seem like a sane or useful default behavior.

In the example put forth above, I would expect `{{ site.get("/page1").some_field }}` to pull in the value of `some_field` from `contents+en.lr` (falling back to `contents.lr` only if the field has no value set in the alt-specific overlay.)  This is not what currently happens.  Currently, only `contents.lr` will be consulted for the value of the field.

### Proposal

I propose changing the signature of `Pad.get` from:
```py
def get(self, path, alt=PRIMARY_ALT, page_num=None, persist=True, allow_virtual=True):
    ...
```
to
```py
def get(self, path, alt=None, page_num=None, persist=True, allow_virtual=True):
    if alt is None:
        primary_alt = self.config.primary_alternative
        alt = primary_alt if primary_alt is not None else PRIMARY_ALT
    ...
```
         
That is, if `Pad.get()` is not passed an explicit value for `alt`, when alternatives are configured, it would return the record for the primary alternative (in the example above, `alt="en"`), rather than the fallback alt (`alt="_primary"`) as it currently does.

#### Comments

Note that in most cases, the proposed change is unlikely to affect existing users.  An overlay file typically does not exist for the primary alternative.  E.g. The complete data for the primary (e.g. English) version of the page will typically be in `contents.lr`, there will be no `contents+en.lr`.

dairiki:
Attachments are independent of `alt` — `lektor build` only emits one copy of each attachment.
Currently `Attachment.url_path` returns a path mangled as appropriate for the records `path`.  I don't think that's right.  The URL path should always point to the single URL corresponding to the <del>PRIMARY_ALT</del> <ins>primary alternative (e.g. `"en"`)</ins> [edited].

This fixes that.

[Edited to add] It also makes a few potentially non-backward-compatible changes:

#### Change `Pad.get` to return record in *primary alternative* if explicit `alt` is not specified

Previously, the `alt` parameter of `Pad.get` defaulted to `PRIMARY_ALT`.  When alternatives are not configured for a site, this is the correctly behaviour.  Records only exists in one `alt` and that is it.
However, when alternatives are in use, I think that is not the correct default.  In that case, there is never any output (any artifacts) generated for records with `alt=PRIMARY_ALT`.  Output is only generated for concrete alts, e.g., `"en"`, `"de"`, etc.  (`Alt=PRIMARY_ALT` is used to flag the fallback data which is chained under alt-specific data for each concrete alt.)

If one writes
```html+jinja
{{{ pad.get("/").title }}}
```
in some template, one does not want the fallback title.  One wants the title for the primary alternative (e.g. typically `"en"`.)

Note that often there is no specific `contents-en.lr` file for the primary alternative (assuming the primary alternative is `"en"`), in which case there is no practical difference between the records with `alt=PRIMARY_ALT` and `alt="en"`, but when the overlay file exists, we almost always want the record for `alt="en"`, not for `alt=PRIMARY_ALT`.

So this PR changes `Pad.get` so that, if no explicit `alt` is requested, the record for the primary alternative is returned.

Note that `PRIMARY_ALT` should probably be named something like `_ALT_FALLBACK`…

See #965 for further discussion.

#### Change `Pad.root` to return record in *primary alternative*.

As for `Pad.get`, `Pad.root` is changed to return the root record for the primary alternative rather than for `alt=PRIMARY_ALT`.

(Note that, as things were, the value returned by `Pad.root` was not included in those returned by `Pad.get_all_roots`.  That doesn't really make sense.)

#### Changed `AttachmentsQuery` to return attachments with the requested `alt`.

`AttachmentsQuery` was crowbarring things so that all returned attachments were in `alt=PRIMARY_ALT`.  (This may well have been typo — I'm not sure.) This has the effect (among other things) that `record.attachments` returns attachments with `alt` in the fallback alt, rather than the record's `alt`.  (For attachments with no meta-data this doesn't really make a difference, but if there is meta-data, it may well vary with alt.  If it does we probably want it for the same alt as the original record.)

### Issue(s) Resolved

This is an alternative fix for #685
Closes #965

### Related Issues / Links

This (I think) is a step on the road to fixing #950.
See also #685.

### Description of Changes

- [x] Added unit test(s) exercising the underlying bug


tomgross:
Here is a minimal example based on the boilerplate blog: https://github.com/tomgross/test-for-lektor-issue
See http://localhost:5000/blog/first-post/ and http://localhost:5000/de/blog/first-post/

yagebu (approved):
LGTM

I also only maintain monolingual sites, so thank you for the detailed reasoning for these changes, they definitely sound like a step in the right direction :)

tomgross:
Unfortunately the URLs within HTML fields still give a 404 for alternative languages.

dairiki:
> Unfortunately the URLs within HTML fields still give a 404 for alternative languages.

I think this PR is only half of the fix.  The other half is #992 which hasn't been merged yet.

(If you wanted to test #992 to see if it fixes your use case, that would be wonderful!)

PR #969, opened 2021-12-16
Refactor lektor.editor.EditorSession

dairiki:
This cleans up the code in `EditorSession` by splitting out the mapping API bits into a separate class.

This also py3k-izes the _mapping_ API provided, by doing away with the `.iter{keys,items, values}` methods and making the `.keys`, `.items`, and `.values` methods return _views_.

Fairly complete test coverage for the mapping parts of EditorSession is added, too. In so doing, a couple of minor bugs in the current implementation were detected and fixed. 

As implemented here, this does change the `EditorSession` API a bit, by moving the mapping API from the `EditorSession` itself to its `EditorSession.data` attribute.  E.g. where one before would do `editorsession.update(newdata)`, one now does `editorsession.data.update(newdata)`.

I think this change makes the API cleaner.  I think it's unlikely that any other packages are using `EditorSesssion` directly, so I think the chance of breakage is slim.  If we do want to preserve the old API where the mapping methods are directly on `EditorSession`, this would be easy enough to do: just make `EditorSession` inherit from the new `MutableEditorData` class (or if one wants to preserver the new `EditorSession.data` attribute, just proxy the mapping methods to `.data`.)


yagebu (approved):
From a quick glance this LGTM. Unless you want me to do a more thorough review, I'd be happy to have this merged :)

dairiki:
I got nervous about the chance of the API change breaking external code, so I added the mapping methods back to EditorSession.
(They just proxy to EditorSession.data, and they emit a DeprecationWarning if they are used.)

PR #974, opened 2021-12-17
Fix sneaky import cycle

dairiki:
There is a sneaky import cycle that is exposed by running:
```sh
$ python -c "import lektor.types"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/dairiki/git/web/lektor-packages/lektor/lektor/types/__init__.py", line 11, in <module>
    from lektor.types.multi import CheckboxesType
  File "/home/dairiki/git/web/lektor-packages/lektor/lektor/types/multi.py", line 4, in <module>
    from lektor.environment.expressions import Expression
  File "/home/dairiki/git/web/lektor-packages/lektor/lektor/environment/__init__.py", line 28, in <module>
    from lektor.types import builtin_types
ImportError: cannot import name 'builtin_types' from partially initialized module 'lektor.types' (most likely due to a circular import) (/home/dairiki/git/web/lektor-packages/lektor/lektor/types/__init__.py)
```

This is import-order dependent. If one imports `lektor.environment` _before_ importing `lektor.types` all is well.

(It appears this cycle was introduced in lektor/lektor@7c68f3f78.)

### Description of Changes

- [x] Added a unit test that exercises the bug
- [x] Fixed the problem

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,12 +6,17 @@ These are all the changes in Lektor since the first public release.
 
 ### Bugs Fixed
 
+- Fixed an import cycle which caused in `ImportError` if
+  `lektor.types` was imported before `lektor.environemnt`. [#974][]
+
 #### CI Tests
 
 - Increased timeout in `test_watcher.IterateInThread` to prevent
   random spurious failures during CI testing.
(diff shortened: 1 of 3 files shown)


yagebu (approved):
Good catch and that test is very good to have :)

PR #976, opened 2021-12-21
Refactor admin URLs

### LINKED ISSUE #610: Lektor backend: when using double dot (colon) in ID, interface screws up
you click add page

you enter as id

`Sometext adasdf: asdfasdf asdf asdf asdf`

the result will confuse lektor, as the `:` is used in lektor backend navigation / routing 

`:` is not treated as illegal, but should be and replaced

### LINKED ISSUE #975: Lektor Admin console edit the wrong document when alternatives are used
Using the latest version of Lektor v3.3.0, when using the Admin console to edit a page. The link return by the little pen in the top left corner is not the right URL.

e.g.: When browsing the page `http://localhost:5000/en/support/` the edit link is define as `http://localhost:5000/admin/edit?path=%2Fen%2Fsupport%2F`.

What links is wrong and doesn't allow proper editing of the page. The appropriate link should be `http://localhost:5000/admin/root:support/edit`

This is a regression when comparing to lektor v3.2.2.

Here my project file:
```
[project]
name = IKUS Software inc
name[fr] = Logiciel IKUS inc
url=https://www.ikus-soft.com/
url_style=absolute
excluded_assets=*.less

[alternatives.en]
name = English
name[fr] = English
locale = en_US
url_prefix = /en/

[alternatives.fr]
primary = yes
name = Francais
name[fr] = Francais
url_prefix = /fr/
locale = fr

[packages]
lektor-make
lektor-git-src-publisher
lektor-disqus-comments = 0.2
lektor-pythonmarkdown = 1.2
```

dairiki:
This is an experiment.  Here the admin React app URLs are refactored:
- Instead of encoding the whole Lektor path of the page to be edited into a single path component by replacing slashes with colons, allow the path to extend to an arbitrary number of path components in the React path.  Instead of `/admin/root:blog:post-1/edit`, we now use `/admin/root/blog/post-1/edit`.
- Instead of packing the _alt_ in with the Lektor path, include that as a query arg.  Where we used to use `/admin/root/about+de/edit` we now do `/admin/root/about/edit?alt=de`.

Note that the `root` is still needed in the URLs at the moment to disambiguate between, e.g.,  `/admin/edit` (the redirector view that is the target of the edit pencil) and `/admin/root/edit` (the edit view for the root page.)

Note that I have nearly zero experience with Typescript or React, so some of this code is a bit of a flail. I'm sure it can be improved.

### Issue(s) Resolved

This approach can be used to fix both #975 and #610.

Fixes #975 
Fixes #610 

### Description of Changes


dairiki:
> Note that the `root` is still needed in the URLs at the moment to disambiguate between, e.g., `/admin/edit` (the redirector view that is the target of the edit pencil) and `/admin/root/edit` (the edit view for the root page.)

I've refactored again.  React app page ("view") names are now prefixed with a `@`.  This allows us to eliminate the `/root` from the admin URLs without ambiguity between the path to the redirector (e.g. `/admin/edit?path=...`) and the _edit_ view for the root page (was `/admin/root/edit`, now `/admin/@edit`).

Examples of old and new URL scheme:

|        | Old  | New |
| ----  | -----------  | ------ |
| Edit root page | `/admin/root/edit` | `/admin/@edit` |
| Preview /blog/first-page  | `/admin/root:blog:first-page/preview` | `/admin/blog/first-page/@preview` |
| Edit French _About_ page | `/admin/root:about+fr/edit` | `/admin/about/@edit?alt=fr` |

yagebu:
I like this a lot - the two kind of paths (colon-separated "url paths" and file system paths) are a real annoyance in the admin UI JS code. I'm happy to give a hand with any Typescript questions/issues you might run into.

Some quick thoughts on this:

- We don't have to keep the "edit redirect" URLs the same if that makes it harder to do this, right? - they're only used internally from the pencil button so we could change them to, e.g., `/admin/redirect?path=...` to avoid a clash.
- I don't really like the `@` in the URLs - it's used in URIs to separate the username from the host and it looks out of place at the end of the URL. I also think it might, like the colon, come back to bite us at some point.
- Heads up that the next version of react-router that we'll probably want to switch to at some point only supports wildcards at the end of URLs: https://reactrouter.com/docs/en/v6/upgrading/v5#note-on-route-path-patterns

Since the record path is user-defined and might contain unexpected characters, I think that we should place it somewhere in the URL where it's easy to parse. I see two options for this: 1) at the end of the URL, so like `/admin/edit/path/to/record?alt=en` or 2) in the URL query, so like `/admin/edit?path=/path/to/record&alt=en`. I think 1) looks nicer.

dairiki:
> * I don't really like the `@` in the URLs - it's used in URIs to separate the username from the host and it looks out of place at the end of the URL. I also think it might, like the colon, come back to bite us at some point.

It's just an idea.  I took that from Pyramid's [traversal-based routing][traversal], where a leading `@@` in a path element stops traversal and forces interpretation as a _view name_.  `@`s are legal without restriction in the _path_ portion of URLs (and the user/host portion of the URL is always syntactically distinct from the path portion) so I don't think it can cause any real trouble, other than it would preclude the use of (leading) `@`s in Lektor db path components.

> * Heads up that the next version of react-router that we'll probably want to switch to at some point only supports wildcards at the end of URLs: https://reactrouter.com/docs/en/v6/upgrading/v5#note-on-route-path-patterns
> 
> Since the record path is user-defined and might contain unexpected characters, I think that we should place it somewhere in the URL where it's easy to parse. I see two options for this: 1) at the end of the URL, so like `/admin/edit/path/to/record?alt=en` or 2) in the URL query, so like `/admin/edit?path=/path/to/record&alt=en`. I think 1) looks nicer.

I like the idea of putting the db-path in the URL-path, but somehow adding the view name before the db-path in the URL path (while it does solve some problems) feels wrong, somehow.  I guess of those two choices, I favor #2, because it's more clear to a human reader just exactly how to parse the URL.

----

While we're at it: It seems the `/admin` URL prefix is hard-coded. I think (though I haven't tested) this will conflict with any (or some) attempts to create Lektor pages at or under `/admin`.  We should probably give some thought to changing the `/admin` prefix.

We could change the admin prefix to `/.admin`.  (I believe leading periods are already disallowed as Lektor page names.)  There may well be better alternatives.

[traversal]: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/traversal.html#a-description-of-the-traversal-algorithm

yagebu:
Merging of #977 caused quite a lot of merge conflicts for this one - however I've moved all creation of these admin URLs to their own modules (except for main.tsx as I've just realized), so the only frontend files that now need to be changed for this PR should be main.tsx, RecordComponent.tsx and the new use-go-to-admin-page.tsx.

> other than it would preclude the use of (leading) @s in Lektor db path components.

That's really my main concern with the @. 

> I like the idea of putting the db-path in the URL-path, but somehow adding the view name before the db-path in the URL path (while it does solve some problems) feels wrong, somehow. I guess of those two choices, I favor #2, because it's more clear to a human reader just exactly how to parse the URL.

Thinking about this again I've realized that putting it in the query string will mean that all slashes will be percent-encoded - that's quite ugly. I'm not If you dislike the view before the path, how about we go back to your original proposal? (and maybe avoid the "edit" clash by renaming the redirect URL route?). If those ever cause problems with the next version of react-router, "parsing" them by hand shouldn't be too hard (we just have to strip off the last path element).

dairiki:
> Merging of #977 caused quite a lot of merge conflicts for this one

No problem.  I did see that coming...

> > other than it would preclude the use of (leading) @s in Lektor db path components.
> 
> That's really my main concern with the @.

We could use a leading period instead.  That's [already disallowed][1] in db-paths.

[1]: https://github.com/lektor/lektor/blob/19486da795ba5dff7a8abf39302e3ec2c6429dd0/lektor/utils.py#L510

> > I like the idea of putting the db-path in the URL-path, but somehow adding the view name before the db-path in the URL path (while it does solve some problems) feels wrong, somehow. I guess of those two choices, I favor #2, because it's more clear to a human reader just exactly how to parse the URL.
> 
> Thinking about this again I've realized that putting it in the query string will mean that all slashes will be percent-encoded - that's quite ugly.


I've realized that we could (I'm pretty sure) do away with the redirector altogether. The [admin server][2], via [LektorInfo.resolve_artifact][3], already resolves URLs to db-paths (when possible), so it would be fairly straightforward to have the pencil directly link to the correct edit view URL.  (This also eliminates the ugliness of having the redirector possibly return a 404.  We can just not show the pencil if we can't resolve the db-path.)

[2]: https://github.com/lektor/lektor/blob/8a32396cd87a5a46ba7b9c6af7cfbd5a12ab2cd5/lektor/admin/modules/serve.py#L154
[3]: https://github.com/lektor/lektor/blob/8a32396cd87a5a46ba7b9c6af7cfbd5a12ab2cd5/lektor/admin/webui.py#L50

If we do eliminate the redirector, what about:
`/.admin/{edit,preview,delete,...}/path/to/record?alt=xx`
`/.admin/api/{pathinfo,recordinfo,...}`

Or, maybe even more clearly (this would also make Flask routing cleaner):
`/.admin/ui/{edit,...}/path/to/record?alt=xx`
`/.admin/api/{pathinfo,...}`

Or, maybe just?
`/.ui/{edit,...}/path/to/record?alt=xx`
`/.api/{pathinfo,...}`

yagebu:
> > other than it would preclude the use of (leading) @s in Lektor db path components.

>   That's really my main concern with the @.

Actually that's not really a problem if each URL has the view at the end - then something like `../page/@edit` and `../page/@edit/@edit` would be precise and easily distinguishable. However we don't really need the @ by that same logic...

> I've realized that we could (I'm pretty sure) do away with the redirector altogether.

:+1: 

Of the proposed options I like the first best. Having "admin" in the URL is good and I think the user-visible URL should be as concise as possible and it's good to have everything related to the admin under one base path.

dairiki:
I've rebased/refactored some more.

The redirector is gone.

URLs to the React app now look like:
`/admin/{edit,preview,delete,upload,add-child}/path/to/record?alt=de`

The API URLs are unchanged at:
`/admin/api/...`

I haven't moved `/admin` to `/.admin` yet, though I still think we probably should (to avoid conflicts if one creates a Lektor page named `admin`.)

lektor/admin/webui.py
```diff
@@ -11,9 +12,15 @@
 from lektor.builder import Builder
 from lektor.buildfailures import FailureController
 from lektor.db import Database
+from lektor.db import Record
 from lektor.reporter import CliReporter
 
 
+ResolveResult = namedtuple(
```
yagebu: We could use the annotation-aware NamedTuple from typing here

lektor/admin/static/js/main.tsx
```diff
@@ -12,63 +14,70 @@ import DeletePage from "./views/delete/DeletePage";
...
+      <Redirect to={adminPath("edit", "/", "_primary")} />
+    </Route>
+  );
+
+  return (
     <BrowserRouter>
-      <Main />
-    </BrowserRouter>,
-    dash
+      <Switch>
```
yagebu, line 22: This used to be built with `<Switch>` and so on and I've intentionally removed all that. I think it's good to be less reliant on all of react-router's stuff - which is why I've mostly trimmed it down to just one use of `useRouteMatch` on the last upgrade of react-router I think. AFAICT we don't really gain much by using `<Switch>` and `<Route>`s here - we have just one type of route and using these things will just complicate the upcoming major upgrade to v6.

You can just leave it as is, I'm planning some further refactorings here anyway (getting rid of RecordProps and using Context, getting rid of NavLink as well probably).

lektor/admin/webui.py
dairiki: We could, I guess, but those would be the only (python) annotations anywhere in the project.

Whenever someone gets the energy to attempt adding type annotations to Lektor, that's the time to convert from `namedtuple` to `NameTuple`.

Or do you want to start incrementally adding annotations on new code right now?

lektor/admin/static/js/main.tsx
dairiki: I've just refactored this again, adding stronger type checking, etc.  Take a look.  I really think it's cleaner now.

There has to be a switch on page name somewhere.  It might as well be in a `<Switch>`.  This eliminates to need to handle invalid URLs in multiple places.

(The stuff in RecordComponent.tsx could certainly be refactored.)

My new `Page` component (in main.tsx) is responsible for grabbing the record path and alt from the route match and query params.  If you want to add a context provider, that's probably where to hook it in.

dairiki:
Sorry for vacillating, but I'm once again coming to think that putting the db-path in a query arg rather than in the url-path is the way to go.

Advantages of putting db path in query params:

- It seems semantically more correct to me.  `/admin/edit?path=%2Fabout&alt=de` make it very clear that we are invoking (conceptually) an `edit(path="/about", alt="de")` operation.  `Path` and `alt` are both parameters to the (conceptual)  "edit" operation.  There's a funky asymmetry about passing one in the URL path and one in the query params.  It's not obvious to the naïve observer where the operation part of the URL ends and the db-path part of it starts.
- Correctly encoding esoteric db-paths to url-paths is not simple.  Db-paths can, in theory, contain unicode as well as URL-reserved characters like `'?'`. Neither `encodeURI` (which doesn't encode, e.g. `?`) nor `encodeURIComponent` (which encodes `/`s) do quite the right thing.  (I think the correct algorithm is the split the path on slashes, map the bits through `encodeURIComponent`, then join them again.)  If the db-path is put in the query args, `encodeURIComponent` just works.

Disadvantages of path in query params:
- It's ugly and hard to type with all the `"%2F"`s.  OTOH, when do we manually type admin URLs?

----

I've just pushed a commit to this PR that moves the db-path to the query params.  See if you like it.

lektor/admin/webui.py
yagebu: > Or do you want to start incrementally adding annotations on new code right now?

Yes, I'd be for incrementally adding annotations.

In this case here your choice of a namedtuple is already an upgrade in readability - adding annotations would be a further step IMHO and make this even easier to review (these attributes are probably all intended to be strings - but is one of them Optional?)

yagebu:
I'm happy to move it to the query arg. I think there's ways for both URL styles to be thought of as semantically correct, but the following is the killer argument for me - just like the colon-separated boths were a real hassle to deal with, we'll probably run in some annoying edge cases by putting them in the URL path as you point out.

>  Correctly encoding esoteric db-paths to url-paths is not simple. Db-paths can, in theory, contain unicode as well as URL-reserved characters like '?'. Neither encodeURI (which doesn't encode, e.g. ?) nor encodeURIComponent (which encodes /s) do quite the right thing. (I think the correct algorithm is the split the path on slashes, map the bits through encodeURIComponent, then join them again.) If the db-path is put in the query args, encodeURIComponent just works.

lektor/admin/static/js/main.tsx
yagebu: I like it

PR #977, opened 2021-12-30
Rewrite in hooks

### LINKED ISSUE #640: Strange scroll behaviour with flow elements on Firefox
A very annoying scroll behaviour happens with flow elements in Firefox:

Firefox

This is what it looks like in Chrome (Chromium in my case) - it's much better:

chrome

I don't want to use Chrome so it would be nice when it can be fixed soon ;)

### LINKED ISSUE #968: Admin GUI breaks if page contains an unknown flowblock type
To reproduce:
1. Fire up the server on the demo project in the tests subdirectory
```sh
cd tests/demo-project
lektor server --browse
```
2. Click the pencil to edit the top "Welcome" page
3. Try to add a _text_ flowblock
     (Nothing happens except there are two tracebacks in the javascript console.)

The cause seems to have something to do with the fact that the _Welcome_ page's [`contents.lr`](https://raw.githubusercontent.com/lektor/lektor/a31d8e57a5b2f4b090072527269c26a65202c736/tests/demo-project/content/contents.lr) contains two flowblocks, one of type `text`, and one of type `quote`.  There is no definition of the _quote_ flowblock in [`tests/demo-project/flowblocks`](https://github.com/lektor/lektor/tree/a31d8e57a5b2f4b090072527269c26a65202c736/tests/demo-project/flowblocks) — this appears to be what causes issues.  If one manually creates an appropriate definition of a _quote_ flowblock in `tests/demo-project/flowblocks/quote.ini`, then all works well.

yagebu:
This PR rewrites all remaining class components into function components using hooks, which allows us to easily address a couple of issues:

- All the components that fetch content are currently susceptible to potential data races - with the useEffect hook one can easily ensure that data returned in outdated fetch requests is ignored.
- Similarly, we've had problems with components not updating - now the eslint rules for React hooks ensure that all changes to variables used in a fetch re-trigger the effect hook.
- It made it easier to improve memoisation and hence the performance of the flow widget.

Overall, I find function components with hooks much easier to work with, they allow for better typing and separation of logic.

Fixes #640 
Fixes #968


dairiki:
+1

I have not looked carefully at your changes, but function-based components do seem more readable (once one manages to wrap one's head around the subtleties of hooks). And they appear to be the [anointed way](https://reactjs.org/docs/hooks-faq.html#should-i-use-hooks-classes-or-a-mix-of-both) of the future, as well.

PR #979, opened 2022-01-04
devserver: use daemon property instead of deprecated setDaemon

yagebu:
threading.setDaemon is deprecated since Python 3.10

### DIFF
--- a/lektor/devserver.py
+++ b/lektor/devserver.py
@@ -15,11 +15,6 @@ from lektor.utils import process_extra_flags
 from lektor.watcher import Watcher
 
 
-_os_alt_seps = list(
-    sep for sep in [os.path.sep, os.path.altsep] if sep not in (None, "/")
-)
-
-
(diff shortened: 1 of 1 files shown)


lektor/devserver.py
```diff
@@ -99,7 +94,7 @@ def browse():
         webbrowser.open("http://%s:%s" % addr)
 
     t = threading.Thread(target=browse)
-    t.setDaemon(True)
+    t.daemon = True
```
dairiki, line 97: One can pass a `daemon` kwarg to `Thread.__init__`. E.g.:
```py
t = threading.Thread(target=browse, daemon=True)
t.start()
```

or even
```py
threading.Thread(target=browse, daemon=True).start()
```

yagebu: Ah, good to know. In the other case the `BackgroundBuilder` class should probably already do that itself.

PR #981, opened 2022-01-05
admin: responsive layouts and mobile and various style fixes

yagebu:
Due to out use of Bootstrap for the admin UI, in theory we had a responsive layout all along in theory, however by using absolute positions in various places and not shipping the page with the proper `<meta>` tags, the desktop UI would render on mobile.

Also, there were quite a few unnecessary rules (to address issues that were fixed in Bootstrap v5 AFAICT) that also created style bugs on mobile or otherwise were removed.

### DIFF
--- a/lektor/admin/static/js/components/ServerStatus.tsx
+++ b/lektor/admin/static/js/components/ServerStatus.tsx
@@ -33,7 +33,7 @@ export default function ServerStatus(): JSX.Element | null {
     return null;
   }
   return (
-    <div className="server-down-panel">
+    <div className="interface-protector server-down-panel">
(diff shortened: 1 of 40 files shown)


dairiki:
Since the default is to serve to `localhost`, I'm not sure how often the admin UI will be used on mobile, but I gave this a quick test drive, and it seems to work fine.  It does make the admin UI usable on mobile.  (I.e. LGTM)

yagebu:
> I'm not sure how often the admin UI will be used on mobile

I host all Lektor sites (with admin) also on password-protected internet-accessible websites, so I (and the other users) often get the chance to use it on mobile ;)

PR #984, opened 2022-01-11
[jinja_debug] Add jinja debug extension when LEKTOR_DEV=1

nixjdm:
### Description of Changes

This enables the Jinja debug extension when the LEKTOR_DEV env var is set to `1` and `lektor server` is used. This can be useful during template development.

- [x] Link to corresponding documentation branch for [getlektor.com](https://github.com/lektor/lektor-website) https://github.com/lektor/lektor-website/tree/jinja_debug - only missing an active version note.

### DIFF
--- a/lektor/devserver.py
+++ b/lektor/devserver.py
@@ -115,6 +115,8 @@ def run_server(
     wz_as_main = os.environ.get("WERKZEUG_RUN_MAIN") == "true"
     in_main_process = not lektor_dev or wz_as_main
     extra_flags = process_extra_flags(extra_flags)
+    if lektor_dev:
+        env.jinja_env.add_extension("jinja2.ext.debug")
 
     if in_main_process:
         background_builder = BackgroundBuilder(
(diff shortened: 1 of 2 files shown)


lektor/devserver.py
```diff
@@ -115,6 +115,8 @@ def run_server(
     wz_as_main = os.environ.get("WERKZEUG_RUN_MAIN") == "true"
     in_main_process = not lektor_dev or wz_as_main
     extra_flags = process_extra_flags(extra_flags)
+    if lektor_dev:
+        env.jinja_env.add_extension("jinja2.ext.debug")
```
dairiki, line 119: What if a custom plugin is already adding the extension. Does it cause problems to add the same extension twice?
I suspect that it is not an issue, but it's worth checking to make sure if you haven't.

nixjdm: It shouldn't. The list of extensions is a `dict` and in that case the value associated with the key `'jinja2.ext.DebugExtension'` would just be replaced by the same, re-imported extension. For this pre-packaged extension there is no important side-effect of re-importing it that I see.

See https://github.com/pallets/jinja/blob/main/src/jinja2/environment.py#L113-L119 and https://github.com/pallets/jinja/blob/main/src/jinja2/environment.py#L373

PR #985, opened 2022-01-12
Test with mypy and work on adding type annotations

dairiki:
I've started playing with adding type annotations and testing with mypy.

I think it will be a while before this is merge-ready, but there's enough here to get mypy to run with meaningful results.

(Bugs have already been discovered, proving that this is probably a worthwhile exercise.)

(Run mypy with `tox -e mypy`.  This requires python>=3.8.)


yagebu:
Nice :)

One thought: I think we could drop support for Python 3.6 (EOL since December) before adding widespread type annotations, which would allow us to do `from __future__ import annotations` (postponed evaluation of annotations). That way we could some of the more modern typing features without negative impact on older Python versions, e.g. avoid a lot the imports from typing and use `list[str] | None` etc.

PR #987, opened 2022-01-13
Refactor the admin Flask app

dairiki:
While working on adding type annotations and getting mypy to run, one of the things I discovered is that the API views are not doing much validation on their input parameters.

This PR fixes that.

While working on that, I just got deeper and deeper...

This PR now pretty much totally refactors the Flask app.

It provides 100% test coverage of the Flask app, and fixes a number of buglets.

### Return 405s when appropriate (rather than 404s)

The default artifact serving route (`/<path:path>` in the serve blueprint) was handling GET requests for admin URLS that were PUT or POST-only.

### Resolve URLs to sources that have a trailing `/index.html`

Note if you browse to http://localhost:5000/index.html you don't get the edit pencil.  This is because the view in the serve blueprint was not resolving the URL to the source object for the root page.  It was just falling back to serving `index.html` from the output path.

### And a number of other buglets

- When serving URLs that can not be resolved to Lektor sources (e.g. sub-artifacts), handle
  directories correctly: add-slash redirect, and serve `index.html` or `index.htm`.
- When serving URLs that resolve to Assets, handle directories
  correctly: add-slash redirect.
- Move the logic for finding `index.html`/`index.htm` for `Directory` assets from the
  asset to the Flask view.
- When mangling HTML to add the edit button, correctly treat html tags as case-insensitive.
- Add edit button to build failure report pages
- Return 404 (not 200) when serving custom 404.html for missing paths
- Fix add-slash redirects when path has internal slashes. (#997)
- The /matchurl API endpoint should only match URLs that correspond to `Record`s.  This method is used by the admin frontend to follow navigation in its preview iframe.  Since the frontend only know how to deal with `Record`s (i.e pages and attachments), /matchurl should not match URLs corresponding to non-Records (e.g. assets and virtual source objects).


yagebu:
This looks great overall :)

Just one thing I'm not totally sure about: This adds two new dependencies (marshmallow and marshmallow-dataclass). marshmallow seems to be quite widely used and available but the latter of which seems to be pretty obscure (https://repology.org/project/python:marshmallow-dataclass/versions), so in most cases someone packaging Lektor will also have to add a package for marshmallow-dataclass. Is that worth it or could this validation also be done with "plain" marshmallow schemas (with slightly more effort)? (or some other more established validation library?)

dairiki:
> [marshmallow-dataclass] seems to be pretty obscure (https://repology.org/project/python:marshmallow-dataclass/versions), so in most cases someone packaging Lektor will also have to add a package for marshmallow-dataclass.

Good point.
(And I was unaware of repology.  It looks like a valuable resource.)

> Is that worth it or could this validation also be done with "plain" marshmallow schemas (with slightly more effort)?

It can, of course, be done with plain marshmallow, but doing so results in more verbose, non-DRY code.  It's not so much the extra effort that is the issue, but rather the loss of readability and maintainability.

Contrast:
```py
@dataclass
class _NewRecordParams:
    id: str
    model: Optional[str]
    data: Dict[str, Optional[str]]
    path: _PathType
    alt: _AltType = PRIMARY_ALT

schema_class = marshmallow_dataclass.class_schema(_NewRecordParams)
schema = schema_class(unknown=marshmallow.EXCLUDE)
validated = schema..load(data)
```
with
```py
@dataclass
class _NewRecordParams:
    id: str
    model: Optional[str]
    data: Dict[str, Optional[str]]
    path: _PathType
    alt: _AltType = PRIMARY_ALT

class NewRecordSchema(marshmallow.Schema):
    id = fields.Str()
    model = fields.Str(missing=None)
    data = fields.Dict(fields.Str(), fields.Str(missing=None))
    path = fields.Str()
    alt = fields.Str(validate=_is_valid_alt)

    @marshmallow.post_load
    def make_record(self, data, **kwargs):
        return _NewRecordParams(**data)

schema = NewRecordSchema(unknown=marshmallow.EXCLUDE)
validated = schema..load(data)
```

In the latter example, the programmer is now responsible for ensuring that the schema is kept in sync with the type annotations in the dataclass.

> (or some other more established validation library?)

If there is some other more popular library that will do the equivalent thing, I am unaware of it (though I'm certainly willing to use it.)

yagebu:
> If there is some other more popular library that will do the equivalent thing, I am unaware of it (though I'm certainly willing to use it.)

pydantic seems to be even more popular than cattrs and marshmallow currently (https://pypistats.org/packages/pydantic) and is also more widely available on distros than marshmallow-dataclass (similarly to cattrs and marshmallow itself). It also has support of type annotations as its main selling point and can validate dataclasses (from types).

I tried its `validate_arguments` decorator (beta, https://pydantic-docs.helpmanual.io/usage/validation_decorator/) as an even shorter solution for a similar use case, which  worked well - but I ended up not using it and writing some validation logic by hand because I wanted to avoid an extra dependency and like the design of the (sadly very new and still obscure) apischema (https://wyfo.github.io/apischema/0.17/) even better...

dairiki:
> pydantic seems to be even more popular than cattrs and marshmallow

Thank you!  I will attempt to refactor to use pydantic.

dairiki:
Now using pydantic.

> I [...] like the design of the (sadly very new and still obscure) apischema (https://wyfo.github.io/apischema/0.17/) even better...

Agree. Apischema and marshmallow-dataclass both have a cleaner API than pydantic.

It's a bit hokey how one must build the deserialization/validation into any custom types that one uses with pydantic.  There seem to be some weird little bugs, and (since the reason we're doing this is to make packagers lives easier) a lot of the packaged versions out there seem pretty old. (My Ubuntu 20.04 packages pydantic 1.2.)

Another option might be to vendor-branch marshmallow-dataclass.  Since it relies on marshmallow to do the actual validation/de/serialization, the package is relatively lightweight. Marshmallow-dataclass is under 1000 lines (30kb) of python code. Both apischema and pydantic are an order of magnitude larger. (As is Lektor.)  Marshmallow-dataclass is distributed under the MIT license.

yagebu:
> (since the reason we're doing this is to make packagers lives easier) a lot of the packaged versions out there seem pretty old. (My Ubuntu 20.04 packages pydantic 1.2.)

I don't think that's really an issue - existing stable distros will keep shipping old versions of Lektor anyway and packagers working on the distros currently in development should have a more recent version of pydantic. I'd be very strongly against vendoring a dependency (and it would have to be unvendored by packagers in some cases anyway).

> It's a bit hokey how one must build the deserialization/validation into any custom types that one uses with pydantic. There seem to be some weird little bugs, 

Thanks for trying it out, but from your description it sounds like adding a dependency to marshmallow-dataclass might be worth it indeed. Your call :)

dairiki:
@jcharaoui As a Debian maintainer, do you have any input as to how much of a pain it would be if we were to add a new pure-python requirement (one that is available and stable in PyPI, but not yet packaged for Debian)?  (See the previous few comments, starting from [this one](https://github.com/lektor/lektor/pull/987#issuecomment-1044966507), for context.)

jcharaoui:
@dairiki I can package and maintain an additional Python dependency, that's fine. I've done it already for some of the other requirements for Lektor. Are we talking about marshmallow-dataclass ?

Thanks for asking, much appreciated!!

dairiki:
> @dairiki I can package and maintain an additional Python dependency, that's fine. I've done it already for some of the other requirements for Lektor. Are we talking about marshmallow-dataclass ?

@jcharaoui Great!  Thank you for the quick reply.  Yes — `marshmallow-dataclass`.

PR #992, opened 2022-01-25
Resolve links in Markdown text. Support mistune 2.x

### LINKED ISSUE #313: Attached image and alternatives

### LINKED ISSUE #950: BUG: attaching images are stored with a broken URL

### LINKED ISSUE #966: URL resolution abiguities in Lektor
## The current state of URL resolution

URL resolution in Lektor generally runs through [`SourceObject.url_to`](https://github.com/lektor/lektor/blob/a31d8e57a5b2f4b090072527269c26a65202c736/lektor/sourceobj.py#L90).  That method has a somewhat confusing, sometimes surprising behavior in that when resolving paths, sometimes the `path` argument is treated as a Lektor db path, while at other times it is treated as a URL path.

Though sometimes the db path to a record is the same as the URL path, in general that is not the case.  The difficulty in specifying which interpretation `url_to` should take, and determining which interpretation was actually made can be quite confusing.

### URL resolution in templates (and from python code)

In a template, if one does:
```jinja2
<img src="{{ 'image.jpg' | url_to }}">
```
what happens is:
- An attempt is made to resolve the path (`"image.jpg"`) via the Lektor db.
- If the resolution is successful, the URL for the resolved db record is returned. Note that while the URL for the record is sometimes identical to the db path to the record, that is not always the case. (E.g. if a custom slug as been set for the record.)
- If the db lookup fails, the path is interpreted as a URL path (relative to the base URL implied by the currently-being-rendered page.)

The resolution through the db is useful in several cases:
- For attachments when alternatives are in use. In this case the attachments are only emitted for the primary alternative, so on non-primary alternative pages, the relative URL to the attachment is not the same as the relative db path. 
- If there is a custom `_slug` set for the record, (or a custom `children.slug_format` set in the parent's data model), again the db path is not the same as the URL path. 

In these cases, `url_to` generally does the right thing.

To disable the attempt at resolving the path via the Lektor db, one may prefix the path with a `"!"`.  That forces the path to be treated as a URL path.

### URL resolution for Markdown text

For markdown links and images, [current][1] [code][2] _always_ prefixes the path with `"!"`, thus disabling db path resolution.  This forces authors to always know and use valid URL paths rather than db paths when authoring markdown text.

This causes problems with links to attachments when alternatives are in use.  (See, e.g. #950, #313, #470, and #88.)  Similar issues arise when custom slugs are in use.

[1]: https://github.com/lektor/lektor/blob/a31d8e57a5b2f4b090072527269c26a65202c736/lektor/markdown.py#L20
[2]: https://github.com/lektor/lektor/blob/a31d8e57a5b2f4b090072527269c26a65202c736/lektor/markdown.py#L31

----

## Issues

The obvious issue, as evidenced by all the bug reports having to do with URL generation for attachments when alternatives are in use, is that it would be nice to have some way to resolve links in markdown as Lektor db paths.

Another issue is the way `url_to` even when attempting db resolution, silently falls back to treating the path as a URL path if db resolution fails.  That's just a set-up for confusion.

----

## Possible Solutions

Spitballing...

### 1. Allow db resolution in markdown.

The simplest solution is just to do away with the `"!"` prefixing in our markdown renderer.

This may break a few existing sites, but I _think_ any such breakage will be minimal.  For those paths which now will be successfully resolved through the Lektor db, generally, either the URL will be the same as the db path (so no change), or db resolution was probably what the writer originally intended anyway.

This does not solve the ambiguity of `url_to` silently treating the path as a URL path with db path resolution fails.

### 2. Add a new argument to url_to to control resolution

We could add a new argument to the `url_to` functions to control path resolution.  E.g.
```py
def url_to(path, alt=none, absolute=None, external=None, resolve=None):
    ...
```
With `resolve=None` (the default), the behavior would be identical to what it is now: attempt db resolution, fall back to interpreting path as a URL path.

Setting `resolve=True` would force db path resolution.  `Url_to` would return `None` (or maybe `Undefined`) if path resolution fails.

Setting `resolve=False` would be equivalent to prefixing the path with `"!"` (but is clearer and less goofy.)

This fixes the resolution ambiguity, allowing the author a way to force db path resolution if that's what they intended.

As there is no way to directly set the optional parameters to `url_to` from markdown links, this change, by itself wouldn't fix the problem that there is no way to resolve markdown links via the Lektor db.

### 3. Add a new custom URL scheme to trigger db resolution

We could make it so that passing something like `path="lektor:image.jpg"` to `url_to` would force interpretation of the path as a db path.  (Equivalent to `resolve=True` from the previous idea.)

----

## My current preference

At the moment, my preference would be to implement solutions #1 and #3 listed above.

So: remove the `"!"` prefixing from our markdown `ImprovedRenderer`.  This has a slight potential to break existing sites, but the fix is easy: just add the `!` explicitly (like you would do in a template) if that's really what you wanted.

Bare paths (with no scheme) would be treated in the current DWIM fashion: try to resolve via db, but fall back to treating the path as a URL path if that resolution fails.

As it is now, paths could be prefixed with `!` to prevent the db resolution.

Additionally, they could be prefixed with `lektor:` to force db resolution (returning an error if resolution fails.)

----

## Questions for y'all

What do you think?

Is it going to give anyone gas if we remove the `"!"` prefixing from the markdown URL resolution?

dairiki:
Here's another giant PR.  This one refactors `lektor.markdown` with two main goals:
- Clean up and fix a number of bugs with link and image resolution in Markdown text.
- Support using the latest version of `mistune` (==2.0.2), while also maintaining support for the previous major version of mistune (0.9.4).

### Issue(s) Resolved

Fixes #313
Fixes #966
Fixes #950
Supercedes #944
Supercedes #685
Supercedes #470

### Related Issues / Links

Doc PR: lektor/lektor-website#338

### Description of Changes

----

### Link and image URL resolution via the Lektor DB

Previously any URLs in Markdown text which were not external links were interpreted as URL paths relative to either the Lektor site root (if the path starts with a `/`) or the page containing the markdown text (if the path does not start with a `/`).  As discussed in #966, this causes issues, e.g.:
- when linking to attachments, when alternatives are in use
- when linking to pages or attachments that use a custom slug

This PR changes the default behavior so that an attempt is made to resolve links as Lektor db paths to a source object.  If that resolution is successful, then the URL to that source object is used.  If that Lektor db lookup fails, then fall back to interpreting the path as a URL path (as is done now.)

Note that this new behavior is exactly the same as is currently implemented by Lektor’s `url` jinja filter.

Since, in many cases, resolution via the DB produces the same result as just interpreting the path as a URL path, while this is a backward-incompatible change, most of the time it will not make a difference.  And when it does, I think, that most of the time, the new behavior is what was probably wanted in the first place.

#### New `resolve_links` markdown field config option

This PR adds a new field configuration option for `markdown` fields.  This option can take one of three values:
- `resolve_links = never` <br>Never resolve links to Lektor source objects.  This is the old (current) behavior, equivalent to prefixing all paths with '!'.
- `resolve_links = when-possible`<br> This is the new default behavior.  When possible, links are resolved to Lektor source objects, then the URL to those source objects is used.  When the resolution to source object fails, the links are interpreted as URL paths.
- `resolve_links = always`<br>Resolve links to Lektor source objects, then use the URL of those source objects. If that resolution fails, an error is raised.

This option may be set in the field definition section of a model.ini file.  E.g. to restore the legacy behavior to a specific body field of your pages, you could put in `models/page.ini`:
```ini=page.ini
...
[fields.body]
label = Body
type = markdown
resolve_links = never
...
```

----

### Cleanup and of `SourceObject.url_to`.

This PR refactors `SourceObject.url_to`.  I think the new code is much more readable than the old version.

#### Change of behavior with `absolute=True`

It does make one behavior change.  Previously if `absolute=True` was passed, the `SourceObject.url_to` would return an "absolute" URL path (one which starts with a `/`), but that path was relative to the site root configured in the project file (if any).  The PR changes that so that a real absolute path is returned (relative to the root of the webserver).  E.g., if in the project file a site URL was set:
```ini
[project]
url = http://example.org/some/path/
```
`SourceObject.url_to("/", absolute=True)` used to return `"/"`.  This PR changes things so that it returns `"/some/path/"`.

It can be argued that the previous behavior was a bug.  Consider writing — not unreasonably — in a template:
```html+jinja
See <a href="{{ this.get('subpage').url_to(absolute=True) }}">this sub-page</a>.
```
If there is a `[project]url` set with a non-empty path, that link will be a broken one under the current behavior.

#### Addition of `resolve` and `strict_resolve` parameters

This PR also adds two new parameters to `SourceObject.url_to`: `resolve`, and `strict_resolve`.
The same new parameters are also added to `Context.url_to` and to the `url` jinja filter.

Passing `resolve=False` will prevent any attempt to resolve the path as a db path.  This is equivalent to what happens currently if one prefixes the path with `'!'`.

Passing `strict_resolve=True` will cause an error to be raised if the path can not be resolved as a db path.

#### Support for `lektor:` URL scheme

<del>As described in #966, `SourceObject.url_to` is modified to recognize the new `lektor:` URL scheme.  Prefixing a path with `lektor:` is now essentially equivalent to passing the new `strict_resolve=True` parameter (see above).</del>

With the addition of the `strict_resolve` parameter, the addition of the `lektor:` URL scheme is probably overkill.  At this point I <del>plan on removing it from this PR before merging</del> <ins>have removed it from this PR</ins>.

----

### Support for mistune 2.x

This also adds support for the latest version of mistune.

Note that it is unlikely that existing Lektor plugins that further customize Lektor's Markdown rendering will “just work” if running under the new version of mistune.  (Most such plugins work by subclassing the mistune Renderer class.  The API to that class has changed in the latest version of mistune.  The arguments to the constructor of the mistune parser have changed as well.)  

Lektor plugins that work now should continue to work if mistune is pinned to the old version (`mistune<2`.)

(A little more thought should probably be given to the issue of making life under multiple versions of mistune easy for plugins.)


dairiki:
@yagebu @xlotlu (and any other interested parties):

I would like to merge this PR soon.  However, since it does introduce a backward-incompatibility — namely that links in markdown text will be resolved via the Lektor DB rather than interpreted only as URLs (see the main PR description, above) — I want to double-check that there are no strong objections before doing so.

Also, would appreciate opinions on whether this is a significant enough change to warrant a major version bump (to 4.x) rather than just a minor version bump (to 3.4.x).

yagebu (approved):
LGTM :)

I'd be happy to ship this with 3.4, since this is, where incompatible, probably closer to the expected/intended behaviour (as evidenced by the long chain of issues and PRs that this PR would finally close :))

PR #995, opened 2022-01-31
GithubPagesPublisher refactor

### LINKED ISSUE #978: Github Pages: Change default branch from "master" -> "main"
The hard-coded default branch name for pushing to github pages (github.io) is `master`. This doesn't work out of the box anymore, as both github and git itself default to `main`. The code in question is in [publisher.py:609](https://github.com/lektor/lektor/blob/0a4bd3b0bca1033c406ffa898fef71b9210303ae/lektor/publisher.py#L609).

Would it be ok to simply change it in future versions? Would a PR be welcome then?

If not, would it be possible to at least give the option to configure the branch in `.lektorproject`? I could try a PR for this, if it were welcome.

dairiki:
What started out as work to support explicitly specifying a branch to push to (#978) has turned into this.

### Issue(s) Resolved

Fixes #978

### Related Issues / Links

Associated doc PR: lektor/lektor-website#340

### Description of Changes

- Add support for specifying which branch to push to by specifying a `branch` query param on the target URL.
- Issues a warning when defaulting to pushing to the `master` branch for a _user_ or _organization_ pages site.  (Recently created _user_ or _organization_ pages repositories default to using the `main` branch.)
- [Optimization] Do a shallow fetch when fetching the current target branch from upstream.
- [Optimization] Avoid copying the HTML output files to a new scratch directory.
- Refactor/cleanup of some code in `lektor.publisher`.

## Question

Is it worth adding an option that would prevent keeping any git history of the deployed site?

If one doesn't care about preserving the history of the deployed gh-pages site, one can just force-push the new
content to the target branch.  This avoids having to fetch the existing branch from upstream, saving network bandwidth.

In general, I'm not sure how useful it is to keep a history of the deployed site anyway.

_Of pertinence_, I just noticed that [our docs](https://www.getlektor.com/docs/deployment/ghpages/) currently say: "_The way this is implemented in Lektor currently is that Lektor will force-push a website into a repository of choice._"
This appears _not to be_ what our current code does.  Current code appears to alway creates a new commit (with the previous branch head as its parent) preserving all previously published version of the site's HTML in the git history.  (No force push involved.)
Later on the same page, our docs say: "_The way this deployment support works is that it commits a new commit into a temporary location and pushes it into the gh-pages or master branch depending on the name of the repository._" (which is not quite the same thing.)

### DIFF
--- a/lektor/compat.py
+++ b/lektor/compat.py
@@ -1,0 +1,53 @@
+import os
+import stat
+import sys
+import tempfile
+from functools import partial
+from itertools import chain
+
+
+def _ensure_tree_writeable(path: str) -> None:
+    """Attempt to ensure that all files in the tree rooted at path are writeable."""
+    dirscans = []
+
+    def fix_mode(path, statfunc):
+        try:
+            # paranoia regarding symlink attacks
+            current_mode = statfunc(follow_symlinks=False).st_mode
+            if not stat.S_ISLNK(current_mode):
+                isdir = stat.S_ISDIR(current_mode)
+                fixed_mode = current_mode | (0o700 if isdir else 0o200)
+                if current_mode != fixed_mode:
+                    os.chmod(path, fixed_mode)
+                if isdir:
+                    dirscans.append(os.scandir(path))
+        except FileNotFoundError:
+            pass
+
+    fix_mode(path, partial(os.stat, path))
+    for entry in chain.from_iterable(dirscans):
+        fix_mode(entry.path, entry.stat)
+
+
+class FixedTemporaryDirectory(tempfile.TemporaryDirectory):
+    """A version of tempfile.TemporaryDirectory that works if dir contains read-only files.
+
+    On python < 3.8 under Windows, if any read-only files are created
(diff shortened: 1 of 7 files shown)


yagebu:
For ghpages, I think it's preferable to force-push an orphaned commit instead of having some history of deployed pages. Since that would be consistent with the current documentation (but not the implementation), I think changing it to work like that would be fine.

dairiki:
> For ghpages, I think it's preferable to force-push an orphaned commit instead of having some history of deployed pages. Since that would be consistent with the current documentation (but not the implementation), I think changing it to work like that would be fine.

I tend to agree that force-push should be the preferred behavior, however, I worry about changing the default behavior without warning (other than in the changelog).  In the rare cases where someone does want to preserve the history of the deployed HTML, changing to force-push will delete any such accumulated history.

The safer way to make the change would be to add a `preserve_history=false` option that would enable the force-push behavior. The default behavior would be left unchanged, but a warning would be issued that the default behavior will be changed in future and that users should add an explicit `preserver_history=true` if they want to keep the non-force-push behavior.

Is it worth the effort to do that?

PR #1000, opened 2022-02-15
Fixes for #998 — tests fail when git is not installed

### LINKED ISSUE #998: Some tests fail when git is not installed
Hello,

While working on the Debian package I noticed that test failures are encountered in `test_devcli.py` when `git` is not available/installed.

I think this is caused by `get_default_author_email()` returning `None`. When I change it to an empty string, the tests succeed. 

Thanks.

dairiki:
### Issue(s) Resolved

Fixes #998

#### Other clean-ups and fixes

Handle potential `KeyError` from `pwd.getpwuid` in `lektor.quickstart.get_default_author`.

If _git_ is not available, or if git's `user.email` is unset, fall back to checking `$EMAIL` in `get_default_author_email`.

`Lektor.quickstart.project_quickstart` had logic intended to generate the new project in the current directory if run in an empty directory.  This logic was however [broken][2] in such a way that it never could be activated.  I've just deleted it — we've survived this long without that functionality, so hopefully, no one will miss it.

### Description of Changes

Added a `noutils` tox _[factor][]_ to run tests with an empty `$PATH` in order to check that tests pass without external utilities (e.g. `git`, `imagemagick`) installed.

[factor]: https://tox.wiki/en/latest/config.html#generating-environments-conditional-settings

A couple of the tests in `tests/test_devcli.py` were making the brittle assumption that `lektor.quickstart.get_default_author_email` would return a non-`None` value.  Here we monkey-patch `get_default_author_email` so return a known value for these tests.

The test for `get_default_author_email` in `tests/test_quickstart.py` was also incorrectly making that assumption.  Note that `get_default_author_email` should be allowed to return `None` in the case that no default can be reasonably guessed.  The only use for the return value is to feed as the default to `click.prompt` — for that use, `None` is an allowed value, used to indicate that no default exists.

I've deleted a couple of unnecessary conditional conversions from `bytes` to `str` in `lektor.quickstart`.  As of python 3, at least, there is no possibility of the given values being `bytes`.

[2]: https://github.com/lektor/lektor/blob/2f701037cfc5f913094aa533a7a42abfff7c8296/lektor/quickstart.py#L210

### DIFF
--- a/lektor/quickstart.py
+++ b/lektor/quickstart.py
@@ -1,19 +1,24 @@
+import getpass
 import os
 import re
 import shutil
-import subprocess
-import tempfile
-import uuid
 from contextlib import contextmanager
 from datetime import datetime
 from functools import partial
+from importlib import import_module
+from subprocess import PIPE
+from subprocess import run
+from tempfile import TemporaryDirectory
+from typing import Optional
 
 import click
 from jinja2 import Environment
 from jinja2 import PackageLoader
 
-from lektor.utils import fs_enc
+from lektor.utils import locate_executable
 from lektor.utils import slugify
+
+pwd = import_module("pwd") if os.name != "nt" else None
 
 
 _var_re = re.compile(r"@([^@]+)@")
@@ -88,23 +93,15 @@ class Generator:
             except OSError as e:
                 self.abort("Bad target folder: %s" % e)
 
-        scratch = os.path.join(tempfile.gettempdir(), uuid.uuid4().hex)
-        os.makedirs(scratch)
-        try:
+        with TemporaryDirectory() as scratch:
             yield scratch
-        except Exception:
-            shutil.rmtree(scratch)
-            raise
-        else:
+
             # Use shutil.move here in case we move across a file system
             # boundary.
             for filename in os.listdir(scratch):
-                if not isinstance(path, str):
-                    filename = filename.decode(fs_enc)
                 shutil.move(
                     os.path.join(scratch, filename), os.path.join(path, filename)
                 )
-            os.rmdir(scratch)
 
     @staticmethod
     def expand_filename(base, ctx, template_filename):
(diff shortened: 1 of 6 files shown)


lektor/quickstart.py
```diff
@@ -1,20 +1,25 @@
...
 
 import click
 from jinja2 import Environment
 from jinja2 import PackageLoader
 
-from lektor.utils import fs_enc
+from lektor.utils import locate_executable
 from lektor.utils import slugify
 
+pwd = import_module("pwd") if os.name != "nt" else None
```
yagebu, line 21: IMHO, the more common and simpler idiom to do this kind of conditional import is `try: ... except ImportError: ...`. Or does that not work in this case?

yagebu (approved):
LGTM, of course :) just one small comment

lektor/quickstart.py
dairiki: I resorted to that syntax to make things clear for mypy.

Once mypy sees an `import` statement — whether in a conditional context or not —- it seems to treat that as _final_.

E.g.
```py
try:
    import pwd
except ImportError:
    pwd = None # Error: Incompatible types in assignment (expression has type "None", variable has type Module)
```

(The same error results if one does a conditional `if os.name != "nt":` around the import rather than the try...except.)

I also tried
```py
pwd: Optional[ModuleType] = None
if os.name != "nt":
    import pwd # Error: Name "pwd" already defined
```

----

Of note, because `sys.platform` is a mypy compile-time constant (as opposed to `os.name` which is not) this does work without error from mypy, since mypy will only "see" one leg or the other of the conditional:
```py
if sys.platform not in ("win32", "cygwin"):  # does cygwin have pwd?
    import pwd
else:
    pwd = None
```
However, I'm not sure that the `sys.platform` test is exactly what we want here.  (In any case Lektor seems to use `os.name` in favor of `sys.platform` for windows detection.) 

----

Anyhow, that's why I did that.  I'm open to alternatives.

yagebu: Thanks for the explanation and trying all these possible workarounds :)

Maybe this gets fixed in mypy at some point https://github.com/python/mypy/issues/1297 (there's a possible `try ... except ... else ...` workaround suggested in that thread but I don't think that that's much prettier)

PR #1002, opened 2022-02-18
builder: fix comparison to sqlite version

yagebu:
We want to add 'without rowid' to some tables to all sqlite versions
supporting it, however this code was doing a naive string comparison of
the versions, which would fail since sqlite3's minor version is in the
double digits and "8" <= "37" is False.

### DIFF
--- a/lektor/builder.py
+++ b/lektor/builder.py
@@ -24,7 +24,7 @@ from lektor.utils import prune_file_and_folder
 
 
 def create_tables(con):
-    can_disable_rowid = ("3", "8") <= tuple(sqlite3.sqlite_version.split("."))
+    can_disable_rowid = (3, 8, 2) <= sqlite3.sqlite_version_info
     if can_disable_rowid:
         without_rowid = "without rowid"
     else:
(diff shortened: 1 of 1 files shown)


dairiki:
Sqlite 3.8.2 was released in December, 2013.  I wonder whether we need to support versions before that at all at this point.

yagebu:
I considered the same thing since I'm always happy to drop ancient code/deps but I'm still a bit torn.

For this kind of system library, there's a case to be made to support an ancient version. Consider a CentOS 7 (just as an example of a super LTS-distro) user - for pure Python dependencies they (or rather pip(x)) can easily install newer versions and even for Python itself, there's additional repos where they could get a newer version (than 3.6.8 - https://repology.org/project/python/versions). However even that newer Python version is probably compiled against the system sqlite (which is on 3.7.17 - https://repology.org/project/sqlite/versions). I think this sort of constellation might be among the reasons why Python 3.10, the first to specify a minimum sqlite version (https://docs.python.org/3/library/sqlite3.html), still supports sqlite 3.7.15.

PR #1003, opened 2022-02-18
move admin frontend to its own top-level dir

yagebu:
This makes for a better separation of the sources for the admin frontend and the lektor Python package (which only gets the "compiled" frontend). Simplifies excluding all the unwanted frontend source files from the Python distribution package(s).

"Follow-up" to #986, will update the MANIFEST.in once that PR is merged.


dairiki:
From a quick look, LGTM.

PR #1004, opened 2022-02-24
drop support for Python 3.6

yagebu:
Python 3.6 has reached EOL two months ago and Python 3.7 provides a
couple of niceties. In particular, the following are or could be of use
in Lektor:

- dataclasses
- importlib.resources
- postponed evalution of type annotations
- better encoding defaults
- big performance improvements related to typing
- all dicts are now ordered (not just as an implementation detail of
  CPython, but also as specified by the language)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,16 +80,12 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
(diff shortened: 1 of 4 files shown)


dairiki:
LGTM

dairiki:
Before this is merged, we should probably make a release, since there are currently a bunch of accumulated fixes/changes in `master`.

Once this is merged, I'll freshen #985 and work on integrating mypy testing into our CI tests.

PR #959, opened 2021-12-13
Problems with `Page.source_filename`, `Page.iter_source_filenames` with alternatives are enabled

dairiki:
In playing around to look at another issue, I noticed that when building the example project, the blog pages are always rebuilt, even if they have not been changed.

I.e.
```sh
$ cd example
$ lektor build
Started build
U de/index.html
[... lots of other pages ...]
U static/css/nix.css
Finished build in 1.07 sec
Started prune
Finished prune in 0.00 sec
```

Then, do it again.  Since no source has been changed, nothing should be rebuilt.  But:
```sh
$ lektor build
Started build
U de/blog/blog-post-5/index.html
U de/blog/blog-post-4/index.html
U de/blog/blog-post-3/index.html
U de/blog/blog-post-2/index.html
U de/blog/first-post/index.html
U blog/blog-post-5/index.html
U blog/blog-post-4/index.html
U blog/blog-post-3/index.html
U blog/blog-post-2/index.html
U blog/first-post/index.html
U fr/blog/blog-post-5/index.html
U fr/blog/blog-post-4/index.html
U fr/blog/blog-post-3/index.html
U fr/blog/blog-post-2/index.html
U fr/blog/first-post/index.html
Finished build in 0.32 sec
Started prune
Finished prune in 0.01 sec
```

The blog posts are always rebuilt.

### Issue(s) Resolved

I believe the root cause of this is that `Page.source_filename` is, in some cases, returning bogus values — specifically
the name of non-existing files.  E.g. if only a `contents.lr` exists, `.source_filename` still returns `contents+de.lr` for the `alt="de"` version of the page.  Since this is used for dependency tracking, things break if this points to a non-existing file.

Looking further, I think that `SourceObject.source_filename` is, or should be deprecated in favor of `SourceObject.iter_source_filenames`.  Again, since this is used for dependency tracking, we usually (always?) care about all the source files, not just the "primary" one.

### Related Issues / Links

Not sure.

### Description of Changes

- [ ] Wrote at least one-line docstrings (for any new functions)
- [x] Added unit test(s) covering the changes (if testable)
- [ ] Link to corresponding documentation pull request for [getlektor.com](https://github.com/lektor/lektor-website)

PR #1007, opened 2022-03-01
Fix dependency tracking on VirtualSourceObjects

dairiki:
I think I finally sussed out what was causing the unnecessary rebuilds described in #959.  (And it was not exactly was I was guessing in that PR.)

(WIP - more later.)

#### The `alt` of virtual source dependencies is not being recorded in the buildstate database

The `artifacts` table in the buildstate database tracks artifacts dependencies on two types of sources:
- source files
- VirtualSourceObjects

It identifies those sources with the string-valued column `path`.

Source files are tracked by setting `path` to the file-system path of the source file.
Currently, virtual sources are tracked by setting `path` to `VirtualSourceObject.path` — i.e. the Lektor db-path of the virtual source.

The issue is that the path is not the full identity key for a VirtualSourceObject — virtual sources vary with `alt` as well as `path`.

As a concrete example, what happens in the issues described in #959 is (e.g.), when building the German page for the first blog post (artifact = `/de/blog/first-post/index.html`), [`post.get_siblings()`][2] is called which registers a dependency on the `Siblings` virtual source object attached to the post.  Since the post has `alt="de"`, this is the virtual source that may be obtained by calling `pad.get("/blog/first-post@siblings", alt="de")`.  Currently, however, the dependency gets recorded only by its path (`/blog/first-post@siblings`).   When the builder, checking for changes that would necessitate a rebuild, goes to check the current state of the virtual source, it fetches [`pad.get("/blog/first-post@siblings")`][1] with no alt specified.  This is a different object — instead of `alt="de"` it has the default alt (`en`, or `_primary` before #958) — so it always compares as changed, forcing an unconditional rebuild.

[1]: https://github.com/lektor/lektor/blob/1999510ea5ba389ba97d8c99417bb336fc381320/lektor/builder.py#L160
[2]: https://github.com/lektor/lektor/blob/1999510ea5ba389ba97d8c99417bb336fc381320/example/templates/macros/blog.html#L30

#### Siblings.get_checksum is being overly-pessimistic

The state of a page's `Siblings` only depends on the _identity_ of the siblings.  As long as the _next_ sibling of `/blog/first-post` is `/blog/blog-post-2` and the _prev_ sibling is `None`, then the `/blog/first-post@siblings` virtual source object can be considered unchanged, even if the content of `/blog/blog-post-2` changes.  (If, while building an artifact, data from that next page is accessed, then that next page is/should be recorded as a dependency, ensuring the artifact gets rebuild if the next page is touched.)

Currently, the [`Siblings.get_checksum`][3] and [`Siblings.get_mtime`][4] compute values dependent on the source files for the siblings.  That is overly pessimistic.

[3]: https://github.com/lektor/lektor/blob/1999510ea5ba389ba97d8c99417bb336fc381320/lektor/db.py#L549
[4]: https://github.com/lektor/lektor/blob/1999510ea5ba389ba97d8c99417bb336fc381320/lektor/db.py#L545

#### Record.\_\_getitem\_\_ should record a dependency on record

I think the reason Siblings.get_checksum (tries to) depend on the contents of the siblings is because this is not currently being done.

I have now implemented this.  In testing the build time of the getlektor.com website, the change does not seem to affect build time in any noticeable way.

**However**, fixing this has made every page depend on its complete lineage (its parent, grandparent, ...)
Strictly speaking, this is correct.  The lineage — e.g. by setting custom slugs — can affect the URL of a page.  Almost every page uses the `url` filter in its template.  By default, `url` returns a relative URL.  Computing a relative URL depends on the URL of the page being built.
I'm not sure if this will be a big deal in terms of forcing more rebuilds that (most of the time) aren't necessary.

[5]: https://github.com/lektor/lektor/blob/1999510ea5ba389ba97d8c99417bb336fc381320/lektor/db.py#L1532

Supercedes #959

### Description of Changes

This PR munges things in a minimal way so as to store both the `alt` and the `path` of `VirtualSourceObjects` in the `artifacts.path` column.

It also cleans up the checksum computation for the `Siblings` virtual source object.


dairiki:
I've just extracted a bunch of what was here to PR #1105, and have rebased this PR onto that one.

dairiki:
Superceded by #1108.  Closing

PR #1010, opened 2022-03-02
Add go/no-go tests to check that various sites are buildable

dairiki:
This adds a set of tests that check that all the sample sites included the Lektor source code build without error.

The sites built include the `example` site as well as the various test projects in the `tests` subdirectory.

A test is included which builds the master branch of [lektor-website]. However, since it is very slow and requires internet connectivity, this test is skipped by default.  A new `--runslow` custom pytest command-line parameter has been implemented which will enable this test.

[lektor-website]: https://github.com/lektor/lektor-website/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,16 +113,16 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
 
-      - name: Install Linux system dependencies
-        if: runner.os == 'Linux'
+      - name: Install ffmpeg (Linux)
+        if: runner.os == 'Linux' && matrix.install-imagemagick
         run: sudo apt-get install -y ffmpeg
 
-      - name: Install macOS system dependencies
-        if: startsWith(runner.os, 'macos') && matrix.install-imagemagick
+      - name: Install imagemagick and ffmpeg (macOS)
+        if: runner.os == 'macOS' && matrix.install-imagemagick
         run: brew install imagemagick ffmpeg
 
-      - name: Install Windows system dependencies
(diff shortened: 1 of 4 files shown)


nixjdm:
Great, thanks for this. Do you want to incorporate the runslow tests into the github action for publishing, as a final test, but still an automatic one?

dairiki:
> Great, thanks for this. Do you want to incorporate the runslow tests into the github action for publishing, as a final test, but still an automatic one?

That's an interesting option, though, if the test is going to fail, it would be nice to find out about it before then.

I suppose we could always --runslow the CI tests.
--runslow adds about a minute to the total test time on my dev machine.  Our `macos-latest py3.10` job already takes over ten minutes to run — so, really, the extra time is pretty much in the noise.

I disabled the test by default mostly for the benefit of developers (i.e. me) who are running the tests interactively, where including that one --runslow test triples the running time of `pytest tests` (or `tox -e py310`).

nixjdm:
I agree, and that marginal increase in the test time of the normal CI runs seems fine to me.

dairiki:
On the GitHub runners, the lektor-website --runslow test seems to add 2 to 3 minutes to the run time.
I've now enabled it on all the python test jobs (though the test is skipped if ImageMagick is not installed.)

I also noticed that `ffmpeg` is not installed by default on the Linux runners.  (Maybe it used to be?)
I've fixed the workflow to install it on the `ubuntu-latest py3.10` job.  (Installing ffmpeg adds 30 seconds or so to the workflow run time.)

PR #1015, opened 2022-03-19
Bug: `title_i18n` is None if not provided

relikd:
https://github.com/lektor/lektor/blob/a7c532a436f3c29dcf365ac9d004414e669dea14/lektor/build_programs.py#L33-L44

Optional method argument `title_i18n=None` can not be checked with `in` operator.

Suggesting a default value of `title_i18n={}` instead.

### DIFF
--- a/lektor/build_programs.py
+++ b/lektor/build_programs.py
@@ -31,7 +31,7 @@ class SourceInfo:
     """
 
     def __init__(
-        self, path, filename, alt=PRIMARY_ALT, type="unknown", title_i18n=None
+        self, path, filename, alt=PRIMARY_ALT, type="unknown", title_i18n={}
     ):
         self.path = path
         self.alt = alt


lektor/build_programs.py
```diff
@@ -31,7 +31,7 @@ class SourceInfo:
     """
 
     def __init__(
-        self, path, filename, alt=PRIMARY_ALT, type="unknown", title_i18n=None
+        self, path, filename, alt=PRIMARY_ALT, type="unknown", title_i18n={}
```
dairiki, line 34: Using a mutable for a default parameter value [is dangerous](https://www.google.com/search?q=python+mutable+default+argument&oq=python+mutable+d).

A better fix, I think, is to condense all the (nine lines of) logic below for initialization of self.title_i18n to the (I'm pretty sure) equivalent:
```
self.title_i18n = {"en": self.path}
if title_i18n:
    self.title_i18n.update(title_i18n)
```

====

While we're mucking about in here, it would probably be of value to expand the docstring for `SourceInfo` to describe how these are used.  (They appear to only be used for the `lektor.sourcesearch.find_files` functionality.  In turn that only appears to be used by the `/find` method of the admin API for the purpose of implementing the search functionality of the admin UI.)

`Find_files` searches for records whose titles or paths contain the search string.  When titles are internationalized, the `"en"` title and the title in the configured lang for he admin UI are searched.  (In Lektor, "en" is always default lang, and all *_18n map values should specify a value for "en".) 

In general, when one is implementing a custom build program for a custom record type, if one wants to support `find_files`, one should probably supply at least a title for lang="en" in `title_i18n`.  (If one does not want ones custom record to be findable via `find_files`, one may simply use the default implementation of  `Builder.describe_source_record` (which returns `None`).

relikd: I agree on the bad default parameter choice.
I am still trying to figure out how to prune a virtual object. Since my virtual object are created on the fly, previously generated files are silently forgotten once the initiating source page does not produce a virtual object anymore (eg. the conditions inside the page have changed). I was looking into `SourceInfo` and `VirtualSourceInfo` to remove a previously tracked dependency. By now I figured that you can't use `SourceInfo` for that purpose. 🤷

dairiki: > I am still trying to figure out how to prune a virtual object. 

Assuming you mean how to have your virtual objects _not_ get pruned...

Artifacts get pruned when none of their "primary sources" exist on disk (or if they have no primary sources).  Primary sources are those listed in the `sources` parameter to `BuildProgram.declare_artifact` (or `BuildState.new_artifact` or `Artifact.__init__`).  For the general case, since VirtualSourceObjects always live under a non-virtual Record, a good starting guess is `sources=list(virtual_source_object.record.iter_source_filenames())` — though, of course, it all depends on the specifics of your VirtualSourceObject.  (PR #1007 fixes VirtualSourceObject.iter_source_filenames() to return just that.)

relikd: No, I really meant pruning virtual objects. E.g., add tag "awesome" to page A, so it will generate a new v-obj for tag "awesome" pointing to page A. Then, edit page A to remove tag "awesome". A new v-obj will not be generated since there are no other pages containing this tag. But since the reference to page A still exists for the v-obj, it will not be pruned. If I were to delete page A, it would finally be pruned, I assume.

The logic to generate said v-obj is rather intricate since it covers more than one field and requires regex parsing. Something I'd like to do just once per build and not every time in a query expression.

My best shot so far, create a self-reference to the generating v-obj and check for the prefix of the virtual-path:

```py
class CustomSource(VirtualSourceObject):
    @property
    def path(self):  # used in VirtualSourceInfo
        build_url([record.path, '@custom-vpath', tag])

class CustomBuildProgram(BuildProgram):
    def build_artifact(self, artifact):
        self.source.pad.db.track_record_dependency(self.source)

def prune(self, build_state, destination_path, output_set, prefix):
    for a, file in build_state.iter_artifacts():
        for path, i in build_state.get_artifact_dependency_infos(a, []):
            if isinstance(i, VirtualSourceInfo) \
                    and path.startswith(prefix) \
                    and path not in output_set:
                reporter.report_pruned_artifact(a)
                prune_file_and_folder(file.filename, destination_path)
                build_state.remove_artifact(a)
```

.... But anyway, this is off-topic 😅

PR #1024, opened 2022-04-02
Save and preview

yagebu:
### Issue(s) Resolved

Supercedes #701

### Description of Changes

The current behaviour is to jump to the preview page after saving. This PR adds a new button ("Save Changes and Preview") which behaves like the current one and changes the main "Save Changes" button to only save the changes (and show a notification in the top right corner that saving succeeded). That behaviour seems to be more in line with how most "editors" behave on save.

For keyboard users, this adds a shortcut (Ctrl/Meta+p) which allows one to quickly jump between the edit (or any other) and the preview page. This was currently only possible with the Ctrl/Meta+s shortcut, which only worked on the edit page if one had actually changed something. Saving and previewing now takes one additional click (but I think having distinct shortcuts is more explicit and avoids saving unwanted changes if one just wants to jump to the preview page).

### DIFF
--- a/frontend/js/components/AdminLinkWithHotkey.tsx
+++ b/frontend/js/components/AdminLinkWithHotkey.tsx
@@ -1,7 +1,7 @@
 import React, { MutableRefObject, useCallback, useEffect, useRef } from "react";
 import { NavLink } from "react-router-dom";
 import { useRecord } from "../context/record-context";
-import { KeyboardShortcut, keyboardShortcutHandler } from "../utils";
+import { getKey, KeyboardShortcut, keyboardShortcutHandler } from "../utils";
 import { AdminLinkProps } from "./AdminLink";
 import { adminPath } from "./use-go-to-admin-page";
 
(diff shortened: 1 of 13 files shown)


dairiki:
What happens when you hit the preview hot-key after making some changes but before saving?

From a brief look at the code, it looks like it does the same thing as hitting the preview button does now which is to display a "You have unsaved information. Are you sure you want to leave this page?" dialog.

Since a very common operation is to make some edits and then want to see the result, I think it might be better to have the hotkey do a save-and-preview.

Or maybe (if there are changes) put up a dialog with three choices:

    You have unsaved changes:
     - Cancel (return to edit view)
     - Discard Changes (and preview)
     - Save (and preview) [default focussed - so `<ctl/alt>-P` `<return>` will do a save and preview.]

yagebu:
> Since a very common operation is to make some edits and then want to see the result, I think it might be better to have the hotkey do a save-and-preview.
> Or maybe (if there are changes) put up a dialog with three choices:

I like the dialog idea. I think it's good if the hotkey is for the preview button (which is present on more than just the edit page) and if the hotkey behaves somewhat consistently (always "open preview page"). The dialog with pre-selected button seems like the right solution to still have the very common save-and-preview easily accessible (either `Ctrl+s Ctrl+p` or `Ctrl+p Enter`)

yagebu:
Closing, let's do #1052 instead

PR #1027, opened 2022-04-13
Support livereloading on dev server

### LINKED ISSUE #453: Server : live reload ability
wish: live reload ability in the server.
As of today I have always to click reload on the browser to check my modifications.

frostming:
### Issue(s) Resolved

Fixes #453

### Description of Changes

When running a dev server with `lektor server`, users no longer need to keep ctrl+r on the browser, it is reloaded after every rebuild.

https://user-images.githubusercontent.com/16336606/163154120-77fc93c6-01b6-4da6-a6a7-130166d21ac8.mov

- [x] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

The implementation is inspired and adapted from [django-browser-reload](https://github.com/adamchainz/django-browser-reload), read the README for how it works


lektor/devserver.py
```diff
@@ -53,6 +53,7 @@ def run(self):
                 for ts, _, _ in watcher:
                     if self.last_build is None or ts > self.last_build:
                         self.build()
+                        trigger_reload()
```
dairiki: There are other paths which trigger a rebuild: hitting the "refresh build" button in the admin UI is one.

To catch all of those, the reload trigger should probably be triggered in [Builder.build_all][build_all] where the `after-build-all` hook is called. (Or, even better, perhaps it should hook `after-build-all`.  That, currently, is only possible from a plugin.)

[build_all]: https://github.com/lektor/lektor/blob/89932bee3b1f2e68163b4f1f3337ce77e8825eb7/lektor/builder.py#L1232

lektor/admin/modules/livereload.py
```diff
@@ -0,0 +1,56 @@
...
+    """Trigger reload after a debounce delay."""
+    global reload_timer
+    if reload_timer is not None:
+        reload_timer.cancel()
+
+    reload_timer = threading.Timer(RELOAD_DEBOUNCE_TIME, reload_event.set)
+    reload_timer.start()
+
+
+def message(type_: str, **kwargs: Any) -> bytes:
```
dairiki: We already have machinery to generate a SSE view.  See the `lektor.admin.utils.eventstream` view decorator which is used by the `/publish` endpoint of the admin API.

lektor/admin/templates/livereload.html
```diff
@@ -0,0 +1,6 @@
+<script
```
dairiki: The whole `listener.js` script (especially if simplified a bit) could just be inlined here.

lektor/admin/modules/static/livereload/worker.js
```diff
@@ -0,0 +1,112 @@
+/* eslint-env worker */
+"use strict";
+
+let eventsPath = null;
+let port = null;
+let currentVersionId = null;
+let eventSource = null;
+
+addEventListener("connect", (event) => {
+  // Only keep one active port, for whichever tab was last loaded.
```
dairiki: XXX: I'm not sure this code does what is intended.  In any case, the whole system does not appear to be working for me.

We want to run a single `EventSource` in the worker, and distribute messages received from that to each connected browser context.

This code appears to run (modulo race conditions) an `EventSource` for each browser context.  (If that's the case, that defeats the purpose of using the SharedWorker.)

If we don't mind hard-coding the SSE source URL in the code (i.e. we dynamically generate worker.js, see below), I think this is sufficient to replace all of worker.js:
```js
let reloadEvents = new EventSource("/__reload__/events");

addEventListener("connect", (event) => {
    let currentVersionId = null;
    let port = event.ports[0];
    let reload = () => {
        port.postMessage("Reload");
    }

    port.start();

    reloadEvents.addEventListener("message", (event) => {
        const message = JSON.parse(event.data);
        if (message.type === "ping") {
            if (currentVersionId !== null && currentVersionId !== message.versionId) {
                reload();
            }
            currentVersionId = message.versionId;
        }
        else if (message.type === "reload") {
            reload();
        }
    });
});
```

lektor/admin/modules/static/livereload/worker.js
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    eventsPath = event.data.eventsPath;
+  }
+};
+
+let connectAttempts;
+let connectTimeoutMs;
+
+const resetConnectTimeout = () => {
```
dairiki: My feeling is that all this retry logic is unnecessary.

`EventSource`s are pretty robust, in that they do a pretty good job of retrying closed connections all on their own.  I'm pretty sure that if the dev server is stopped for a period then restarted a plain old `EventSource` will recover the connection all on its own.

lektor/admin/modules/static/livereload/worker.js
```diff
@@ -0,0 +1,112 @@
...
+  // Only keep one active port, for whichever tab was last loaded.
+  if (port) {
+    port.close();
+  }
+  port = event.ports[0];
+  port.addEventListener("message", receiveMessage);
+  port.start();
+});
+
+const receiveMessage = (event) => {
```
dairiki: Is this an unnecessary complication?

Perhaps the worker.js JS could be dynamically generated by the API (so that the server URL can be hard-coded in the worker code.)

E.g. JS comes from `/__reload__/worker.js`.  Hard-coded in that JS would be the `/__reload__/events` URL.

lektor/admin/modules/livereload.py
```diff
@@ -0,0 +1,56 @@
...
+from werkzeug.exceptions import NotAcceptable
+
+PING_DELAY = 1.0
+RELOAD_DEBOUNCE_TIME = 0.05  # seconds
+
+bp = Blueprint("livereload", __name__, static_folder="static")
+# Use a random ID to detect reloading, which will be changed after reloaded
+version_id = secrets.token_urlsafe(16)
+
+reload_event = threading.Event()
```
dairiki: I don't think an Event will work correctly here in the case that there is more than one /__reload__/events SSE stream running at the same time (e.g. if the user has two different browsers open.)  In that case, when a reload is triggered, there is a race — I don't think all the threads waiting on the Event are guaranteed to be woken (since the first thing each of those threads does when woken is the clear the Event.)

Look into using a `Condition` here instead.

lektor/admin/modules/livereload.py
```diff
@@ -0,0 +1,56 @@
...
+reload_timer: Optional[threading.Timer] = None
+
+
+def trigger_reload():
+    """Trigger reload after a debounce delay."""
+    global reload_timer
+    if reload_timer is not None:
+        reload_timer.cancel()
+
+    reload_timer = threading.Timer(RELOAD_DEBOUNCE_TIME, reload_event.set)
```
dairiki: Is the debounce logic necessary?  (How often are two build_all cycles going to complete within 50ms of each other?)

dairiki (commented):
Good ideas here, though I think there are some issues.

I have taken a quick look at this, and have tried to run your code.  (It didn't seem to quite work for me — If I had multiple preview windows open, only one would reload.)

In addition to specific details described below, here are some general comments.

----

I think that perhaps just sending a "reload everything" signal at the end of each "build all" cycle is less than ideal.

One issue is that there is a spurious reload after editing a page in the Admin UI:
- Use Admin UI to edit a page.
- Hit the save button.  Admin UI saves and switches to "Preview" mode, loads preview
- The resulting build_all takes a while, finally completes, triggers a rebuild SSE event
- Preview reloads a second (probably redundant) time.

Another is that every page is reloaded when any page is edited.

Perhaps an alternative would be to publish a SSE version of the `lektor.reporter` log.  (That's the mechanism responsible for reporting the "Started build", "U index.html", "Finished build in 3.50 sec" messages normally displayed on the terminal console.)  Individual preview pages could then watch that log, only triggering a reload when it indicates that that specific page has been rebuilt.

The SSE view of the log could also be useful for other things, not least of which is displaying it in some way in the Admin GUI.

lektor/devserver.py
frostming: > There are other paths which trigger a rebuild: hitting the "refresh build" button in the admin UI is one.

It seems not the case, it only fires a `clean/prune` action.

lektor/admin/templates/livereload.html
frostming: Sure, will do.

lektor/admin/modules/static/livereload/worker.js
frostming: Since the worker script is loaded by `SharedWorker` and there is no template rendering there, we are not able to do so(at least with little effort). The events URL will be passed to the worker via message post and the worker can compare the URL with the one stored previously to determine whether to re-initialize the event source(In case you changed the URL and don't want to refresh the page).

lektor/admin/modules/static/livereload/worker.js
frostming: Sure, will do.

lektor/admin/modules/livereload.py
frostming: Removed it.

lektor/admin/modules/livereload.py
frostming: The `Event` wraps a `Condition` inside:

```python
    def wait(self, timeout=None):
        with self._cond:
            signaled = self._flag
            if not signaled:
                signaled = self._cond.wait(timeout)
            return signaled
```

When multiple SSE views are waiting for the same event(blocked at line 5), a flip of the event will trigger the reload message for all waiters. Unless it comes late to see the flag already cleared.

frostming:
Thanks for your review comments, I take some of your advice and others remaining to be discussed.

> One issue is that there is a spurious reload after editing a page in the Admin UI:

Yeah that is a problem, I don't come up with a good way to solve this, what about disabling live reload for admin preview.

> Another is that every page is reloaded when any page is edited.
>
> Perhaps an alternative would be to publish a SSE version of the lektor.reporter log. (That's the mechanism responsible for reporting the "Started build", "U index.html", "Finished build in 3.50 sec" messages normally displayed on the terminal console.) Individual preview pages could then watch that log, only triggering a reload when it indicates that that specific page has been rebuilt.

Maybe we can do this by replacing the `Event` with a `Queue`, but it doesn't solve the race condition: the messages are consumed exclusively across multiple SSE views. 

To implement broadcasting, we have to register the SSE views into a global state when a new SSE view is established, and when files are rebuilt, the information is broadcast to all registerees. What do you think?

lektor/admin/modules/static/livereload/worker.js
dairiki: It's not hard to serve `worker.js` from a flask view which renders a template.
```py
@bp.route("/events/worker.js")
def worker_js():
    response = make_response(render_template("worker.js", events_url=url_for("livereload.events")))
    response.content_type = `"application/javascript"`
    return response
```

lektor/devserver.py
dairiki: > > There are other paths which trigger a rebuild: hitting the "refresh build" button in the admin UI is one.
> 
> It seems not the case, it only fires a `clean/prune` action.

Erm.  That appears to be a bug in the frontend, I think (maybe?) The reload button is poking the `/api/clean` which, as you say, just does a prune.  There is an `/api/build` which does a build_all, though.  (It appears to be invoked by the frontend as part of the publish process.)

In any case, are there any times build_all is called when you wouldn't want to trigger a reload?  If not, then build_all is the right place to call the trigger.

lektor/admin/modules/livereload.py
dairiki: > Unless it comes late to see the flag already cleared.

Well, yes.  That's my point: there's a race.

In this case, we don't really care about the value of the flag.  We just want to wake up all the threads that are waiting for a rebuild.  Using just a plain `Condition` will do exactly this.  (Use Condition.notify_all to wake up all the waiting threads.)

frostming: > We just want to wake up all the threads that are waiting for a rebuild. Using just a plain Condition will do exactly this. (Use Condition.notify_all to wake up all the waiting threads.)

What I mean is, if the view is already in the waiting queue, `Event` also does the trick under the hood. But if it fails to catch the last bus, it will also miss the `Condition.notify_all`. Anyway it is true that `Event = flag + Condition`, we can ignore the flag and use a plain condition instead, but it seems not that different from the current approach. I am Okay to change it both ways.

frostming:
> If I had multiple preview windows open, only one would reload.

Yes, that is intended, because reloading all opening tabs may be expensive

For selective reload, I have another concern. What about a summary page that shows the contents of child pages? When child pages changed, should we trigger reload on the summary page? The current simple and brute approach may be acceptable, yeah? (I tried Next.js, it doesn't seem to perform selective reload either).

lektor/devserver.py
frostming: I didn't hook into `Builder` class because I don't want it to affect other behaviors than dev server, where `BackgroundBuilder` seems the proper place to do it.

lektor/admin/modules/livereload.py
dairiki: Having thought about this overnight, you're right in that using an `Event` will work _most of the time_.  But neither `Event` nor a plain `Condition` is actually correct.

The listener threads (the thread(s) servicing the `/__reload__/events` request(s)) do external I/O. Specifically, they are responsible for sending data to the HTTP clients.  As such, they can spend significant periods of time waiting for I/O to complete.  If a rebuild happens while the thread is blocked for I/O (rather than waiting in the `Event.wait` call), then, if there are multiple listeners, it is likely to miss the notification, as one of the other listeners will likely see the event and clear the flag before the blocked thread gets a chance to `wait` on the event.  Using a lone `Condition` is even worse in this regard, since the event will definitely be missed by all threads that aren't blocked in the `Condition.wait` call.

A better solution would be to use a counter protected by a `Condition`.  On each build the counter gets incremented.  Every time a listener sees a counter value it hasn't yet seen, it knows a build has happened.

frostming: SGTM, I will try this approach tomorrow.

dairiki:
> Yes, that is intended, because reloading all opening tabs may be expensive

A not unreasonable use case to me seems to be to have one "preview" browser window open to a particular page, while editing (using the admin UI) in another window.  As things are now, that doesn't work — the preview iframe in the admin UI which is displayed after saving changes grabs the live-reload (when it doesn't really need it) and the open preview window must be manually reloaded to see any changes.

I would certainly vote for reloading all open and affected windows.

> For selective reload, I have another concern. What about a summary page that shows the contents of child pages? When child pages changed, should we trigger reload on the summary page?

My thinking is that a reload would be triggered whenever the content (e.g. the `index.html`) of the page being viewed in the current window is re-built. In a case where a parent page includes snippets of its children, Lektor's dependency tracking system would ensure that the parent gets rebuilt whenever one of its children is modified.

> The current simple and brute approach may be acceptable, yeah? (I tried Next.js, it doesn't seem to perform selective reload either).

Have you tried using the admin UI to edit pages with live reload enabled?  Yes, it works, but the reload some seconds after each save is distracting.  A simple solution may be to disable live-reload in the preview iframes altogether?

> Maybe we can do this by replacing the Event with a Queue, but it doesn't solve the race condition: the messages are consumed exclusively across multiple SSE views.

> To implement broadcasting, we have to register the SSE views into a global state when a new SSE view is established, and when files are rebuilt, the information is broadcast to all registerees. What do you think?

Yes.  Each consumer (view thread) would need to create its own Queue which it then would add to a (lock-protected) "global" list.  (The list shouldn't really be global, but should be attached to, e.g., the parent `flask.Application`, or perhaps the (to be written) Lektor `Reporter` subclass responsible for reporting the event stream.  Instead of a list, perhaps a `WeakValueDict` should be used, to make cleanup of defunct Queues automatic.)  The producer would then add each message to each of the Queues in the list.

frostming:
Some changes based on the last review comments:

- Live reload is disabled in the admin preview iframe
- All opening tabs and windows will be reloaded
- Change the reload event to a change stream queue.
   - The built artifact will be put into the queue after each individual build finished
   - The change will be broadcast to all view threads listening to the change stream
   - The live reload view will only trigger reload when the changed artifact name matches the current page

@dairiki any chance to try the new approach?

lektor/reporter.py
```diff
@@ -112,6 +122,7 @@ def build_artifact(self, artifact, build_func, is_current):
         try:
             yield
         finally:
+            report_artifact_built(artifact, is_current)
```
frostming: This will be called from ANY reporter, but the function will only send messages to `ChangeStreamReporter`s

relikd:
chiming in here to add another perspective ;-)

I sometimes leave a tab with the old version open just to compare the two. So I can check my progress and see if my changes made an improvement. And if not, reverse to the previous version. For example, on plugin development or tweaking a CSS file. The proposed changes, especially reloading all tabs, will make that impossible for me.

I am not saying I am against this PR, just that there are use cases against live reloading. Also, I save my source files way more often than I reload my preview tab. I am sure my programming / debugging style will change with this PR, I just don't know if I will be happy with this. But I assume many others will appreciate the change.

frostming:
thanks, so enabling livereloading only for the most recently opened tab seems a good trade-off?

dairiki:
> so enabling livereloading only for the most recently opened tab seems a good trade-off?

To me, that still seems like confusing and unexpected behavior to me.

If we do want to support having some previews auto-reload and others not, adding a control to the page seems to be the right way to do that.  This could be something like a "reload" icon (next to the existing edit pencil) which, when clicked, toggles between enabling and disabling live reload.  (When live-reload is disabled, the icon would have a slash through it, or be dimmed-out or something, to visually indicate the change of state.)

> Yes, that is intended, because reloading all opening tabs may be expensive

Note that if server load is a real concern when reloading multiple previews at once, the [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) could be used to defer the reload on hidden tabs/windows until they are made visible.

lektor/reporter.py
```diff
@@ -177,6 +188,19 @@ class NullReporter(Reporter):
     pass
 
 
+class ChangeStreamReporter(Reporter):
```
dairiki: Since all `Reporter` subclasses call `report_artifact_built`, I think this should not be a subclass of `Reporter`.  It doesn't really serve the same purpose as the rest of the `Reporter` subclasses, rather this is a per-listener instance responsible for receiving the artifact-built notifications.
It should maybe be called something like `ChangeStreamListener` rather than `ChangeStreamReporter`.

Perhaps this class isn't really necessary at all.  Maybe the view code in `livereload.py` could just look something like (ping logic removed for simplicity):

```py
from lektor.reporter import reporter

@bp.route("/events")
@eventstream
def events():
    updated_artifacts = queue.Queue()
    with reporter.register_artifact_built_callback(updated_artifacts.put):
        while True:
            artifact = updated_artifacts.get()
            yield {"type": "reload", "path": artifact.artifact_name}

```

(If it is renamed, then `_change_stream_reporters` should also be renamed  — e.g. to `_change_stream_listeners`).

lektor/reporter.py
```diff
@@ -10,6 +12,14 @@
...
 _reporter_stack = LocalStack()
 _build_buffer_stack = LocalStack()
+_change_stream_reporters = WeakSet()
+
+
+def report_artifact_built(artifact, is_current):
+    if is_current:
+        return
+    for reporter in _change_stream_reporters:
+        reporter.change_stream.put(artifact)
```
dairiki: Can `_change_stream_reporters` and `report_artifact_built` be properties/methods of `Reporter` rather than globals?

lektor/reporter.py
dairiki: Nice.  I think this does make more sense than having a specialized Reporter class to do this.

dairiki (commented):
I do like the idea of only reloading when the viewed page is rebuilt.  And, thankfully, it doesn't appear that it requires that much extra code.

A few comments below.

dairiki:
> @dairiki any chance to try the new approach?

I apologize, but I've been busy with things other than Lektor and have not had much time as of late.  (I'm not sure when the situation will improve — my schedule for the next couple of months is pretty packed, though, hopefully, I can spare a bit of time for Lektor.)

I've just looked at your latest patches, and made a few comments.

lektor/reporter.py
frostming: `report_artifact_built` and `_change_stream_reporters` *broadcasts* to all threaded reporters, while `Reporter` is stored on a `werkzeug.local.LocalStack`, which is specific for each thread.

frostming:
> just that there are use cases against live reloading

@relikd While I acknowledge this use case, I am curious what the frontend people would do, where live reloading is quite common in webpack, vite or other packers.

lektor/reporter.py
dairiki: Of course, you're right about that.

`_change_stream_reporters` could still be made a class property.  Especially if it is only accessed by methods of `Reporter`, I would be in favor of that.

frostming: Sure, I changed it to a class attribute

xlotlu:
> > just that there are use cases against live reloading
> 
> @relikd While I acknowledge this use case, I am curious what the frontend people would do, where live reloading is quite common in webpack, vite or other packers.

This is not an issue, just provide a config option to disable live-reloading. I say disable, because I think the default should be with it enabled. It's just a very important feature, so important that I find it essential both for development and content editing. I.e., modifying a page's template, content or attachments should result in a page refresh in the browser, if that page is currently being watched (read: open). Or more to the point, a rebuild should trigger a live reload notification.

This would also fix the duplicate codepath in admin. Admin would thus refrain from any refresh logic, and stick to being an editing frontend.

Regards

relikd:
One more thing to think of: what happens if user has `/pth/` open and renames the folder to `/path/` which will trigger a slug change. Will the live-reload refresh to a 404 page? Or redirect to the new slug which would require more logic? (I do not think this is the right way). Before live-reload, you could click on a parent element (admin UI), or directly manipulate the url field. … I do not expect an answer on this comment, just a thought that came up.

dairiki:
I've just pushed a couple of small changes.  (I hope you don't mind.)

### Use [BroadcastChannel](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API)

This avoid losing sleep over the lack of garbage collection for the list of MessagePorts in `ports`. (There seems to be no clean way to clean up stale ports. This [warning](https://html.spec.whatwg.org/multipage/web-messaging.html#ports-and-garbage-collection) confirms my gut feeling that keeping stale ports around and continuing to send them messages is not a great practice.)

### Per-window opt-out of live-reload via `?livereload=false` query param

As described [above](https://github.com/lektor/lektor/pull/1027#issuecomment-1108466572), some workflows may involve having some preview windows live-reload while others do not.

What remains is to add a nicer UI to control the control the `?livereload` query param.

Perhaps it would be enough to make live-reload opt-in (i.e. require explicit `?livereload=true` to enable), then change the "return to website" ("eyeball") button in the admin UI to point to the live-reload version?
(While we're at it, we should change the "return to website" button to a bona fide link so that one can use mousebutton 2 or 3 to open it in new tab/window.)

lektor/reporter.py
```diff
@@ -97,6 +99,14 @@ def build(self, activity, builder):
             self.builder_stack.pop()
             self.finish_build(activity, now)
 
+    @contextmanager
```
dairiki, line 102: Sorry for the wishy-washing, but I think the API would be cleaner if the Reporter just kept a list of generic callbacks, rather than a list of Queues. The queue is an implementation detail about how the caller wants to deal with notifications (i.e. pass them via queue to another thread) — the callback should deal with that part.  If refactored this way, the API would also support clients who don't need the queue (notification processed in same thread) or clients who wish the pass the notification on to a subprocess, etc.

E.g. usage of propose API:

```py
built_artifacts = queue.Queue()

# the lambda is not really necessary, but added for clarity of intent
with reporter.notify_on_artifact_built(lambda artifact: built_artifacts.put(artifact)):
    while True:
        artifact = built_artifacts.get()
        # Do something with the artifact
```

I'm happy to work on this (after the weekend) if you don't want to.

lektor/reporter.py
```diff
@@ -97,6 +99,14 @@ def build(self, activity, builder):
...
+    def register_change_stream(self, stream):
```
dairiki: Sorry for the wishy-washing, but I think the API would be cleaner if the Reporter just kept a list of generic callbacks, rather than a list of Queues. The queue is an implementation detail about how the caller wants to deal with notifications (i.e. pass them via queue to another thread) — the callback should deal with that part.  If refactored this way, the API would also support clients who don't need the queue (notification processed in same thread) or clients who wish the pass the notification on to a subprocess, etc.

E.g. usage of propose API:

```py
built_artifacts = queue.Queue()

# the lambda is not really necessary, but added for clarity of intent
with reporter.notify_on_artifact_built(lambda artifact: built_artifacts.put(artifact)):
    while True:
        artifact = built_artifacts.get()
        # Do something with the artifact
```

I'm happy to work on this (after the weekend) if you don't want to.

frostming: SGTM, just the lambda isn't necessary, you can use `build_artifacts.put`

frostming:
@dairiki Hi, is there any work remaining here?

dairiki:
> @dairiki Hi, is there any work remaining here?

I apologize for letting this drop for so long.

I think my hesitation about merging the last I looked at this had to do with the addition of the `--no-reload` CLI parameter.  At the time, it seemed like both overkill, and at the same time, not as useful as it could be.  It seems like a per-window control to disable the live-reload functionality would be more useful (e.g. allowing for a side-by-side view of the original and edited page).

I had started on https://github.com/dairiki/lektor/tree/feature.tool-drawer to provide a place in the UI for a per-page toggle-live-reload control but then got distracted/busy with other things.

At this point, I say it's ready to merge.  (I can work on the per-page live-reload toggle later...)
I'll give it a couple of days to see if anyone else — @xlotlu, @relikd, @nixjdm — has something to say.

dairiki:
Merged to `master`, and available in release 3.4.0b4.

Thank you @frostming! Apologies for the delay.

PR #1030, opened 2022-04-15
allow absolute urls in `asseturl`

relikd:
### Description of Changes

This PR adds a new `absolute` parameter option to `asseturl`.
This way you can have absolute asset URLs while the remaining website stays relative. Especially useful for error pages ([doc](https://www.getlektor.com/docs/guides/error-pages/)) with stylesheets etc. As long as you think of manually adding absolute URLs in your error pages when necessary.

We could also allow all `**kwargs` to pass to `url_to`, but `alt` does not make sense for assets ~and `external` does not seem to work with `!`-urls.~
EDIT: my bad, `absolute` overrides `external`, thats why it didn't work.

Disclaimer: I have not run any tests, though it should not break anything (fingers crossed).

### DIFF
--- a/lektor/context.py
+++ b/lektor/context.py
@@ -18,7 +18,7 @@ def url_to(*args, **kwargs):
     return ctx.url_to(*args, **kwargs)
 
 
-def get_asset_url(asset):
+def get_asset_url(asset, absolute=None):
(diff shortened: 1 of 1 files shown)


relikd:
my bad, `absolute` overrides `external`, thats why it didn't work.

PR #1045, opened 2022-07-19
Possible fix for annoying admin page scrolling

### LINKED ISSUE #1: Page is moved up on input in long textboxes [Firefox]
When editing long text the page moved up as far as possible on input.
Firefox only.

### LINKED ISSUE #1038: administrator interface: page scrolling up at each keypress
When I try to edit my blog in the administrator interface, and the window is highr than the viewport, and I'm editing at the bottom, each keypress makes the window immediatilly scroll to the top.
Lektor, version 3.3.4 on firefox 101.0.1 - win 10

dairiki:
Here is a possible fix for the annoying admin page scrolling behavior described in #1038.

The issue is caused by the way the MultilieTextInputWidget auto-sizes the textarea. Currently, whenever the content of the textarea is changed, [the code][] momentarily sets the CSS height of the textarea to `auto` in order to be able to deduce the correct `scrollHeight` of the element.  That causes the textarea to shrink, so can cause the window to scroll.  (When the CSS height is finally reset to a fixed value, the window does not scroll back.)

[the code]: https://github.com/lektor/lektor/blob/791659667b9a53fa40e39e85b99b5850d795bfb4/frontend/js/widgets/MultiLineTextInputWidget.tsx#L16

This PR currently fixes the issue by using the `TextareaAutosize` component provided by the [react-textarea-autosize](https://github.com/Andarist/react-textarea-autosize) library.  That library uses a separate off-screen textarea to compute the natural height, thus avoiding the step of setting CSS height to `auto`.

## Added Drag-Handle

In the second commit of this PR, I've enabled the ability to resize the textarea (using the browser-native drag-handle.)
When editing pages that have large content boxes, sometimes I find it useful to be able to make the textarea for that large content smaller, so as not to have to scroll across all of that content to get to fields above and below the textarea.

As things stand, `react-textarea-autosize` still resizes the textarea to its natural height whenever the number of lines of textarea content changes.

### Issue(s) Resolved

Fixes #1038
Fixes #1

### DIFF
--- a/frontend/js/widgets/MultiLineTextInputWidget.tsx
+++ b/frontend/js/widgets/MultiLineTextInputWidget.tsx
@@ -1,4 +1,5 @@
-import React, { ChangeEvent, useCallback, useEffect, useRef } from "react";
+import React, { ChangeEvent, useCallback } from "react";
+import TextareaAutosize from "react-textarea-autosize";
 import { getInputClass, WidgetProps } from "./types";
 
 export function MultiLineTextInputWidget({
@@ -8,16 +9,6 @@ export function MultiLineTextInputWidget({
   disabled,
   onChange: onChangeProp,
 }: WidgetProps) {
-  const textarea = useRef<HTMLTextAreaElement | null>(null);
-
-  const recalculateSize = useCallback(() => {
-    const node = textarea.current;
-    if (node) {
-      node.style.height = "auto";
-      node.style.height = node.scrollHeight + "px";
-    }
-  }, []);
-
   const onChange = useCallback(
     (event: ChangeEvent<HTMLTextAreaElement>) => {
       onChangeProp(event.target.value);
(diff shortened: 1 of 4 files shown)


nixjdm:
This does seem to fix the issue of jumping to the top of the page to me! I'll be actually using lektor from this branch until it merges. :)

The drag handle also works, but feels, odd. When testing this with a very long markdown field, trying to shrink the box feels slugish, and a bit confused between resizing the box and scrolling in it. I think it might be because that the box starts out larger than the viewport. Looking at GitHub's drag handle for this very comment, if I input many lines, the resizing seems totally fine and snappy, unless I've made the box very long and scrolled all the way down. Then shrinking it feels the same as it does in our admin on this branch. So then, maybe in addition to the drag handle, the perception of it feeling slow to me would be fixed if the original vertical height was not set to the max / greater than the viewport. Maybe set default window height to fit content up to a little less than the viewport height?

dairiki:
Closing in favor of #1050.

PR #1048, opened 2022-07-22
Fixes for #203: visible children of hidden parents

### LINKED ISSUE #203: overriding parent _hidden field in content does not have desired effect
Per docs [here](https://www.getlektor.com/docs/api/db/system-fields/hidden/) on the `_hidden` system field:

> This also automatically applies to all children of a page unless they forcefully override this setting.

I'm trying to use this in a website to expose urls like:

/podcasts/pistons/3

while having a 404 for

/podcasts/

Here's a branch illustrating:

https://github.com/krosaen/brosaen-lektor/compare/unhidden-content-children?expand=1

note I'm hiding `/podcasts/` and attempting to unhide `/podcasts/pistons/`. Should this be possible? Not sure if this is a bug or a feature request / misunderstanding of the docs.

dairiki:
This PR allows one to have visible children under a hidden parent page.  No artifact (e.g. `index.html`) will be built for the hidden parent, but artifact(s) will now be built for any explicitly non-hidden children.

Contrary to what the documentation says, up until now, Lektor made no attempt to build children of a hidden parent.

Other changes here:
- Fixes some related bugs having to do with URL resolution to pages that have hidden parents.
- Fix the builders pruning logic to prune artifacts that belong to hidden pages.  (These were presumably built at a time when the page was not marked as hidden.)
- Fix the BooleanInputWidget’s logic for setting checkboxes to the _indeterminate_ state.  Previously, checkboxes were only set to indeterminate if their default value was true.  I think the proper (more useful) behavior is to set the checkbox to indeterminate if the `contents.lr` file does not explicitly set a value for the field, regardless of what the default value is.
- Updates the BooleanInputWidget (checkbox) so that typing \<Delete> or \<Backspace> (while it has the focus) will revert it to _unset_ (or _indeterminate_) state, rather than `true` or `false`.  This is important for the `_hidden` field, since when the `_hidden` field is unset it means something different (e.g. "inherit from parent") than when `_hidden = false` (or `_hidden = true`).

### Issue(s) Resolved

Fixes #203

### Related Issues / Links

Clarified documentation on the `_missing` field in lektor/lektor-website@7a6ccd0e38d1c645da18bfbe167ad96756bb63b8

### Description of Changes

- [x] Should implement a real tri-state checkbox widget in the admin UI.  (`_hidden` unset ≠ `_hidden = no`)

- [x] Link to corresponding documentation pull request for [getlektor.com](https://github.com/lektor/lektor-website) (See lektor/lektor-website@7a6ccd0e38d1c645da18bfbe167ad96756bb63b8.)

### DIFF
--- a/frontend/js/widgets/BooleanInputWidget.tsx
+++ b/frontend/js/widgets/BooleanInputWidget.tsx
@@ -37,7 +37,7 @@ export function BooleanInputWidget({
           onChange={(ev) => {
             onChange(ev.target.checked ? "yes" : "no");
           }}
-          onKeyPress={(ev) => {
+          onKeyDown={(ev) => {
             if (isDeleteOrBackspace(ev)) {
               ev.preventDefault();
               ev.stopPropagation();
--- a/frontend/js/widgets/BooleanInputWidget.tsx
+++ b/frontend/js/widgets/BooleanInputWidget.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useRef } from "react";
 import { WidgetProps } from "./types";
 import { trans_obj } from "../i18n";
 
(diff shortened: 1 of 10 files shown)


dairiki:
This is now fixed in the `master` branch (and the [3.4.0b1](https://pypi.org/project/Lektor/3.4.0b1/) pre-release).

If there's a desire for this to be backported to the 3.3 branch, let me know.

PR #1046, opened 2022-07-21
Another possible fix for annoying admin page scrolling

dairiki:
This is an alternative to PR #1045.

I've now found the [fit-textarea](https://github.com/fregante/fit-textarea) which seems both cleaner and light-weight than the `react-textarea-autosize` library used in #1045.

A limitation of `fit-textarea` is that it always wants the textarea to be sized to exactly fit its content. (A minimum height can be set, but no maximum.)  Without some ugly hacks, there is no way to support user-resizing via drag handles.

That's both good and bad.
- Con: If you have some fields with many, many lines of data, you may have to scroll the page a lot to get to the fields below it.
- Pro: Since textareas never have scroll-bars, there's never any confusion about what your scroll-wheel is scrolling. 

### Issue(s) Resolved

Fixes #1038
Fixes #1

### Related Issues / Links

Supercedes #1045

### Description of Changes

- [ ] Wrote at least one-line docstrings (for any new functions)
- [ ] Added unit test(s) covering the changes (if testable)

- [ ] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))
- [ ] Link to corresponding documentation pull request for [getlektor.com](https://github.com/lektor/lektor-website)

PR #1050, opened 2022-07-22
Yet another possible fix for annoying admin page scrolling

### LINKED ISSUE #1: Page is moved up on input in long textboxes [Firefox]

### LINKED ISSUE #1038: administrator interface: page scrolling up at each keypress

dairiki:
This is yet another try at fixing the wonky scrolling described in #1038.

This requires no external libraries.  It's a primarily CSS-based approach based on ideas from [this article](https://css-tricks.com/the-cleanest-trick-for-autogrowing-textareas/).

This seems to work as well or better than #1046, plus it requires no external libraries.

[Edit: I think this might be fixed.]
<del>In playing with this, sometimes, especially on pages with a lot of content and thus a very large textarea, there is sometimes one spurious page scroll that happens when the textarea is first typed in.  It only seems to happen once and does not reoccur (until the page is reloaded.)</del>

This PR currently supports manual resizing of textareas via drag-handles.  
<del>(Once a textarea is manually resized, automatic resizing is disabled until the page reloads.)</del>

<del>I'm still undecided as to whether or not I like the addition of the drag-handles.</del>

Currently, this is tuned as follows:

On initial page load, textareas are limited to a given height (currently the limit is half the window height).  If the content fits in that height, then the textarea will be set to "autosize" mode and will expand (or contract) as necessary as new content is added (or deleted) to the textarea.  If the content is to tall to fit in the target height, the textarea is set to manual-sized mode.  There will be a scrollbar.  One can resize with the drag-handle.

If the drag-handle is used to make an autosized textarea smaller, it will become statically sized.
If the drag-handle is used to make a statically sized textarea larger than it's natural size, it will become automatically sized.

Also, there is a little button in the upper-right corner of the text area which can be used to toggle the autosize state.

### Issue(s) Resolved

Fixes #1038
Fixes #1

### Related Issues / Links

Supercedes #1045 
Supercedes #1046

### DIFF
--- a/frontend/js/widgets/MultiLineTextInputWidget.tsx
+++ b/frontend/js/widgets/MultiLineTextInputWidget.tsx
@@ -1,4 +1,4 @@
-import React, { ChangeEvent, useCallback, useEffect, useRef } from "react";
+import React, { ChangeEvent } from "react";
 import { getInputClass, WidgetProps } from "./types";
 
 export function MultiLineTextInputWidget({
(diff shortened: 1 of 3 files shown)


dairiki:
I've backported the first commit in the PR to the `3.3-maintenance` branch just so that we can get a (hopefully) working release out.

I've only backported the "simple" textarea-always-autosized-to-fit-its-content part of this PR.

I do like the hairier part of this PR that also allows for manually resizing the textareas via drag-handles, but will wait a bit before merging that.

PR #1053, opened 2022-08-12
Admin preview view fails to track links with anchors

dairiki:
In the _preview_ view of the Admin UI, normally when one clicks on links to other pages in the site, the admin interface tracks the navigation and updates appropriately for the page that one navigated to.

Currently, that tracking fails if the link followed includes a fragment (#anchor) (or a query string.)

Here's a fix.  This makes [`Pad.resolve_url`](https://github.com/lektor/lektor/blob/6d7bd4db43584ffa14158ccf5f4f5211a2567f14/lektor/db.py#L1656) ignore any query or fragment on any URL paths passed to it.

### Description of Changes

## Breaking change in `cleanup_path`

This also makes a *conceivably breaking* change in `lektor.utils.cleanup_path` (which is used to normalize Lektor DB paths.)

While mucking about to fix the primary issue, I noticed that `lektor.utils.cleanup_path` currently does not handle double-dots (`..`) in paths in the manner that most of us would expect.
 also makes a *conceivable breaking* change in `lektor.utils.cleanup_path` (which is used to normalize Lektor DB paths.)

Until now, `cleanup_path` was treating path components consisting of two dots (`..`) the same as a single dot and just stripping them from the path.  This is quite different than the normal interpretation of two dots, which indicates traversal to the parent directory.

I've added a commit to this PR which fixes that so that double-dots are now interpreted as a traversal to the parent.

In practice, this change is unlikely to affect many users.  They would notice it if they were doing something like:
```py
pad.get("/parent/child/../child2")
```
which, currently would retrieve the record at `/parent/child/child2`, but after this PR will retrieve `/parent/child2` (as one would naïvely expect in the first place.)

Note that relative URL resolution uses `lektor.utils.join_path` (rather than `cleanup_path`) and `join_path` already handles double-dots in an expected manner.  So things like `{{ "../child2" | url }}` already work as expected (if invoked from `/parent/child`, it will compute the URL to `/parent/child2`). Such uses are unaffected by this PR.

### DIFF
--- a/lektor/db.py
+++ b/lektor/db.py
@@ -35,6 +35,7 @@ from lektor.imagetools import ThumbnailMode
 from lektor.sourceobj import SourceObject
 from lektor.sourceobj import VirtualSourceObject
 from lektor.utils import cleanup_path
+from lektor.utils import cleanup_url_path
 from lektor.utils import fs_enc
 from lektor.utils import locate_executable
 from lektor.utils import make_relative_url
@@ -1660,7 +1661,10 @@ class Pad:
         might be an attachment.  If a record cannot be found or is unexposed
         the return value will be `None`.
         """
-        pieces = clean_path = cleanup_path(url_path).strip("/")
+        try:
+            clean_path = cleanup_url_path(url_path).strip("/")
+        except ValueError:
+            return None
 
         # Split off the alt and if no alt was found, point it to the
         # primary alternative.  If the clean path comes back as `None`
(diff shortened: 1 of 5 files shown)


yagebu (approved):
LGTM :)

> This also makes a conceivably breaking change in lektor.utils.cleanup_path (which is used to normalize Lektor DB paths.) [...] I've added a commit to this PR which fixes that so that double-dots are now interpreted as a traversal to the parent.

This is both a clear bugfix and a "breaking change" - I tend heavily towards the former, I think we can disregard potentiel breakage here :)

Issue #1058, opened 2022-08-31
Server not compatible with marshmallow-dataclass>8.5.3


nixjdm:
marshmallow-dataclass 8.5.3 works.

With marshmallow-dataclass 8.5.4 through 8.5.8, we get

```
$ lektor server
/home/joe/miniconda3/envs/lektor/lib/python3.10/site-packages/marshmallow_dataclass/__init__.py:353: UserWarning: ****** WARNING ****** marshmallow_dataclass was called on the class <function NewType.<locals>.new_type at 0x7f0af2e076d0>, which is not a dataclass. It is going to try and convert the class into a dataclass, which may have undesirable side effects. To avoid this message, make sure all your classes and all the classes of their fields are either explicitly supported by marshmallow_dataclass, or define the schema explicitly using field(metadata=dict(marshmallow_field=...)). For more information, see https://github.com/lovasoa/marshmallow_dataclass/issues/51 ****** WARNING ******
  warnings.warn(
Traceback (most recent call last):
  File "/home/joe/miniconda3/envs/lektor/lib/python3.10/dataclasses.py", line 1197, in fields
    fields = getattr(class_or_instance, _FIELDS)
AttributeError: 'function' object has no attribute '__dataclass_fields__'. Did you mean: '__dataclass_params__'?

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/joe/miniconda3/envs/lektor/lib/python3.10/site-packages/marshmallow_dataclass/__init__.py", line 350, in _internal_class_schema
    fields: Tuple[dataclasses.Field, ...] = dataclasses.fields(clazz)
  File "/home/joe/miniconda3/envs/lektor/lib/python3.10/dataclasses.py", line 1199, in fields
    raise TypeError('must be called with a dataclass type or instance')
TypeError: must be called with a dataclass type or instance

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/joe/miniconda3/envs/lektor/lib/python3.10/site-packages/marshmallow_dataclass/__init__.py", line 364, in _internal_class_schema
    created_dataclass: type = dataclasses.dataclass(clazz)
  File "/home/joe/miniconda3/envs/lektor/lib/python3.10/dataclasses.py", line 1185, in dataclass
    return wrap(cls)
  File "/home/joe/miniconda3/envs/lektor/lib/python3.10/dataclasses.py", line 1176, in wrap
[...]

PR #1061, opened 2022-09-02
Disuse pkg_resources in favor of importlib.metadata

dairiki:
Pkg_resources is deprecated. Importlib.metadata is the new standard way of dealing with distribution metadata: entry points, versions, etc.

This PR eliminates our usage of `pkg_resources` in favor of `importlib.metadata` (or `importlib_metadata` for python < 3.10 — in this PR we use `importlib.metadata.packages_distributions()` which is only available in 3.10.)

### Issue(s) Resolved

This would have sort of fixed the issue noted in [#1058](https://github.com/lektor/lektor/issues/1058#issuecomment-1234619566) having to do with the misspelling of a dependency.  (The real problem there was the misspelling; however, using the same distribution grokking machinery as python itself uses helps minimize weirdness due to differences in behavior between the import system and the introspection system.)

### Description of Changes

### DIFF
--- a/lektor/cli.py
+++ b/lektor/cli.py
@@ -6,7 +6,6 @@ import time
 import warnings
 
 import click
-import pkg_resources
 
 from lektor.cli_utils import AliasedGroup
 from lektor.cli_utils import buildflag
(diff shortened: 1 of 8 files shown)


yagebu (approved):
LGTM :)

PR #326, opened 2016-11-20
Allow extra flags for `lektor deploy`

singingwolfboy:
Replacement for #138. This pull request deprecates the `--build-flag` option, and introduces a new `--extra-flag` option to replace it. This pull request also allows the `deploy` command to also accept extra flags. I've also added some basic tests for the Click CLI to cover this change.

PR #1062, opened 2022-09-02
Remove the deprecated --build-flag option

### LINKED ISSUE #525: Remove the already deprecated build-flag option
This flag was deprecated as of #326. It should actually be removed by the next major release. I'm making this issue to track that.

dairiki:
The `--build-flag` command-line option has been deprecated  since PR #326, which was included Lektor version 3.1.0, released in early 2018.  It's probably time to remove it.

Anyone who is still using `--build-flag` can just switch to using `--extra-flag` instead.

Fixes #525.

### DIFF
--- a/lektor/cli.py
+++ b/lektor/cli.py
@@ -1,5 +1,4 @@
 # pylint: disable=import-outside-toplevel
-import itertools
 import os
 import sys
 import time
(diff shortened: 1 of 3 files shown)


yagebu:
LGTM :)

PR #1064, opened 2022-09-03
ci: use generic lts/* and current specifiers

yagebu:
### Description of Changes

That way, we always test the "current" and LTS release of Node on all
platforms. For completeness we also test the old 14 LTS release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,11 +22,6 @@ jobs:
       fail-fast: false
     steps:
       - uses: actions/checkout@v3
-      - name: Cache pip
-        uses: actions/cache@v3
-        with:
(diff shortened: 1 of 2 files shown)


dairiki:
Do we need to test with 14?

Node is a build tool of ours.  We can specify whatever version we want to be required, right?
(It is nice to support non-bleeding-edge node's for those of us developers who may not want to install it, but I would guess that supporting LTS and current is sufficient.)

Other than that, LGTM!

yagebu:
> Do we need to test with 14?

We don't, I guess. I'll remove it. It's nice if we support older node versions for developers who might only have an older version in their distribution repo. But that's really up to our dependencies and testing it in our CI is probably just a waste of CPU cycles.

PR #1066, opened 2022-09-04
cleanup manifest

yagebu:
### Description of Changes

setuptools_scm only includes files that are tracked in git, so there is no need
to have excludes for files like `*.pyc`. With this we can avoid warnings on
package build.

- manifest: remove unnecessary excludes
- makefile: clean built frontend before re-building


dairiki:
LGTM!

PR #1074, opened 2022-09-21
Enable a default set of mistune2 plugins for feature parity with mistune0

dairiki:
The features provided by the [mistune2 plugins] `url`, `strikethrough`, `footnotes`, and `table` provide features which were supported by mistune 0.8.4 out of the box.

For the sake of backwards-compatibility, since these features were all supported by older versions of Lektor (due to use of mistune 0.x) we should enable these plugins by default when running with mistune2. 

This PR does that.

[mistune2 plugins]: https://mistune.readthedocs.io/en/latest/plugins.html

### Issue(s) Resolved

When using mistune >= 2, markdown features which were previously supported under mistune 0.x are disabled by default.

### Related Issues / Links

This builds upon the work to support mistune 2 in PR #992.

### Description of Changes

### DIFF
--- a/lektor/markdown/controller.py
+++ b/lektor/markdown/controller.py
@@ -127,6 +127,10 @@ class RendererHelper:
         )
 
 
+class UnknownPluginError(LookupError):
+    """Exception raised when a mistune2 plugin name can not be resolved."""
+
+
 class RenderResult(NamedTuple):
     html: str
     meta: Meta
--- a/lektor/markdown/mistune2.py
+++ b/lektor/markdown/mistune2.py
@@ -1,4 +1,6 @@
 """MarkdownController implementation for mistune 2.x"""
+from __future__ import annotations
+
 import sys
 from typing import Any
 from typing import Callable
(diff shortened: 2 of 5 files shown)


nixjdm:
This is great.

Also, the test is the only Lektor plugin I can recall that uses `on_markdown_lexer_config`. I think it's worth adding a mention of it to the docs https://www.getlektor.com/docs/api/plugins/events/markdown-lexer-config/

dairiki:
> Also, the test is the only Lektor plugin I can recall that uses `on_markdown_lexer_config`. I think it's worth adding a mention of it to the docs https://www.getlektor.com/docs/api/plugins/events/markdown-lexer-config/

The test could just as well have used `on_markdown_config` in this case.

I'm not quite sure myself what the exact purpose of `on_markdown_lexer_config` is. The only real distinction distinction between the two is that `on_markdown_lexer_config` has access to the _renderer_ instance, while `on_markdown_config` does not (since it is called before the renderer is constructed).

nixjdm:
> The test could just as well have used on_markdown_config in this case.

Ah, I see. Yeah, then I can't recall ever seeing a use of `on_markdown_lexer_config`.

I suppose someone could modify the render after it's created to add methods to it after creation. Perhaps that might seem more natural than customizing the markdown config and improved renderer that we have. That seems less explicit though, and maybe should be discouraged, in favor of supplying an explicitly made renderer to the config. Also off topic for this PR I suppose.

PR #1079, opened 2022-10-06
Added missing dependency to README

miratcan:
I have started developing lektor on my local computer. When I follow the readme I found that imagemagick is missing from dependency list. Also, I have changed "vitualenv" command to "python -m  venv" which is new way to handle virtual environments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,13 +28,20 @@ For installation instructions head to the official documentation:
 
 ## Want to develop on Lektor?
 
-This gets you started (assuming you have Python, pip, Make and pre-commit
-installed):
+These must be installed on your computer:
+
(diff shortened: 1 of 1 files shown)


nixjdm:
I think including imagemagick is fair, and by the same token ffmpeg. Both are listed https://www.getlektor.com/docs/installation/, but they could do with a readme mention. Thank you for the feedback.

PR #1057 touches on a handful of readme updates and related things. I need to respond to and update that one, and I think it's probably best we close this PR in favor of updating that one to have these related changes in one place. I'll get on that soon. Again, thank you though! Sometimes I even forget about these dependencies on a new system until a build fails.

PR #1082, opened 2022-11-02
Packaging: Convert from setuptools to hatch

### LINKED ISSUE #1081: [Build] Missing frontend assets when building from the source
Currently, the frontend assets are missing if we install from a Git URL or zipball provided by GitHub. This is because we build the frontend files in a Makefile and it is called when publishing to PyPI.

~~Plus, are you open to switching to [hatch](https://hatch.pypa.io) as the build backend?~~
Oh, you are already doing it!

dairiki:
I started working on packaging Lektor using [hatch] a few weeks ago.

The use of `hatch` allows for:

- Building the frontend javascript and css as part of the PEP517 build process using a [build hook].  In addition to simplifying the distribution building process by eliminating the `make build-js` step, this allows Lektor to be installed directly from a git repo. 
- Management of development (and, potentially, test) virtual environments.

[hatch]: https://hatch.pypa.io/latest/
[build hook]: https://hatch.pypa.io/latest/config/build/#build-hooks

### Issue(s) Resolved

Fixes #1081
Fixes #1057

### Description of Changes

(This work probably needs a bit of freshening.  I'm creating this PR now since it relates to the newly created #1081.)

All or most of our `tox` tests could probably be converted to `hatch` scripts.  At present, however, hatch does not support running scripts in parallel (e.g. there is no equivalent of `tox -p`) so running the full set of tests via hatch is painfully slow.


.github/workflows/publish.yml
```diff
@@ -13,16 +15,16 @@ jobs:
...
         with:
           node-version: "lts/*"
-      - name: Build frontend
-        run: make
       - name: Install dependencies
-        run: python -m pip install tox twine
-      - name: Build release artefacts
-        run: tox -e build-dist
+        run: python -m pip install hatch
+      - run: hatch build
```
frostming, line 25: Keep it simple by using `build`?

.github/workflows/ci.yml
```diff
@@ -129,8 +136,10 @@ jobs:
       - name: Install python dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install tox tox-gh-actions coverage[toml]
+          python -m pip install hatch
       - name: Run python tests
-        run: tox
+        run: hatch run +mistune +python=${{ matrix.python }} test:test-python
```
frostming, line 145: It seems `tox.ini` isn't needed anymore.

dairiki: The reason I hesitate to get rid of tox is that hatch currently doesn't have any equivalent of `tox -p` — that is, it doesn't yet have the ability to run the various test environments in parallel (though, reportedly, that feature is planned). For the CI tests, it doesn't matter, but it does matter for the dev running the tests on their workstation.
And if we're going to keep the tox tests for that, I figure we should use them in CI.

dairiki:
Closing in favor of #1112, which includes just the hatchling PEP 512 builder bits from this PR.

I don't think that using hatch to manage test and dev environments buys us much.

Our test requirements are complex enough that I think it's better we stick with tox for that.

For building distributions, however, hatchling is a flexible PEP-621build backend that is customizable enough that we can hook the esbuild-ing of our backend code into the PEP-512 build process, which is a big win.

PR #1093, opened 2022-11-23
Fix: order_by to respect alt

### LINKED ISSUE #1092: `query` with alternative records dependency on different alt
Lets say, I use `site.query('/path', this.alt)|list` in a template. This will not only include the records of the provided alt, but also the records for the default alt.

Assuming the default alt is EN:

- `layout-en.html` with `query('/pages', alt='en')` tracks dependency on `page1-en.html` and `page2-en.html`.
- `layout-de.html` with `query('/pages', alt='de')` tracks dependency on `page1-en.html`, `page2-en.html`, `page1-de.html`, and `page2-de.html`.

relikd:
Add missing alt in pad query.

### Issue(s) Resolved

Fixes #1092 

### Description of Changes

### DIFF
--- a/lektor/db.py
+++ b/lektor/db.py
@@ -763,7 +763,9 @@ class Attachment(Record):
     def parent(self):
         """The associated record for this attachment."""
         return self.pad.get(
-            self._data["_attachment_for"], persist=self.pad.cache.is_persistent(self)
+            self._data["_attachment_for"],
+            alt=self.alt,
+            persist=self.pad.cache.is_persistent(self)
         )
 
     @cached_property
--- a/lektor/db.py
+++ b/lektor/db.py
@@ -1078,7 +1078,7 @@ class Query:
         """Returns the order that should be used."""
         if self._order_by is not None:
             return self._order_by
-        base_record = self.pad.get(self.path, self.alt)
+        base_record = self.pad.get(self.path, alt=self.alt)
         if base_record is not None:
             if self._include_attachments and not self._include_pages:
                 return base_record.datamodel.attachment_config.order_by
--- a/lektor/db.py
+++ b/lektor/db.py
@@ -1078,7 +1078,7 @@ class Query:
         """Returns the order that should be used."""
         if self._order_by is not None:
             return self._order_by
-        base_record = self.pad.get(self.path)
+        base_record = self.pad.get(self.path, self.alt)
         if base_record is not None:
             if self._include_attachments and not self._include_pages:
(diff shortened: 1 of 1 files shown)


lektor/db.py
```diff
@@ -1078,7 +1078,7 @@ def get_order_by(self):
         """Returns the order that should be used."""
         if self._order_by is not None:
             return self._order_by
-        base_record = self.pad.get(self.path)
+        base_record = self.pad.get(self.path, alt=self.alt)
```
dairiki, line 1083: It looks like, in this case, we are interested only in `base_record.datamodel`.  The datamodel does not depend on any of the record's `.lr` files (regardless of alt).  Perhaps we should figure out a way that we end up only declaring a dependency on the datamodel `.ini` file (and and `.ini` files of datamodels from which it inherits)?

More investigation is probably required to determine whether such an optimization would buy us much (in terms of reducing unnecessary recorded dependencies, and thus reducing unnecessary rebuilds).

dairiki: Perhaps this is related to part of PR #1007, where dependency tracking is done in Record.__getitem__.

Moving dependency tracking to attribute accessors (of both Records and DataModels) rather than recording the dependency upon record traversal would fix this...

relikd: Currently this is an issue for all queries with non-default alt. `self.pad.get` will access the primary alt in that case and record a dependency. I would be happy if, in this case, the datamodel could be accessed without requesting the record and thus avoid a dependency. But for the time being (if an actual fix will take too long) I would appreciate if the alt would be set as a hotfix. This at least prevents unnecessary rebuilds of alternatives if a primary alt page changes.

And dependency tracking via attributes would be superb. There is another issue from 2016 I think where this was proposed. So dependency tracking could finally be applied to individual fields rather than files.

dairiki: > And dependency tracking via attributes would be superb. There is another issue from 2016 I think where this was proposed. So dependency tracking could finally be applied to individual fields rather than files.

Yes, that would be nice, but is probably a lot of work, and is not quite what I'm referring to.

My point is that just because a record is fetched (via `Pad.get`) during the building of an artifact, that does not mean, necessarily, that that artifact depends on that record's source files.  It might be better to defer the recording of the dependency until such time as one of the record's fields is accessed, since that does imply a dependency on the source file.

relikd:
I've seen in some release notes that this may have been fixed? Should I close the PR?

Issue #774, opened 2020-05-03
Remove obsolete desktop app code


xlotlu:
I propose we remove `/gui` entirely from the code base. We could `git filter-branch` it to a new repo, mostly for inspirational purposes, I doubt someone will reuse the code.

Right now it's dead weight, its only purpose to give the bot a reason to annoy us. #737 #738 #771.

yagebu:
:+1: from me.

> We could git filter-branch it to a new repo

I think having it still in the history (and available with `git checkout 3.1.3` for example) should be enough (and hide it a bit better so that no one will get the impression that it's working)

runfalk:
:+1: from me as well.

I agree that checking out an older version should be enough. I don't think there is that much left that we want to reuse anyway.

PR #1099, opened 2023-02-19
Add webp as image format in default config

kno10:
Webp should be added, and rather fix the OSX build then?

### DIFF
--- a/lektor/environment/config.py
+++ b/lektor/environment/config.py
@@ -27,6 +27,7 @@ DEFAULT_CONFIG = {
         ".jpg": "image",
         ".jpeg": "image",
         ".png": "image",
+        ".webp": "image",
         ".gif": "image",
         ".svg": "image",
         ".avi": "video",


dairiki:
Note that Lektor's thumbnail support currently doesn't work with *webp* images.  (Lektor’s [get_image_info](https://github.com/lektor/lektor/blob/74bce9096116caffe27ab69327ed4ba274e66e1e/lektor/imagetools.py#L343) doesn't know how to recognize and deduce the dimensions of *webp* images, so the [Image.thumbnail](https://github.com/lektor/lektor/blob/74bce9096116caffe27ab69327ed4ba274e66e1e/lektor/db.py#L835) fails with `RuntimeError("Cannot process unknown images")`.)

It seems the main reason for recognizing a file type as an image is to have it wrapped in an `Image` rather than just an `Attachment`.  Until `Image` knows how to deal with it, it is probably premature to mark it as an image.

It would be great to support thumbnailing of *webp* images.
That said, I am no expert, but webp supports more knobs and options than other image formats — lossy or not, how to deal with alpha, animations, etc.  It might take some careful thought to figure out how best to implement webp thumbnailing in a reasonably general but not too confusing manner.

----

What about OSX?  I know (I think) that at one time there was a macOS single-file executable released of Lektor, but that has not been the case for quite some time.  (#774)

kno10:
As the thumbnailing functionality relies on imagemagick anyway, why not also use imagemagick to get the image size?

Or python-magic, as `file` can also identify the image size, and that is exactly the functionality that is replicated here?

kno10:
Relevant part for reading the size of a typical "VP8 " file (lossy) - it seems that this should work similar to the other file formats by checking some header bytes, then reading at the appropriate offset:
https://github.com/file/file/blob/1c9f670f69bd508da2b4d05a28d27bd92407f2c8/magic/Magdir/riff#L143-L148
Note sure of VP8L and VP8X are different. From a quick check, file would not report the size of a lossless VP8L image, imagemagic does.

dairiki:
> As the thumbnailing functionality relies on imagemagick anyway, why not also use imagemagick to get the image size?

We would like `Image.width`, `Image.height` to be fast and available even if `imagemagick` is not installed.
(`Image.exif`, too, but we use [exifread](https://github.com/ianare/exif-py) for that. I think exifread already supports webp.)

> Or python-magic, as `file` can also identify the image size, and that is exactly the functionality that is replicated here?

`Python-magic` relies on the *libmagic* library, so not necessarily available everywhere.  (In particular, I suspect that using python-magic on Windows might be an issue, but I don't know.)

----

A quick PyPI search found [pymage-size](https://github.com/kobaltcore/pymage_size), a pure-python image format and size detection library that claims to do what we need.  (I haven't tried it or vetted it carefully.)  There may be others.

[ETA: Further searching has found [imagesize](https://github.com/shibukawa/imagesize_py), another pure-python lib which claims to do what we want.  *Imagesize* seems to be in more [broad usage](https://libraries.io/pypi/imagesize/dependents) and supports a wider range of Python versions that *pymage-size*.]

Big picture (bigger project): we may want to consider using [Pillow](https://github.com/python-pillow/Pillow) instead of imagemagick for thumbnailing.  Pillow provides binary wheels for a large range of target architectures.

dairiki:
> Relevant part for reading the size of a typical "VP8 " file (lossy) - it seems that this should work similar to the other file formats by checking some header bytes, then reading at the appropriate offset:
> https://github.com/file/file/blob/1c9f670f69bd508da2b4d05a28d27bd92407f2c8/magic/Magdir/riff#L143-L148
> Note sure of VP8L and VP8X are different. From a quick check, file would not report the size of a lossless VP8L image, imagemagic does.

Here's a more general top-level spec for the WEBP format https://developers.google.com/speed/webp/docs/riff_container

dairiki:
For reference, some test images: https://github.com/webmproject/libwebp-test-data

etoimene:
@dairiki 
Looks like some of this has been done with https://github.com/lektor/lektor/pull/1196

dairiki:
Closing in favor of #1196

PR #1107, opened 2023-02-27
Remove BuildState.make_named_temporary method

dairiki:
This PR deletes the [BuildState.make_named_temporary](https://github.com/lektor/lektor/blob/91e205cdbf8f4694ad6fe2ce0e350e6f87314bca/lektor/builder.py#L134) method.
The method was unused by Lektor, and, anyway, was broken so that it was unusable by anyone else.  (Its [`os.urandom(20).encode("hex")`](https://github.com/lektor/lektor/blob/91e205cdbf8f4694ad6fe2ce0e350e6f87314bca/lektor/builder.py#L146) throws `"AttributeError: 'bytes' object has no attribute 'encode'"`.)

### Potentially Breaking Change

This is a potentially breaking change, though I think the likelihood of breaking external code is small enough not to worry about it.

This PR removes the [context-manager API bits](https://github.com/lektor/lektor/blob/91e205cdbf8f4694ad6fe2ce0e350e6f87314bca/lektor/builder.py#L113-L124) from `BuildState` — their only purpose was to clean up the temporary files (that could never be created).  This is potentially a *breaking change*, however, it breaks none of our existing tests, and I don't *think* plugins should be messing about with *creating* (and context-managing) `BuildState` instances.

If anyone feels that this is an issue, we could leave stub `__enter__` and `__exit__` methods in place (and make `__enter__` issue a deprecation warning).

### Issue(s) Resolved

Removes dead code.

### DIFF
--- a/lektor/builder.py
+++ b/lektor/builder.py
@@ -90,7 +90,6 @@ class BuildState:
     def __init__(self, builder, path_cache):
         self.builder = builder
 
-        self.named_temporaries = set()
         self.updated_artifacts = []
         self.failed_artifacts = []
         self.path_cache = path_cache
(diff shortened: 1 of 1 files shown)


yagebu (approved):
LGTM, nice cleanup :)

Issue #1109, opened 2023-02-28
Installation doesn't install required pytz


bockstaller:
Installing lektor via pipx and pip doesn't install the required dependency pytz.
It is imported in the code and in lektors setup.cfg, but as far as I can tell it isn't included in the wheel or the metadata of the source distribution.

ZakiChammaa:
I have the same problem. It was working fine yesterday.

dairiki:
Thank you for the report.

I believe this only affects the 3.3 (stable) branch, where we (erroneously) do not list `pytz` as a dependency.

Up until recently, the `babel` package depended unconditionally on `pytz`, so it was being installed transitively.  Since `babel==2.12.0` (published 9 hours ago), `pytz` is only a conditional dependency (python-babel/babel#940) so only gets pulled in for python<3.9.

dairiki:
Fixed in [release 3.3.8](https://github.com/lektor/lektor/releases/tag/v3.3.8).

CAM-Gerlach:
(For reference: The previous fix for 3.4.x was PR #1036)

dairiki:
> (For reference: The previous fix for 3.4.x was PR #1036)

@CAM-Gerlach Unfortunately, that PR never made it into the 3.3-maintenance branch.  And, thank you for pointing that out, because now I see that the rest of that PR did not make it into 3.3-maintenance either. :-/

CAM-Gerlach:
Yup, that's what I figured. (BTW, FWIW noticed this as part of releasing the updated Lektor 3.3.8 Conda-Forge packages in conda-forge/lektor-feedstock#23 ).

PR #1110, opened 2023-02-28
Remove dependency on pytz

dairiki:
Since the introduction of <del>PEP 614</del> PEP 615 (the `zoneinfo` module in the standard library, available since python 3.9), the future, apparently, is to transition away from `pytz` to `zoneinfo` for timezone support.

Babel, since 2.12, will now work with either `pytz` or `zoneinfo`, and no longer requires `pytz` on pythons >= 3.9.

This change also fixes some hokey-ness in the previous implementation of DateTimeType.

### Related Issues / Links

See #1109.

### Description of Changes

### DIFF
--- a/lektor/types/primitives.py
+++ b/lektor/types/primitives.py
@@ -1,9 +1,9 @@
+import re
 from datetime import date
 from datetime import datetime
 
 from babel.dates import get_timezone
 from markupsafe import Markup
-from pytz import FixedOffset
 
 from lektor.constants import PRIMARY_ALT
 from lektor.i18n import get_i18n_block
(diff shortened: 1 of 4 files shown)


dairiki:
Hmmph.  [Apparently](https://docs.python.org/3/library/zoneinfo.html#data-sources)

> Some systems, including notably Windows systems, do not have an IANA database available, and so for projects targeting cross-platform compatibility that require time zone data, it is recommended to declare a dependency on tzdata.

I've added `pytz` back to our dependency list, but only on Windows.  Perhaps, for uniformity, we need to keep it for all platforms?  OTOH, it's not the end of the world if a user has to manually install `pytz` if their system doesn't provide a suitable timezone database and they want to use named timezones in their `datetime` fields.

CAM-Gerlach:
Hmm; is there a reason to not just add a dependency on [`tzdata`](https://pypi.org/project/tzdata/) instead, as the message indicates? That just ships the time zone db that `zoneinfo` needs rather than all of `pytz`.

dairiki:
> Hmm; is there a reason to not just add a dependency on [`tzdata`](https://pypi.org/project/tzdata/) instead, as the message indicates?

Because I can't read.  Thank you.

PR #1128, opened 2023-04-16
Back-port #1065 to the 3.3.x branch

### LINKED ISSUE #1127: pip>=23.1 breaks the installation of local plugins in Lektor 3.3.8
The latest release of `pip` (23.1) [removes support](https://pip.pypa.io/en/stable/news/#v23-1) for the `--install-option` parameter to `pip install`, which [is used](https://github.com/lektor/lektor/blob/5745acc1a112b59d3d041abcf49611f8319f016d/lektor/packages.py#L119) by our `install_local_package` function.

Pinning `pip` does not completely fix the issue, since the latest version of pip may still end up being installed in Lektor's private package cache (and, if/when that happens, that version of pip will be used by `install_local_package`, since the package cache is in `sys.path` at that point).

A possible fix would be to backport #1065 to the 3.3 branch.  I am working up a PR that does that.

If there were a simpler non-intrusive fix (*a la* pinning pip) that would probably be a better option...

dairiki:
This PR back-ports #1065 (and #1061) to the 3.3-maintenance branch.

### Issue(s) Resolved

Fixes #1127

This is one possible approach to fixing the breakage caused by the latest release of `pip` (which no longer supports the `--install-option` flag to `pip install`).

### Description of Changes

### DIFF
--- a/lektor/cli.py
+++ b/lektor/cli.py
@@ -6,7 +6,6 @@ import time
 import warnings
 
 import click
-import pkg_resources
 
 from lektor.cli_utils import AliasedGroup
 from lektor.cli_utils import buildflag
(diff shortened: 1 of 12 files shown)


yagebu (approved):

dairiki:
As it turns out, recent versions of `pip` appear to support the use of `--target` in conjunction with `--editable`.

PR #1129 presents an alternative, more minimal, fix for #1127 which uses that fact.

dairiki:
Rejected in favor of #1129.  Closing.

PR #1130, opened 2023-04-16
frontend: reduce size; use @fontsource

### LINKED ISSUE #1115: Can we reduce the size of our compiled frontend code?
In working on #1112, I noticed that our distribution is not small.  The wheel is currently 1.3Mb zipped — 3.5Mb unzipped.

A large part of that size (2.7M, unzipped) comes from the compiled frontend code:

```sh
$ ls -s1h lektor/admin/static
total 2.7M
184K app.css
240K app.css.map
480K app.js
628K app.js.map
436K fontawesome-webfont-3KIJVIEY.svg
 76K fontawesome-webfont-5GKVPAEF.woff2
164K fontawesome-webfont-FMJ3VJ65.eot
164K fontawesome-webfont-RJ6LE7IU.ttf
 96K fontawesome-webfont-Z4ARLA73.woff
 72K roboto-slab-all-300-CD7CKZ7B.woff
 56K roboto-slab-all-300-ZPJL4QTS.woff2
 56K roboto-slab-all-400-26JDMZKQ.woff2
 72K roboto-slab-all-400-JK3TGTJI.woff
```

Questions (mostly for @yagebu):

We probably don't need the SVG, EOT, or TTF fontawesome fonts, though I am unsure of the easiest way to remove them from our CSS.

Do we need the full sourcemaps, or case we disable [sourcesContent](https://esbuild.github.io/api/#sources-content) for production?

Should we turn up the minimize knobs more for production?  Currently, we enable only [minimizeSyntax](https://esbuild.github.io/api/#minify).  Should we enable `minimizeIdentifiers` and `minimizeWhitespace` as well? (I.e. just `minimize: !dev`.)

Is there anything else we can do to reduce the size?

(Doing all of the above reduces the unzipped size of `lektor/admin/static/` from 2.7M to 1.2M, and the total size of the zipped wheel from 1.3M to 812K.)

yagebu:
### Issue(s) Resolved

Fixes #1115

### Description of Changes

Only include woff2 fonts and do not include full source code in
sourcemaps.

The npm packages were one best gets fonts seem to change every couple of
years, @fontsource seems to be a good option now, @openfonts seems to
have been abandoned.

Previously, we also included font weights 300 and 400 of Roboto Slab but
only 400 and 700 were used - fix that to include weights 400 and 700
now.


frontend/build.ts
```diff
@@ -39,14 +39,17 @@ async function runBuild(dev: boolean) {
     format: "iife",
     bundle: true,
     loader: {
-      ".eot": "file",
-      ".ttf": "file",
-      ".svg": "file",
-      ".woff": "file",
+      ".eot": "empty",
```
dairiki, line 42: Very clever!

It does result in CSS with a bunch of empty `url()`s like:
```css
/* node_modules/font-awesome/css/font-awesome.css */
@font-face {
  font-family: FontAwesome;
  src: url();
  src:
    url() format("embedded-opentype"),
    url(./fontawesome-webfont-5GKVPAEF.woff2?v=4.7.0) format("woff2"),
    url() format("woff"),
    url() format("truetype"),
    url() format("svg");
  font-weight: 400;
  font-style: normal;
}
```

That looks a bit weird, but doesn't seem to cause trouble...

dairiki:
LGTM!

PR #1136, opened 2023-04-22
refactor: Disuse `watchdog` in favor of `watchfiles`.

### LINKED ISSUE #550: Even read only git commands trigger build with `lektor serve`
If the `.git directory` is in the site root, every git command triggers a build.

This is annoying when running `lektor serve` in the background in the same terminal when doing these git operations.

Lektor should have a way to ignore changes to git's internal state(i.e. lockfiles, etc).

### LINKED ISSUE #1116: Infinite build-loop in watch mode with output in project directory (sometimes).
This applies both to the current master branch (e.g. 3.4.0b5) as well as the 3.3 branch (and probably earlier).

Currently, if you run
```sh
lektor --project example server -O _htdocs
```
An initial build-all cycle will run.  (Here, we assume that we're in the Lektor source directory, so that `example` refers to the example project directory — but I think it can be any project directory that is not the current working directory.)

Now edit a source file. Any edit will do, but one that causes a build error will really do it.  In another terminal:
```sh
echo "{{ badfunc() }}" >> example/templates/layout.html
```

That will kick off another build-all.  And then, usually, another build-all.  And then another...
It will keep going for some time.  Usually, it stops building, eventually. (I don't really understand  why it ever stops.)

The root cause is that Lektor's filesystem watcher, in this case, is not correctly ignoring changes to files in the output directory.  The output directory path is interpreted (by the `Builder`) relative to the project root, while the `Watcher` is interpreting it relative to the current working directory.  (This is somewhat related to #1103.)

Passing an absolute path for the output directory works around the issue.  This does not suffer the same problem:
```sh
lektor --project example server -O $PWD/example/_htdocs
```

----

The situation is worse when running `lektor build --watch` — in that case, the `Watcher` doesn't even get passed the output path at all.  So, in that case, quasi-infinite build loops result if the output directory is placed within the project root, no matter whether specified as an absolute or relative path.

Try:
```sh
lektor --project example build --watch -O _htdocs
```
Then, in another terminal:
```sh
echo "x" >> example/content/contents.lr
```

dairiki:
[Watchfiles](https://github.com/samuelcolvin/watchfiles) has a much cleaner API.
It has a (Rust-based) extension module, but wheels are distributed for a [wide variety](https://pypi.org/project/watchfiles/#files) of targets.

Also, it's a flail, but I hope it might help with the hanging CI tests (see #1132) on macOS (which seem to be related to watchdog.)

### Issue(s) Resolved

It simplifies our code quite a bit.

It might fix the CI test hangage on macOS.

Fixes #550
Fixes #1116

### Related Issues / Links

#1132 
#861

### Description of Changes

### DIFF
--- a/lektor/cli.py
+++ b/lektor/cli.py
@@ -2,7 +2,7 @@
 import os
 import sys
 import warnings
-from contextlib import suppress
+from itertools import chain
 
 import click
 
@@ -151,17 +151,19 @@ def build_cmd(
 
     reporter = CliReporter(env, verbosity=verbosity)
     with reporter:
-        success = _build()
-        if not watch:
-            return sys.exit(0 if success else 1)
+        builds = ["first"]
+        if watch:
+            from lektor.watcher import watch_project
 
-        from lektor.watcher import Watcher
+            click.secho("Watching for file system changes", fg="cyan")
+            builds = chain(
+                builds, watch_project(env, output_path, raise_interrupt=False)
+            )
 
-        click.secho("Watching for file system changes", fg="cyan")
-        with Watcher(env) as watcher, suppress(KeyboardInterrupt):
-            while True:
-                watcher.wait()
-                _build()
+        success = False
+        for _ in builds:
+            success = _build()
+        return sys.exit(0 if success else 1)
 
 
 @cli.command("clean")
--- a/lektor/cli.py
+++ b/lektor/cli.py
@@ -130,27 +130,7 @@ def build_cmd(
 
     env = ctx.get_env()
 
-    def _build():
-        builder = Builder(
-            env.new_pad(),
-            output_path,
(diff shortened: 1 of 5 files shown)


nixjdm:
Might this enable closing out #550 easily?

dairiki:
> Might this enable closing out #550 easily?

Yes.  Actually, I think it (inadvertently) fixes that already.

I chose to have our [WatchFilter](https://github.com/lektor/lektor/blob/3273d0c7e44da281d95691a3172beeddef7efcd6/lektor/watcher.py#L24) inherit from watchfiles' `DefaultFilter`.  DefaultFilter is [already set up](https://watchfiles.helpmanual.io/api/filters/#watchfiles.filters.DefaultFilter.ignore_dirs) to ignore things like `.git`, `node_modules`.

PR #1137, opened 2023-04-22
refactor!: remove the --profile option from the build command

dairiki:
Running `lektor build --profile` just runs `Builder.build_all` under `cProfile`, then prints the stats.

The profile information is only really interesting to Lektor developers.  Lektor can be profiled directly like this:

    python -m cProfile -o prof.out -m lektor build
    python -m pstats prof.out


yagebu (approved):
LGTM :)

PR #1138, opened 2023-04-22
Refactor: Use Pillow for image info

dairiki:
This replaces our homegrown code in [get_image_info()](https://github.com/lektor/lektor/blob/b843775f9480295c916f9c933e290f760dbbe531/lektor/imagetools.py#L392) code, and the [exifread](https://pypi.org/project/ExifRead/)-based based code in [exif_info()](https://github.com/lektor/lektor/blob/b843775f9480295c916f9c933e290f760dbbe531/lektor/imagetools.py#L474) with Pillow-based code.

This eliminates two dependencies (on exifread and filetype) and will make it much easier to support other image formats (e.g. WEBP) going forward.

It also guarantees that get_image_info and make_image_thumbnail will agree on things like image type and what images are thumbnailable.

I've done some rudimentary profiling.  Currently, the new Pillow-based version of get_image_info is a bit slower (maybe by 30-50%) than the current one.  (Getting the image dimensions is very fast, the slower bit is reading the Exif tags to determine whether height and width need to be swapped.)

### Issue(s) Resolved

Fixes #

### Related Issues / Links

#1099

### Description of Changes


yagebu:
Nice one, thanks :)

PR #1145, opened 2023-05-18
Fix: "FATAL: exception not rethrown" from `lektor server`

dairiki:
### Issue(s) Resolved

When `lektor server` is terminated (e.g. via SIGINT) a "FATAL: exception not rethrown" error is displayed and a core dump is attempted.

This was introduced by #1136. The error arises when `watchfiles.watcher` is run in a daemon thread. When the thread is killed (when the main thread terminates) something is not getting cleaned up properly. The solution taken here is to run the watcher in a non-daemon thread, making sure to shut down the thread on exit.

### Related Issues / Links

- samuelcolvin/watchfiles#128, samuelcolvin/watchfiles#132

### Description of Changes

I've also cleaned up the code in `lektor.devserver` a bit.

### DIFF
--- a/lektor/admin/context.py
+++ b/lektor/admin/context.py
@@ -1,9 +1,8 @@
+from __future__ import annotations
+
 from typing import Any
 from typing import NamedTuple
-from typing import Optional
-from typing import Sequence
 from typing import TYPE_CHECKING
-from typing import Union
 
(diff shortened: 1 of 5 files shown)


yagebu (approved):
Nice cleanup as well :)

PR #1147, opened 2023-06-03
Fix #908: Omit shadowed assets from build

### LINKED ISSUE #908: Custom css file inconsistent behavior during deployment
Currently I use the [Resumé](https://github.com/hannylicious/lektor-theme-resume/tree/e2c32d5a4ab877fa6029a0b2e3edb13d6425afab) theme. I created a copy of the `resume.min.css` file in the `assets/css` folder. Here I change a color value for a certain property. When running locally `lektor server` the changed behavior of the color is visible in the `/admin/` interface. But when deploying locally to a folder the css color from the theme is used. Somehow my own values are being override by the theme value during build. 

The curious thing is that when I _remove_ the property theme property is not used. So it's not entirely inconsistent. I did not worked out the cases for other changes.

### LINKED ISSUE #1111: Buglet with lowercasing of asset file extensions
This is not a real important bug, but I'm writing this down for posterity.  There's an xfail-marked test in the test suite for this, and I keep forgetting the details about it, then spending time sussing it out.  Here are the details:

As things currently stand, when Lektor installs *assets* (by copying) to the output directory, if (and only if) the filename extension of the asset matches one of the known extensions in `config["ATTACHMENT_TYPES"]`, [the extension is lowercased](https://github.com/lektor/lektor/blob/26e700e62b3c02a18761cfd7cc7f274ee171dd89/lektor/assets.py#L52-L53) when generating the output filename (i.e. the target `url_path`.)

However, `Asset.resolve_url` when looking up an *asset* by its URL, does not account for this (possible) mangling of the file extension.  I.e. an asset that gets installed to `/Asset.txt` may have come from a source file named any of `/Asset.txt`, `/Asset.TXT`, `/Asset.tXt`, ... — currently only the first of those will be found by `Asset.resolve_url_path`.

There is an *xfailing* test in `test_assets.py` that exercises this: [these](https://github.com/lektor/lektor/blob/26e700e62b3c02a18761cfd7cc7f274ee171dd89/tests/test_assets.py#L46) [tests](https://github.com/lektor/lektor/blob/26e700e62b3c02a18761cfd7cc7f274ee171dd89/tests/test_assets.py#L59) demonstrate that `TEST.TXT` gets installed to `TEST.txt`. [This xfailing test](https://github.com/lektor/lektor/blob/26e700e62b3c02a18761cfd7cc7f274ee171dd89/tests/test_assets.py#L118) shows that `resolve_url_path` doesn't find it.

### Why is Lektor mucking with the asset extensions at all?
I understand that many web servers may be set up to send a `Content-Type` base of the file extension and that those web servers may possibly be confused by, e.g., an upper-cased extension.  But that's not really our bailiwick.  If the user wants an attachment with a specifically cased extension, let him create one.

For the sake of matching attachments against our `config["ATTACHMENT_TYPES"]`, we can do that in a case-insensitive manner.

### If we do really want to lowercase some file extensions (why not all of them?), then what do we do about the name conflicts that can arise?

E.g. suppose there are two assets named `text.txt` and `text.TXT`.  Currently, Lektor will install both to the same place, with one overwriting the other.  Which one wins is, I think, indeterminate.

dairiki:
### Issue(s) Resolved

Fixes #908
Fixes #1111 

### Description of Changes

The issue described in #908 happens when themes (and so multiple asset trees) are in play.  If there is an asset in one asset tree at the same path as an asset in a second asset tree, both are currently built.  Besides being inefficient (one overwrites the other) the order of building seems not to be completely deterministic, so that it's not clear which version wins.

Currently, when themes are in play, so that there are multiple on-disk directory trees of assets to be merged to a single output tree, Lektor keeps a list of the multiple assets tree roots, and traverses each of those separately to find all the assets to be built.

In this PR we refactor `lektor.assets.Directory` so that it can represent a logically merged asset directory.  After this PR a single `lektor.assets.Directory` instance can represent an overlayed set of directories.  This gets us back to having a single *asset root* which represents the merged asset tree.

### DIFF
--- a/lektor/admin/modules/serve.py
+++ b/lektor/admin/modules/serve.py
@@ -138,7 +138,7 @@ class ArtifactServer:
         Raise NotFound if no index is found.
         """
         for name in "index.html", "index.htm":
-            index = directory.get_child(name, from_url=True)
+            index = directory.resolve_url_path([name])
             if index is not None:
                 break
         else:
--- a/lektor/assets.py
+++ b/lektor/assets.py
@@ -92,10 +92,6 @@ class Asset(SourceObject):
             return self.parent.artifact_name.rstrip("/") + "/" + self.url_name
         return self.url_path
 
-    def build_asset(self, f) -> None:
-        # FIXME: unused? delete?
-        pass
-
     @property
     def children(self) -> Iterable[Asset]:
         return ()
--- a/lektor/assets.py
+++ b/lektor/assets.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import os
 import posixpath
+import warnings
 from contextlib import suppress
 from itertools import takewhile
 from operator import methodcaller
--- a/lektor/assets.py
+++ b/lektor/assets.py
@@ -1,48 +1,66 @@
+from __future__ import annotations
+
 import os
 import posixpath
-import stat
+from contextlib import suppress
+from itertools import takewhile
+from operator import methodcaller
+from pathlib import Path
+from pathlib import PurePosixPath
+from typing import Generator
+from typing import Iterable
+from typing import Sequence
+from typing import TYPE_CHECKING
 
 from lektor.sourceobj import SourceObject
+from lektor.utils import deprecated
+
+if TYPE_CHECKING:
+    from _typeshed import StrPath
+    from lektor.db import Pad
 
 
-def get_asset(pad, filename, parent=None):
-    env = pad.db.env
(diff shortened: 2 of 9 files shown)


dairiki:
I <del>am going to</del> <ins>have now</ins> refactor<ins>ed</ins> this.

Having slept on it, I think the right solution is to prevent the asset tree as a single merged tree containing the directories/files accessible via any of the asset directories.

This <del>will be</del> <ins>was</ins> a significant refactor of `lektor.assets`, but <del>will solve</del> <ins>solves</ins> the problem of `Directory` assets having children that should be ignored.

PR #1167, opened 2023-09-19
Test under python 3.12

dairiki:
Python 3.12rc3 — the final release candidate — is out.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,7 +89,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        python: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12-dev"]
         exclude:
           - os: "macos-latest"
             python: "3.8"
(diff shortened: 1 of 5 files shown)


dairiki:
See python/cpython#109590 for an explanation of the failure under python 3.12 on Windows.

This is happening during the package build stage, here
https://github.com/lektor/lektor/blob/efd15bc4fac35a3640b6534fd0f0fea9bb655ae8/build_frontend.py#L68-L77

where `shutil.which` in python 3.12 finds a bash script for `npm` rather than the cmd string `npm.cmd`.  (Windows doesn't know about bash scripts.)

Note that this (I think) only affects the package build process.  In shouldn't affect users who install from PyPI.

I'm not sure what the best fix is.  I guess wait to see what happens on python/cpython#109590 before worrying too much.

dairiki:
> See python/cpython#109590 for an explanation of the failure under python 3.12 on Windows.

So python/cpython#109590 was fixed in python/cpython#109995, however, that fix did not make it into python 3.12.0 :frowning_face: (It is, I think, scheduled for inclusion in 3.12.1.)

So here I've made things work by changing [build_frontend.py](https://github.com/lektor/lektor/blob/master/build_frontend.py) to run *npm* using the `shell=True` argument to `subprocess.run` (instead of using `shutil.which` to resolve an absolute path to `npm`, which is broken on Windows in python 3.12.0 by python/cpython#109590).

The first workaround I tried, running *npm* without an absolute path, e.g. `subprocess.run(("npm", "-v"))` does not seem to work on Windows (⇒ `FileNotFoundError` for undetermined reasons).

PR #1184, opened 2024-05-14
dash blueprint: quote add-child endpoint

mjoerg:
The add-child endpoint must be quoted, otherwise parsing fails and the endpoint will not be accessible.
For werkzeug < 3.0.2, parsing fails silently, for later versions, it fails with an exception.


dairiki:
@mjoerg Thank you!

dairiki:
This has now been published to PyPI in [Lektor 3.4.0b12](https://pypi.org/project/Lektor/3.4.0b12/)

PR #1189, opened 2024-06-22
frontend: npm update and major bumps of dev dependencies

yagebu:
### Issue(s) Resolved

Includes fixes for vulnerabilities reported by npm audit (and
dependabot).

### Description of Changes


dairiki:
LGTM.  Thank you!

dairiki:
@yagebu Do we need to worry about updating the npm deps for the 3.3-branch?

I haven't looked to closely but the vulnerabilities that dependabot is worried about seem more related to XSS or DOS issues from processing untrusted input.  Since we only use JS for code intended to be run locally by the site maintainer, we don't have great concerns about untrusted inputs. 

So I think leaving the old deps in the 3.3-branch is probably okay — but, again, I haven't looked or thought about it in great detail.

yagebu:
> Do we need to worry about updating the npm deps for the 3.3-branch?

We don't need to, no. These vulnerabilities are only in dev dependencies of Lektor so these couldn't affect site maintainers at all and they also don't affect us since the Lektor source code isn't untrusted input.

PR #1192, opened 2024-08-13
(CI) Disuse codecov.io

dairiki:
We have been using codecov.io to aggregate and report test coverage data.

In #1191, I've broken the reporting with the upgrade to the latest version of the codecov action, which [now requires an API token](https://github.com/codecov/codecov-action#:~:text=Tokenless%20uploading%20is%20unsupported) to be explicitly passed to the action.  (I do not have sufficient access to the github project to set a new project secret, so I can not fix that on my own.)

Anyway, I'm not sure that anyone is using this coverage reporting.

This PR just removes it.  In its place, it adds a coverage reporting job in CI that combines the coverage from across the pytest matrix.

### Issue(s) Resolved

Coverage reporting to codecov.io is current broken.


dairiki:
Currently deferring this PR in favor of #1193

webknjaz:
I suppose this can be closed now.

dairiki:
> I suppose this can be closed now.

As far as I can tell, our codecov.io reporting is currently not working.
I think this is due to having the incorrect codecov token configured in our GitHub project secrets.  I don't think that I have sufficient privs to view or manipulate either the token at codecov or the secrets at GitHub, so there's not much I can do to fix it.

Personally, my preferred solution would be to just get rid of the codecov.io reporting.  Hence this PR lingers.

Screenshot taken just now:

image

webknjaz:
> I don't think that I have sufficient privs to view or manipulate either the token at codecov or the secrets at GitHub, so there's

Try logging into Codecov via GitHub and seeing if something's disabled. Check if it's deactivated or something like that. I think any org member should be able to do that and see the token.

Personally, I've started putting the repo-scoped token into the config file, in the open, instead of using the repo secrets. It's not that important to keep hidden. And the upside is that it improves stability for PRs from forks (those don't have access to the secrets context).

I see the PR uploads are still happening: https://app.codecov.io/gh/lektor/lektor/pulls. There's no checks reported to PRs, likely because https://github.com/apps/codecov needs to be installed into the repo for that.

Though, yesterday they announced having a toggle to allow uploads w/o needing any tokens. This is targeted at public projects. I think I saw it in the org settings on Codecov.

webknjaz:
> I think this is due to having the incorrect codecov token configured in our GitHub project secrets.

@dairiki yes, I was able to confirm that this is the case. This is the workflow run for the latest commit on master: https://github.com/lektor/lektor/actions/runs/11827808111/job/32956672703#step:10:49.

> ```console
> info - 2024-11-13 23:48:28,314 -- Your upload is now processing. When finished, results will be available at: https://app.codecov.io/github/nixjdm/lektor/commit/60a54627c472039923597f3dc552b2ab747b6b64
> ```

So the token used belongs to the `nixjdm/lektor` project, not to `lektor/lektor`.

If you follow my advice of hardcoding the repo-scoped token (e.g. https://github.com/aio-libs/propcache/blob/d6980a7/.codecov.yml#L10), you should be able to restore the functionality w/o needing other people.

dairiki:
> If you follow my advice of hardcoding the repo-scoped token

Again, I don't think that I have access to that token. Maybe @nixjdm does? (But I'm not sure.)

webknjaz:
@dairiki I was under impression that being in the org is enough to access it. Have you checked?

dairiki:
> @dairiki I was under impression that being in the org is enough to access it. Have you checked?

When I log in to codecov and browse to https://app.codecov.io/gh/lektor/lektor, it says that I'm "viewing as a visitor".  I'm pretty sure that I cannot access the appropriate tokens on my own.

image

yagebu:
Since this has been broken for more than a year, none of the active maintainers seem permissions to fix this, I'd be for removing it and merging this PR. Re-adding in a working state should be a trivial revert (or switch to whatever action or tool is needed then to submit it) after having added a working token.

dairiki:
> Since this has been broken for more than a year

To be completely accurate, I think it's only partially-broken.  Coverage data for PRs seems to still update, but not the coverage data for the main branch.  I have no idea why this is.

But I agree, if none of the active maintainers can fix it, and none of us use it, we might as well get rid of it.

I would like to add a step to the test workflow that aggregates the coverage data from the python test jobs to produce a single coverage report. I will work on it.

PR #1193, opened 2024-08-13
CI: Provide explicit API token to codecov-action

dairiki:
### Issue(s) Resolved

In #1191 the codecov action was upgrade to the latest version.  This broken codecov reporting because the latest version of the action now requires an explicit API token.

This is an alternative PR to #1192, which just removes the codecov.io reporting altogether.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,6 +149,8 @@ jobs:
           TOX_SKIP_ENV: ${{ matrix.tox_skip_env }}
       - name: Publish coverage data to codecov
         uses: codecov/codecov-action@v4
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
 
   ############################################################################
   # Package and (Possibly) Publish to PyPI


dairiki:
(There seems to be no good way to test this without merging it.  The codecov action uses "tokenless uploading" when running from a fork, but, apparently, insists on a token when running from the main repo.  Weird.)

.github/workflows/ci.yml
```diff
@@ -149,6 +149,8 @@ jobs:
           TOX_SKIP_ENV: ${{ matrix.tox_skip_env }}
       - name: Publish coverage data to codecov
         uses: codecov/codecov-action@v4
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
webknjaz, line 153: So this may still be flaky, in my experience. In many projects, I've migrated to using a token hard-coded in the `.codecov.yml` config, in the open. It's not a very sensitive value, after all. But it does make uploads from PRs much more stable.

dairiki: Well, we're stuck at the same problem.  I don't think that I have sufficient privs at codecov.io to update/access the appropriate token.  (Once we get the right one, I would assume the GitHub secrets would work as designed to propagate that value?)

Anyway, any current issues I think are due to not having the correct token.

(Getting this working is not a high priority, personally.)

PR #1196, opened 2024-10-12
Add support for webp and avif images and add a global thumbnail format config 

### LINKED ISSUE #1190: Add native support for avif and webp formats?
Hello team! I a have been working on a new website recently with lektor and noticed there is no native support for these image formats.

I used to contribute to Lektor in the past so I would be happy to help again to implement this. I am a bit rusty so any pointers in the codebase would help

I think we could use these packages:
- https://pypi.org/project/webp/ but https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#webp already seem to support it
- https://pypi.org/project/pillow-avif-plugin/

Cheers!

goanpeca:
### Issue(s) Resolved

Fixes #1190

### Description of Changes

- Added support for webp and avif formats as native lektor formats.
- Also added a configuration parameter to the project section (`thumbnail_format` to globally override the default thumbnail format to use everywhere. It takes values without the `.` extension so `jpeg` and not `.jpeg` this needs probably a better name or location within the config, I just needed a global switch, hence the current implementation.
- Also added a `format` parameter to the `thumbnail` filter/func if this needs to be handled individually.

I still need to write the tests, but wanted to provide something in the meantime as I will be using this branch directly to build some projects :)


goanpeca:
Question: There are two files named https://github.com/lektor/lektor/blob/master/tests/demo-project/content/icc-profile-test/ `LICENSE.txt` and `license.txt`, which is causing issues as mac is not case sensitive. I guess windows users will also have issues.

I renamed one of the license, but what would be a more streamlined solution?

Thanks!

etoimene:
What would be the easiest way for me to test this?
Also, how can I revert back in case something goes wrong?

I have latest lektor installed. Sorry for newbie questions. 

Looks like I have lektor at ~/.local/lib/python3.10/site-packages/lektor$

Can I just unzip this on top of it?

lektor/db.py
```diff
@@ -809,8 +809,17 @@ def format(self):
...
 
-    def thumbnail(self, width=None, height=None, mode=None, upscale=None, quality=None):
+    def thumbnail(
+        self,
+        width=None,
+        height=None,
+        mode=None,
+        upscale=None,
+        quality=None,
+        format=None,
```
dairiki, line 819: I'd suggest that the new `format` parameter be made keyword-only.

(We should give some consideration to making `mode`, `upscale`, and `quality` keyword-only as well.  This would be a breaking change, but I think the likelihood of breaking actual user code is not too great.)

lektor/imagetools/image_info.py
```diff
@@ -20,6 +20,7 @@
 from xml.etree import ElementTree as etree
 
 import PIL.Image
+import pillow_avif  # noqa: F401  # pylint: disable=unused-import
```
dairiki, line 23: How much confidence do we have that [pillow-avif-plugin](https://github.com/fdintino/pillow-avif-plugin/) is ready for prime time?  (E.g. is it compatible with a suitably wide variety of Pillow versions?)

Should we just wait on AVIF support until python-pillow/Pillow#5201?

Should we put this import into a try-except block to make installation optional?

lektor/imagetools/thumbnail.py
```diff
@@ -414,6 +426,7 @@ def make_image_thumbnail(
     mode: ThumbnailMode = ThumbnailMode.DEFAULT,
     upscale: bool | None = None,
     quality: int | None = None,
+    format: str | None = None,
```
dairiki, line 429: As noted above, make new `format` parameter keyword-only.

Here, I think we can change `mode`, `upscale`, and `quality` (and maybe event `width` and `height`) to keyword-only without much fear of breaking user code.

tests/demo-project/content/icc-profile-test/license2.txt
dairiki, line 1: I should have combined LICENSE.txt and license.txt into a single LICENSE.txt file, with a short preamble before each license explaining what it applies to.  It's probably worth doing that now.

In particular the preambles would note (this is also noted in README.md):
- The Apache license in LICENSE.txt applies to the `rgb-to-gbr-test.jpg` test image (from [codelogic/wide-gamut-tests](https://github.com/codelogic/wide-gamut-tests)).
- The CC license in license(2).txt applies to the ICC color profile in `CGATS001Compat-v2-micro.icc` (from [saucecontrol/Compact-ICC-Profiles](https://github.com/saucecontrol/Compact-ICC-Profiles)).)

dairiki (changes requested):
Thank you for working on this! Apologies for the delayed review.

I'm a little leery about adding a dependency on pillow-avif-plugin. I don't know much about it, it doesn't appear to be highly popular, so I worry about whether it might cause issues with some {OS,python,Pillow} version that we support.  Perhaps make it optional?  Or defer AVIF support until Pillow supports it natively?

And, I'd like to add tests before merging this:
- Checks that WEBP (and AVIF) thumbnails can, in fact, be generated.
- Checks that the various ways of specifying thumbnail format work:
   - via explicit argument to the `image.thumbnail` method
   - via configuration in the `.lektorproject` file
   - via default to source image format

A few other nits are noted in place.  Open to discussion on all.

dairiki:
Another issue for discussion: should we add some way to control other parameters of WEBP generation?

E.g. WEBP supports both lossy (think JPEG) and lossless (think PNG) formats.  Do we want to support generating both sorts?

If so, should we (by default), e.g., default to lossless WEBP thumbnails if the source image is lossless (e.g. PNG) (and lossy WEBP thumbnails if the source is lossy, e.g. JPEG)?

Here are the Pillow docs on [supported WEBP save parameters](https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#webp).

etoimene:
All I used from this was ability to select webp images. Works great.
I have to try thumbnail generation.

etoimene:
One issue I noticed is that this version
```
$ lektor --versionfile 
Lektor, version 3.4.0b13.dev10+ga4d67d6
```

produces files with 600 privileges in the build folder.
```
~/.cache/lektor/builds/3d044ef991a7728c2329e15363871edf$ ls -al
total 3228
drwxrwxr-x 13 usernam usernam   4096 Dec  4 10:40  .
drwxrwxr-x 22 usernam usernam   4096 Oct 21 18:17  ..
-rw-------  1 usernam usernam  36508 Dec  4 10:40 'BVCER_SansQR-ISO 14001.png'
-rw-------  1 usernam usernam   6635 Dec  4 10:40  clouds6.jpg
drwxrwxr-x  2 usernam usernam   4096 Dec  4 10:40  contact
drwxrwxr-x 10 usernam usernam   4096 Dec  4 10:40  en
-rw-------  1 usernam usernam   7801 Dec  4 10:40  home-1.jpg
-rw-------  1 usernam usernam  32286 Dec  4 10:40  home-2.png
-rw-------  1 usernam usernam   6476 Dec  4 10:40  home-3.png
-rw-------  1 usernam usernam   4575 Dec  4 10:40  home-4.png
```


Any way to control that?

dairiki:
> One issue I noticed is that this version
> 
> ```
> $ lektor --versionfile 
> Lektor, version 3.4.0b13.dev10+ga4d67d6
> ```
> 
> produces files with 600 privileges in the build folder.

I believe this is a dup of #1213, which was recently reported.

(I think all versions of Lektor currently do this.)

etoimene:
Could be. I had one that I installed with sudo apt install lektor on Linux Mint 21.3 before which made good permissions.
Thanks for this! I'll check the plugin. :)

dairiki:
> Could be. I had one that I installed with sudo apt install lektor on Linux Mint 21.3 before which made good permissions.
> Thanks for this! I'll check the plugin. :)

You're right.  This was introduced in 3.4.0b7.  (See this [comment](https://github.com/lektor/lektor/issues/1213#issuecomment-2518104921) for more.)

PR #1199, opened 2024-11-13
📦 Upload dists to PyPI in a dedicated job

webknjaz:
This is important because it allows migrating to Trusted Publishing safely [[1]]. Which in turn will enable the maintainers to revoke the API token secret making the publishing tokenless and will start uploading digital attestations.

The environment called `pypi` will have to be set up as trusted on the PyPI side and might benefit from having manual approvals within the repository settings.

[1]: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,10 +153,9 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
 
   ############################################################################
-  # Package and (Possibly) Publish to PyPI
+  # Build distribution packages
   ############################################################################
-  deploy:
-    needs: ["lint", "node-tests", "python-tests"]
+  build:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


dairiki:
+1 for moving to PyPI Trusted Publishing.

My one hesitation about touching anything to do with the publishing workflow is that I do not have permissions to administer Lektor on the PyPI side (or the GitHub side, for that matter), making it difficult to fix issues if things go wrong.

Do you have sufficient permissions to make the necessary changes on the PyPI side?

webknjaz:
@dairiki I don't have those permissions, so somebody will have to set it up. Good news, however, is that this is just a preparatory PR. I didn't remove the use of the secret, so it should keep being used. My action only enables Trusted Publishing if the `password` input is empty. So whenever trust is configured within PyPI, a separate PR removing that input will be needed. Deleting the secret from the repo settings should do the same. Once no longer in use, I'd recommend removing the secret anyway and maybe getting somebody to revoke it on PyPI.

.github/workflows/ci.yml
```diff
@@ -177,14 +177,35 @@ jobs:
...
+        uses: actions/download-artifact@v4
+        with:
+          name: python-package-distributions
+          path: dist/
+
+      - name: Publish distribution 📦 to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          password: ${{ secrets.PYPI_PASSWORD }}
+          password: ${{ secrets.PYPI_PASSWORD }}
```
webknjaz: Urgh.. I wonder why my editor dropped the trailing LF here. It doesn't usually break text files like that.

webknjaz:
The test failures are unrelated.

dairiki:
> The test failures are unrelated.

Yes, see #1200.  I'll try to look at that further today or tomorrow.

webknjaz:
https://github.com/lektor/lektor/pull/1201 could be merged before this PR to keep the CI status green

webknjaz:
I reshuffled `needs` since `build` is now just another linting job outside the release context, so it's no longer necessary to block it on running tests or other linters.

webknjaz:
@dairiki meanwhile, you can go to https://github.com/lektor/lektor/settings/environments, click `New environment`, type in `pypi`, check `☑️ Required reviewers` and add yourself (maybe somebody else too, or even a GH org team). Optionally, add a cooldown timer.

This is not mandatory, though. The environment gets auto-created on first use. But no protections are created automatically — this is something I recommend (and included in the PyPUG guide). If you set up the review requirement, GHA will pause before starting the publishing job, send out the notifications to everybody listed and wait until somebody approves it before it gets to execute.

dairiki:
> @dairiki meanwhile, you can go to https://github.com/lektor/lektor/settings/environments, click `New environment`,

Sadly, I do not have sufficient privileges to access the project settings.  (This is one of the reasons I've avoided messing with the publishing workflow as much as possible.)

webknjaz:
Alright, just merge this then, and address it later. It won't change the process.

dairiki:
Thank you!

webknjaz:
Wonderful! I'm going to submit a few more PRs against the CI configs.

dairiki:
> Wonderful! I'm going to submit a few more PRs against the CI configs.

Note, I'm mucking with ci.yml in #1203.  I hope to have that merged soon.

webknjaz:
Next one: #1204

PR #1204, opened 2024-11-14
🧪 Introduce a unified gate/check GHA job

webknjaz:
This adds a GHA job that reliably determines if all the required dependencies have succeeded or not.

It also allows to reduce the list of required branch protection CI statuses to just one — `check`. This reduces the maintenance burden by a lot and have been battle-tested across a small bunch of projects in its action form and in-house implementations of other people.

I was curious about the spread of use. And I've just discovered that it is now in use in aiohttp (and other aio-libs projects), CherryPy, some of the Ansible repositories, all of the jaraco's projects (like `setuptools`, `importlib_metadata`), some PyCQA and pytest projects, a few AWS Labs projects (to my surprise), CPython and many others. Admitedly, I maintain some of these but it does address some of the real pain folks have:
https://github.com/jaraco/skeleton/pull/55#issuecomment-1106638475.

The story behind this is explained in more detail at https://github.com/marketplace/actions/alls-green#why.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -170,17 +170,40 @@ jobs:
           path: dist/
 
   ############################################################################
-  # (Possibly) Publish to PyPI
+  # Branch protection gate
   ############################################################################
-  publish-to-pypi:
-    name: >-
-      Publish Python 🐍 distribution 📦 to PyPI
-    if: github.ref_type == 'tag' # only publish to PyPI on tag pushes
+  check: # This job does nothing and is only used for the branch protection
+    name: All tests passed
+    if: always()
+
     needs:
       - build
       - lint
       - node-tests
       - python-tests
+
+    runs-on: ubuntu-latest
+
+    timeout-minutes: 1
+
+    steps:
+      - name: Decide whether the needed jobs succeeded or failed
+        uses: re-actors/alls-green@release/v1
+        with:
+          jobs: ${{ toJSON(needs) }}
+
+  ############################################################################
+  # (Possibly) Publish to PyPI
+  ############################################################################
+  publish-to-pypi:
+    name: >-
+      Publish Python 🐍 distribution 📦 to PyPI
+    if: >- # only publish to PyPI on tag pushes
+      always()
+      && github.ref_type == 'tag'
+      && needs.check.result == 'success'
+    needs:
+      - check
     runs-on: ubuntu-latest
     environment:
       name: pypi


webknjaz:
@dairiki so eventually, with this in, it should be possible for somebody with the right amount of access to set up a branch protection ruleset, making `All tests passed` a required check. This in turn will make it possible for the project to enable fun features like auto-merge or merge queues. I think it would be useful to file a number of issues post-merge describing tasks to be performed by the people with better accesses, separately.

dairiki:
> @dairiki so eventually, with this in, it should be possible for somebody with the right amount of access to set up a branch protection ruleset, making `All tests passed` a required check. This in turn will make it possible for the project to enable fun features like auto-merge or merge queues. I think it would be useful to file a number of issues post-merge describing tasks to be performed by the people with better accesses, separately.

I don't have much familiarity with auto-merge or merge queues. They do sound potentially useful.
At the same time, this is not a large project (in terms of the number of active contributors and maintainers) we seem to be getting along okay without them.

For the past two years, the bulk of the work — both in terms of maintenance and new features — on Lektor has been done by two people: @yagebu, who among other things has done a great job keeping the JS frontend healthy, and myself.
I don't have sufficient permissions to access the GitHub project settings.  (I'm not sure about @yagebu.)  This makes me hesitant to play with any features that require access to the project settings.
We've been blocked before by issues requiring updates to GitHub (and PyPI) settings.  These have required emails and significant turnaround time to resolve.

Anyhow, I'm hesitant to move forward with this.  If a maintainer who does have access to project settings is interested in shepherding this, that would be a different case.

webknjaz:
@dairiki it is safe to merge as is. It is still a nice restructuring. W/o the repo settings updates, the workflow remains the same. However, adding jobs that depend on "everything important is green" is easier this way. All those other things I mentioned are possible improvements of the configuration on top, nothing more.

webknjaz:
@dairiki perhaps, #1207 would be less controversial?

dairiki:
> However, adding jobs that depend on "everything important is green" is easier this way.

But we only have one job that depends on "all green".  If I understand correctly, this doesn't buy us much until we add more of those.
And it has the downside of introducing one more third-party action into our workflow (which comes with security implications — however slim they are, in this case).


> @dairiki perhaps, #1207 would be less controversial?

Yes, I'm fine with the renaming.

webknjaz:
That's fair. Perhaps its current value is encoding what's considered required in PRs (on the summary workflow run page, in the job graph), as code.

webknjaz:
@dairiki on the point of more jobs, I could bring-in something provenance-related from my other workflows (like SLSA, GH-native attestations and regular Sigstore) if that's desired. Until somebody with access configures Trusted Publishing, the PyPI ones won't be available.

PR #1207, opened 2024-11-15
🧪 Rename the CI/CD workflow to reflect contents

webknjaz:
This patch renames the main GHA workflow definition file to mention the CD bit (since it includes publishing to PyPI). It also changes its name so that it does not take up a lot of space in the UI and is not misleading in the same way. The job names remain the same, meaning that well-known checks/statuses still appear in the PR UI.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -1,4 +1,5 @@
-name: Tests master
+---
+name: 🧪 CI/CD
 
 on:
   # This avoids having duplicate builds for a pull request
--- a/README.md
+++ b/README.md
@@ -1,11 +1,14 @@
 # Lektor
 
-[![Tests master](https://github.com/lektor/lektor/workflows/Tests%20master/badge.svg)](https://github.com/lektor/lektor/actions?query=workflow%3A%22Tests+master%22)
+[![CI/CD Badge]][CI/CD]
 [![Code Coverage](https://codecov.io/gh/lektor/lektor/branch/master/graph/badge.svg)](https://codecov.io/gh/lektor/lektor)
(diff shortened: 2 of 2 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -1,4 +1,5 @@
-name: Tests master
+---
+name: 🧪
```
webknjaz: This could also be
```suggestion
name: 🧪 CI/CD
```
which is short and accurate.

Or
```suggestion
name: CI/CD
```
if you're feeling conservative regarding icons in the UI.

dairiki: I feel conservative (code- and doc-wise, at least).  (I'm old.)

I'd prefer to see the written "CI/CD" in the description so that I don't have to guess as to what the icon means.
Personally, I'd omit the icon altogether — it doesn't add value for me (again, I'm old) — though I don't feel strongly about that.

webknjaz: Let's try having both the icon and text, then. Here's an example of what it looks like when jobs are also labeled similarly: https://github.com/ansible/awx-plugins/actions/runs/11707588320.

webknjaz:
@dairiki I've updated the name, going for the suggested middle ground.

.github/workflows/ci-cd.yml
Firas2515, line 1: https://github.com/solana-labs/security-audits/blob/master/spl/NeodymeStakePoolAudit-2023-11-14.pdf

Firas2515, line 1: https://github.com/solana-labs/security-audits/blob/master/spl/NeodymeStakePoolAudit-2023-11-14.pdf

PR #1208, opened 2024-11-16
Update to pylint 3.3

dairiki:
This updates the version of pylint we use to 3.3.1 (the latest).

### Issue(s) Resolved

Pylint 3.2.x does not support running under python 3.13 (see, e.g., pylint-dev/pylint#10000).

### Related Issues / Links

Pylint 3.3 introduces a new check [too-many-positional-arguments] that looks for functions with (drum roll) too many (by default, more than five) positional arguments.

This does seem to be a valid stylistic concern.

Currently, to address this new warning, this PR does two things:
- Increases `max-positional-arguments` to 7 (from the default of 5).
- For a number of existing functions that currently accept large numbers of positional arguments, mark excess arguments as keyword-only.
 
Most of the functions whose signatures are modified are internal-use only.  None are documented API.  I think the chance of breaking user code is minimal.

An alternative to making these changes would be to simply ignore the new check. 

[too-many-positional-arguments]: https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/too-many-positional-arguments.html

### DIFF
--- a/lektor/builder.py
+++ b/lektor/builder.py
@@ -666,6 +666,7 @@ class Artifact:
         artifact_name,
         dst_filename,
         sources,
+        *,
         source_obj=None,
         extra=None,
         config_hash=None,
--- a/lektor/cli.py
+++ b/lektor/cli.py
@@ -92,6 +92,7 @@ def cli(ctx, project=None, language=None):
 @pass_context
 def build_cmd(
     ctx,
+    *,
     output_path,
     watch,
     prune,
@@ -175,7 +176,7 @@ def build_cmd(
 @click.confirmation_option(help="Confirms the cleaning.")
 @extraflag
 @pass_context
-def clean_cmd(ctx, output_path, verbosity, extra_flags):
+def clean_cmd(ctx, *, output_path, verbosity, extra_flags):
     """Cleans the entire build folder.
 
     If not build folder is provided, the default build folder of the project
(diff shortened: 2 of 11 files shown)


webknjaz:
Are you not going to drop it in favor of #1209?

dairiki:
> Are you not going to drop it in favor of #1209?

That would be my personal preference.  I have made motions to drop pylint before (since we also have flake8 in our pre-commit hooks), but there was resistance to that.  (From @yagebu, IIRC, among perhaps others.)

Pylint certainly does catch some things that flake8 does not.

Ruff includes a number of pylint's rules, but the overlap is not 100%. (See astral-sh/ruff#970.)
I don't really know the details, but pylint does some deep AST inspection (hence the slowness, and the need to have all dependencies in the environment) so I doubt the overlap will ever be complete.


----

In any case, this PR fixes things so that pylint passes when run under python 3.13.  Until such time as we decide to jettison pylint, that's a good thing.  (The only reason our CI passes right now is that the actions/setup-python, with python-version=3.x is still installing 3.12.x for whatever reason.)

webknjaz:
Yeah, I normally keep both pylint+flake8. pylint also performs some semi-dynamic checks, not just scanning AST but also importing modules IIRC.
I wasn't asking because I don't like it, but mostly because it seemed like Ruff might be covering this. Although, I'm yet to look into Ruff deeper. I know that Ruff has limitations in terms of allowing people to have own plugins and if that's the case, I'd still use it for project-local plugins where they exist.

dairiki:
I'm, too, am unsure of the differences, but the fact that pylint requires all dependencies to be installed means it must do something different.
As a trivial example, the following code passes flake8 and `ruff check --select ALL`, but pylint will fail with an import-error:

```py
"""Test."""

import missing_module

missing_module.foo()
```

It's unfortunate that pylint is so slow to run.  For me, that makes it unsuitable for pre-commit usage.

PR #1210, opened 2024-11-18
🧪💅 Unify running linters via pre-commit

webknjaz:
Previously, pylint was specialcased. However, there are a lot of advantages to having a unified invocation. For example, a single tox env run that can be integrated into many places. Or, enabling pre-commit.ci eventually.

The only downside is having to keep track of a copy of the direct runtime dependencies in another config.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,3 +40,34 @@ repos:
         entry: ./frontend/node_modules/prettier/bin/prettier.cjs --write --list-different
         require_serial: true
         files: \.(ts|tsx|js|less|ya?ml|md)$
+
+  - repo: https://github.com/PyCQA/pylint.git
+    # rev: v3.3.1
+    rev: v3.2.7
+    hooks:
+      - id: pylint
+        additional_dependencies:
+          - hatchling # distribution package build backend
+          - Babel # runtime
+          - click >= 6.0 # runtime
+          - Flask # runtime
+          - inifile >= 0.4.1 # runtime
+          - Jinja2 >= 3.0 # runtime
+          - MarkupSafe # runtime
+          - marshmallow # runtime
+          - marshmallow_dataclass >= 8.5.9 # runtime
+          - mistune >=0.7.0, <3 # runtime
+          - Pillow >= 6 # runtime
+          - pip # runtime
+          - python-slugify # runtime
+          - requests # runtime
+          - typing-extensions >= 3.10; python_version < '3.10' # runtime
+          - tzdata; sys_platform == 'win32' # runtime
+          - watchfiles >= 0.12 # runtime
+          - Werkzeug >= 2.1.0 # runtime
+          - ipython # runtime, additional through `ipython` extra
+          - traitlets # runtime, additional through `ipython` extra
+          - pytest >= 6 # tests
+          - pytest-mock # tests
+          - importlib_metadata; python_version < "3.10" # tests
+          - iniconfig # tests
--- a/tox.ini
+++ b/tox.ini
@@ -1,12 +1,17 @@
 [tox]
 minversion = 4.1
 envlist =
-    lint
+    pre-commit
     {py39,py310,py311,py312,py313}{,-mistune0}
     py311-noutils
     py311-tzdata
     cover-{clean,report}
 isolated_build = true
+
+[python-cli-options]
+byte-errors = -bb
+max-isolation = -E -s -I
+warnings-to-errors = -Werror
 
 [gh-actions]
 python =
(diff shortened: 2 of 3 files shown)


webknjaz:
@dairiki this might seem big, but it enables future unification of the DX+CI experience on top. It should eventually be possible to have a reusable workflow that would run different tox envs. Additionally, when someone is able to install pre-commit.ci, it would be picked us there too (which is the only way to run pylint in that env, really).

.pre-commit-config.yaml
```diff
@@ -40,3 +40,34 @@ repos:
         entry: ./frontend/node_modules/prettier/bin/prettier.cjs --write --list-different
         require_serial: true
         files: \.(ts|tsx|js|less|ya?ml|md)$
+
```
dairiki, line 43: While I appreciate the purity of having all the lint checks as pre-commit hooks, I am unconvinced this a good move.

1. I really don't like the idea of having to maintain an independent copy of all our dependencies here.
2. Pylint is really pretty slow and so painful to run as a pre-commit hook.  On my dev machine running pylint on just one source file (admittedly one of our larger ones) `pylint lektor/db.py` takes 3.5 seconds.  (A full run, `pylint lektor tests`, takes 21 seconds.)

tox.ini
```diff
@@ -64,17 +69,128 @@ allowlist_externals = env
...
+    {[python-cli-options]max-isolation} \
+    {[python-cli-options]warnings-to-errors} \
+    -c \
+      'cmd = "{envpython} -m pre_commit install"; \
+      scr_width = len(cmd) + 10; \
+      sep = "=" * scr_width; \
+      cmd_str = "    $ \{cmd\}";' \
+      'print(f"\n\{sep\}\nTo install pre-commit hooks into the Git repo, run:\
+      \n\n\{cmd_str\}\n\n\{sep\}\n")'
+commands_post =
```
dairiki, line 103: What does all this in `commands_post` do?
(As best I can tell, it doesn't do anything for us?  Maybe its copypasta, or preparation for something more?)

tox.ini
```diff
@@ -64,17 +69,128 @@ allowlist_externals = env
...
-    importlib_metadata; python_version<"3.10"
-    iniconfig
+[testenv:pre-commit]
+description =
+  Run the quality checks under {basepython}; run as
+  `SKIP=check-id1,check-id2 tox r -e pre-commit` to instruct the underlying
+  `pre-commit` invocation avoid running said checks; Use
+  `tox r -e pre-commit -- check-id1 --all-files` to select checks matching IDs
+  aliases{:} `tox r -e pre-commit -- mypy --all-files` will run 3 MyPy
+  invocations, but `tox r -e pre-commit -- mypy-py313 --all-files` runs one.
```
dairiki, line 79: We don't have mypy configured to run anywhere (yet), right?

tox.ini
```diff
@@ -64,17 +69,128 @@ allowlist_externals = env
...
 commands =
-    pylint {posargs:lektor tests}
+  {envpython} \
+    {[python-cli-options]byte-errors} \
+    {[python-cli-options]max-isolation} \
+    {[python-cli-options]warnings-to-errors} \
+    -m pre_commit \
```
dairiki, line 85: What does running python with all the paranoia options (-bb -E -s -i  -Werror -mI) buy us when running pre-commit?  Is it worth the verbosity?

dairiki (commented):
I'm not adverse to adding a `pre-commit` testenv to tox.ini.

For me, the best place to unify all the linters is in tox.ini.  (With pre-commit hooks still configured in .pre-commit-config.yaml, but so that — for DX — a simple `tox r` will run all tests.)

In my experience, neither pylint nor mypy work very well as pre-commit hooks. They rely on the installation of project dependencies, and they are slow to run, leading to painful delays during git commits.

(An alternative to configuring all the linters via tox could be to move to something like PDM, where one can use PDM scripts (configured in pyproject.toml) to run the linters via, e.g. `pdm run lint`.  (Perhaps `uv` could be another alternative to using `PDM` for this.  I haven't dug into `uv` enough yet to say.)

Yikes, though. This patch looks awfully verbose to me.

dairiki:
While I like this idea in the abstract, I really don't like the aspect in this PR of having to keep two lists of dependencies — one in pyproject.toml, one in .pre-commit-config.yaml — up-to-date.

I'm going to close this for now.

Perhaps running pylint as a local pre-commit hook via `uv run` would be a better option, though I haven't investigated that very thoroughly. (See my comment at https://github.com/lektor/lektor/pull/1252#issuecomment-3916598086)

PR #1214, opened 2024-12-04
Fix file mode of created files

### LINKED ISSUE #1213: File mask for build artifacts
```
> ls -l build/*
-rw------- 1 xyz xyz 20451 Nov 26 00:08 build/index.html

build/de:
total 24
-rw------- 1 xyz xyz 20518 Nov 26 00:08 index.html

build/static:
total 16
drwxr-xr-x  2 xyz xyz 4096 Nov 26 00:03 css/
drwxr-xr-x  2 xyz xyz 4096 Nov 26 00:03 img/
drwxr-xr-x  2 xyz xyz 4096 Nov 26 00:03 js/
```

Looks like currently directories are created with the convenient `drwxr-xr-x` permissions, files however are assigned `read`/`write` only to the user `-rw-------`. In my case it would be much more convenient, if file group would also have the same or just the read permission: `-rw-r-----`.

I've managed to achieve the desired effect with the following plugin:
```
from lektor.pluginsystem import Plugin
from pathlib import Path


class UmaskPlugin(Plugin):
    name = 'umask'
    description = "640 permissions on output files."

    def on_after_build(self, prog, **extra):
        filemode = "640"
        artifact = prog.primary_artifact
        if not artifact:
            return
        path = Path(artifact.dst_filename)
        if path.is_file():
            path.chmod(int(filemode, 8))
```

But despite my great success, I guess ideally there should be `umask` setting within the `Project File`, so that users may choose whatever works best for them? :thinking:

dairiki:
### Issue(s) Resolved

Fixes #1213

This PR also fixes a related buglet having to do with the mode of files created/edited using `lektor.utils.atomic_open`.
`Atomic_open` is used when editing `contents.lr` files via the admin GUI.

### Description of Changes

This fixes things so that artifacts are created with the same permissions as any normally created file would have.
I.e. newly created files will now have mode `0o666 & ~umask` — previously, artifacts either got mode 0o644 (prior to 3.4.0b7) or 0o600 (since then).

### DIFF
--- a/lektor/builder.py
+++ b/lektor/builder.py
@@ -6,7 +6,6 @@ import shutil
 import sqlite3
 import stat
 import sys
-import tempfile
 from collections import deque
 from collections import namedtuple
 from contextlib import contextmanager
@@ -23,6 +22,7 @@ from lektor.constants import PRIMARY_ALT
 from lektor.context import Context
 from lektor.reporter import reporter
 from lektor.sourcesearch import find_files
+from lektor.utils import create_temp
 from lektor.utils import fs_enc
 from lektor.utils import process_extra_flags
 from lektor.utils import prune_file_and_folder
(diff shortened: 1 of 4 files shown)


dairiki:
### Open question

We may want to modify `atomic_open` so that the mode of an existing file (if there is one) is preserved when it is edited.

PR #1222, opened 2025-04-18
feat: Support marshmallow 4

### LINKED ISSUE #1221: Lektor tests failing with latest version of marshmallow (4.0.0)
Breaking changes that break lektor are:
- context gone
- validators must raise a validation error on validation failure now, not return False.

do3cc:
Marshmallow 4 contained backward breaking changes. This commit fixes the broken parts.
All fixes also work with previous marshmallow versions

### Issue(s) Resolved

Fixes #1221 

### Description of Changes

- All custom marshmallow validators now raise ValidationErrors on failure, instead of returning False
- No more usage of context in schemas, instead using ContextVar

### DIFF
--- a/lektor/admin/modules/api.py
+++ b/lektor/admin/modules/api.py
@@ -2,6 +2,7 @@ import os
 import posixpath
 from collections.abc import Iterator
 from collections.abc import Mapping
+from contextvars import ContextVar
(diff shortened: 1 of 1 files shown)


yagebu (approved):
Thanks for the fix, LGTM :)

PR #1223, opened 2025-04-20
fix(builder): stop treating @ as virtual by default

### LINKED ISSUE #1183: Always rebuilding assets with @ in name
I have javascript assets with an `@` in their name, e.g., <https://www.npmjs.com/package/@fontsource/fira-sans>.

Lektor appears to "rebuild" (i.e., copies) these assets every time, making it unnecessary slow.

(But its not the only thing that makes it slow for large projects, I already have build times of 26+ seconds, unfortunately., c.f. #180 )

do3cc:
### Issue(s) Resolved

Fixes #1183

### Description of Changes

The bug report was about filenames starting with @ getting rebuilt every time. Upon inspection, the path is treated like a virtual path. I've changed the code so that the virtual information becomes part of the database, thus the @ char in the path does not automatically mean something special.

I did not find any migration scripts, so I did it directly in the builder, dropping old tables.
This will force a full rebuild upon the first start.

I do not feel 100% confident with the change. The tests are passing, but I am not sure I can always assume that dependencies are non virtual.


dairiki:
@do3cc  First: apologies for letting this lie so long.
Second: I just merged master into this PR, then fixed one test that it broke.  (I hope you don't mind.)

I initially had some concerns that this would conflict with our db-path notation.  E.g. `pad.get("@foo")` refers to the virtual source `@foo` before the (non-virtual) source at `/`.
But upon reflection, I think I'm convinced this only affects the builder and the management of the `artifacts` table.  As long as we only allow `@` in filenames of assets (and continue to disallow them in page/source names), I think it's okay.  (As things stand, I think `pad.get(path)` is never going to return a file-based source object if `path` contains an `@` — except maybe in the case of pagination.)

While refactoring the `artifacts` table schema, perhaps it would be a good time to normalize it a bit more.
Fully describing an artifacts' sources involves specifying not only the path to the source file, but also, for virtual sources, the virtual path and the `alt`.  Currently, there are hacks like [_pack_virtual_source_path](https://github.com/lektor/lektor/blob/7522911c482efbbe8024e6a2f30b8785b5fa13f0/lektor/builder.py#L598) used to store the (path, vpath, valt) triple all packed into the single `artifacts.path` column.  It might be good to split those into their own columns in the sql database.

On the other hand, the more I think about it, the more my head hurts.  That sort of refactor will probably take some effort.

For now, I'm inclined to merge this as is.

I'll leave it for a few days in case anyone has comments.

yagebu:
> While refactoring the artifacts table schema, perhaps it would be a good time to normalize it a bit more.
Fully describing an artifacts' sources involves specifying not only the path to the source file, but also, for virtual sources, the virtual path and the alt. Currently, there are hacks like [_pack_virtual_source_path](https://github.com/lektor/lektor/blob/7522911c482efbbe8024e6a2f30b8785b5fa13f0/lektor/builder.py#L598) used to store the (path, vpath, valt) triple all packed into the single artifacts.path column. It might be good to split those into their own columns in the sql database.

That's what I also thought when last taking a look at this PR. But that's maybe a nice-to-have refactor and this PR fixes an issue (and doesn't block future refactors to clean this up even further).

> I did not find any migration scripts, so I did it directly in the builder, dropping old tables.
This will force a full rebuild upon the first start.

That approach doesn't seem like it scales nicely for further migrations - but that's something we can also just tackle when we do further migrations. (note to future self or migration author: sqlite's `PRAGMA user_version` seems like a decent fit for storing a Lektor schema version - doing a drop all and recreating if not matching).

yagebu (approved):

dairiki:
> `PRAGMA user_version` seems like a decent fit for storing a Lektor schema version - doing a drop all and recreating if not matching

+1  I did not know about `PRAGMA user_version`.  It's sqlite only, but I don't think that's an issue for us — we're pretty committed to sqlite.

Perhaps we should incorporate that into this PR (so that we don't have to rely on a comment on a closed PR when it comes up next time).
I will try to work on it.  ("I think it's straightforward...", he said, naïvely.)

dairiki:
> I will try to work on it. ("I think it's straightforward...", he said, naïvely.)

Having started to work on that, I'm going to merge this PR, and start that in a separate "cleanup the DB code" PR.

Merging.  @do3cc Thank you for the PR!

Issue #358, opened 2017-02-20
Neither site.query() nor site.get().children.include_hidden(True) seem to include hidden records


niijv:
`site.query('/')` should include all child records of `'/'` (as per documentation).
`site.get('/').children.include_hidden(True)` should aswell.

I can't get this to work, in both cases hidden child records are not returned.

You can reproduce it with this single-page-demo I built: https://github.com/niijv/single-page-demo
Just remove the `hidden = false` setting from https://github.com/niijv/single-page-demo/blob/master/models/doc-pages.ini#L7

tariquesani:
Reproducible indeed.

dairiki:
This is a "pre-dup" of #1229.

I'm going to close this one in favor of that one, since I just wrote a somewhat detailed comment over there.

PR #1251, opened 2026-02-12
Fix.issue 1229 include hidden

### LINKED ISSUE #1229: Hidden children inaccessible even with include_hidden
Let's say I have a model called Members, and a child model called Member. I'd like to display members all on the Members page, but not give them separate pages.

```ini
[model]
name = Members

[children]
model = member
hidden = True
```

This does, as advertised, prevent them from building as separate HTML pages, and hides them from `this.children` in the template.

However, I can't seem to get them even with [`include_hidden`](https://www.getlektor.com/docs/api/db/query/include-hidden/) — or via querying, which should include them by default.

That is, *all* of these return nothing for me:

```python
this.children
this.children.include_hidden(True)
site.query("/members")
site.query("/members").include_hidden(True)
```

Am I missing something?

dairiki:
Fixes for `Query.include_hidden`, which currently, unless (`include_undiscoverable` is also set), does not actually include hidden descendants in the query.

### Issue(s) Resolved

Fixes #1229, #358.

#### Record.is_discoverable

Currently, `Record.is_discoverable` is forced to `False` when the record is hidden, even if `_discoverable` is set.
https://github.com/lektor/lektor/blob/7522911c482efbbe8024e6a2f30b8785b5fa13f0/lektor/db.py#L364-L367
I don't see how this makes sense.  (I suspect it is a remnant of initial attempts to support the hidden flag.)
This PR makes the `Record.is_discoverable` logic independent of `Record.is_hidden`.

 
This is a change in semantics, and so could affect current projects.  I'm having trouble imaging how the original semantics might be useful, however.

#### Query._include_hidden tri-state logic

Currently, `Query._include_hidden` is treated as a tri-state (`None`, `False`, or `True`) value, with `None` meaning "not explicitly set".

There is logic in `Query.include_undiscoverable` that sets `include_hidden` to False, if it hadn't been explicitly set, when `include_undiscoverable` is set.

https://github.com/lektor/lektor/blob/7522911c482efbbe8024e6a2f30b8785b5fa13f0/lektor/db.py#L1074-L1083

There was also logic in `Query._matches` that treated Queries with no explicit setting of include_hidden specially.  On careful inspection, this all seems to do nothing.  `Query._matches` treats `_include_hidden = None` the same as `_include_hidden = False`. 

https://github.com/lektor/lektor/blob/7522911c482efbbe8024e6a2f30b8785b5fa13f0/lektor/db.py#L1001-L1005

The PR removes all the tri-state logic around `Query._include_hidden`

### DIFF
--- a/lektor/db.py
+++ b/lektor/db.py
@@ -343,28 +343,29 @@ class Record(DBSourceObject):
         return self._data["_alt"]
 
     @property
-    def is_hidden(self):
-        """Indicates if a record is hidden.  A record is considered hidden
-        if the record itself is hidden or the parent is.
+    def is_hidden(self) -> bool:
+        """Indicates whether a record is hidden.
+
+        Artifacts are not built for hidden objects.  Also, by default, hidden records
+        are not included in `Query` results.
         """
-        if not is_undefined(self._data["_hidden"]):
-            return self._data["_hidden"]
-        return self._is_considered_hidden()
+        hidden = self._data["_hidden"]
+        if not is_undefined(hidden):
+            return hidden
+        return self._is_hidden_by_parent_config()
 
-    def _is_considered_hidden(self):
-        parent = self.parent
-        if parent is None:
-            return False
+    def _is_hidden_by_parent_config(self) -> bool:
+        # Records may be implicitly hidden by their parents' configuration.  The details
+        # depend on record type.
 
-        hidden_children = parent.datamodel.child_config.hidden
-        if hidden_children is not None:
-            return hidden_children
-        return parent.is_hidden
+        # The Page and Attachment subclasses provide concrete implementations for this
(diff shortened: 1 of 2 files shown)


dairiki:
I guess one possible justification for forcing _hidden_ records to be _undiscoverable_ is that any URLs generated for such records won't work (without other customization/hacking), since no artifact is produced for those records.

OTOH, even though the URLs generated by Lektor for such records is not useful, I can see hidden but discoverable records useful for something like site-menu generation for a SPA (single-page app).

If we really care about this, we could emit a warning or error when an attempt is made to generate a URL for a hidden record.  I'm not sure that's worth it.

yagebu (approved):
Thanks, LGTM :)

PR #1252, opened 2026-02-17
Move test dependencies into pyproject.toml (PEP735)

dairiki:
#### Test/dev dependencies declared in pyproject.toml

This PR moves the declaration of test dependencies from `tox.ini` to  `pyproject.toml`'s `dependency-groups` ([PEP735]).

This allows developers to initialize a local dev environment with these same dependencies by running, e.g.
```sh
python3 -m venv .venv
# need pip>=25.1 for --group support
.venv/bin/python -m pip install -U pip
.venv/bin/pip install --group dev -e .
```
(Of course, `uv`, `pdm`, and other project management tools also support PEP735 dependency-groups.)

#### Other tox.ini cleanup

It also unifies the configuration for the `*-noutils` test environments with that for the default test environment.

#### Update development instructions in README

This PR also updates to development howto at the bottom of `README.md` to match changes made.

### Issue(s) Resolved

Resolves #1057

### Related Issues / Links

This partially overlaps with #1218

[PEP735]: https://peps.python.org/pep-0735/

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ For installation instructions head to the official documentation:
 
 ## Want to develop on Lektor?
 
-This gets you started (assuming you have Python, pip, npm, and pre-commit
+This gets you started (assuming you have Python, npm, and pre-commit
 installed):
 
 ```bash
@@ -40,10 +40,13 @@ $ cd lektor
 $ python -m venv _venv
 $ . _venv/bin/activate
 
-# pip>=21.3 is required for PEP 610 support
-$ pip install -U "pip>=21.3"
+# pip>=25.1 is required for PEP 735 support
+$ pip install -U "pip>=25.1"
 
-$ pip install --editable .
+$ pip install --group dev --editable .
+
+# build the frontend javascript (requires npm to be installed)
+$ make build-js
 
 # If you plan on committing:
 $ pre-commit install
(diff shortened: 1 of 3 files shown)


dairiki:
## TODO

We probably should add dependency-groups for installing dev-time checkers, e.g. `ruff`, `mypy`.

Our current mypy config is out-of-date. Anyway, `mypy` is not currently run in CI tests, so that should probably be revisited separately.

I'm not sure whether these should be included into the default `dev` group.  My current thinking is that they should not.
The specifics of which checkers should be installed depend on individual developer preference, e.g. choice of editor/IDE.  (Some IDEs manage the installation of checkers on their own.  For others, requirements will vary depending on whether the editor supports LSP, etc.

### syncing checker versions with pre-commit

There still seems to be no "standardized" solution for how best to keep requirements for checkers (e.g. ruff, pylint versions) in sync between `.pre-commit-config.yaml` and `pyproject.toml`.
Some options:

- Run external program to update pre-commit versions from lock file, e.g. [sync-with-uv](https://github.com/tsvikas/sync-with-uv), [sync-pre-commit-lock](https://github.com/GabDug/sync-pre-commit-lock)
- Configure a local pre-commit hook that [runs the checker using `uv run`](https://github.com/astral-sh/uv-pre-commit/issues/14#issuecomment-3584167511)
It appears that support for this may be added to the `astral_sh/uv-pre-commit` hook at some point (e.g., astral-sh/uv-pre-commit#53).

PR #1253, opened 2026-02-20
Small cleanups/refactors/features in pluginsystem

dairiki:
### Use importlib.metadata.entry_points to find entry points

The API to [importlib.metadata.entry_points] only stabilized in Python 3.10. Because of this, we had rolled-our-own solution to finding entry points that involved `importlib.metadata.distributions`.

Here we switch to using the `metadata.entry_points` API.  Not only does it result in cleaner code, but it also avoids a weird edge-case.

In certain circumstances, e.g. when setuptools installs a plugin in editable mode, one can wind up with both an `.egg-info` directory in the project source directory, as well as a `.dist-info` directory in the venv's `site-packages`, both for the same distribution packages. When this happens, `importlib.metadata.distributions` will return two `Distribution` instances for the same package.  This would confuse Lektor and elicit a `"RuntimeError: Plugin <xxx> is already registered"` exception. (Ref python/importlib_metadata#481, pypa/setuptools#3649.)

### Add properties to the `Plugin` base class to provide sane defaults for `.name` and `.description`

The default name is taken from the plugin subclass name, word-split on camel-case, and with any trailing `Plugin` discarded.  E.g. as plugin class named `MinifyPlugin` will get the default `.name` of `"Minify"`.

The default description is retrieved from the _Summary_ field of the distribution that contains the plugin.  This value is set by the `description` key in the `[project]` section of `pyproject.toml`, or by the `description` parameter to `setup()`, for example.

In any case, the old way of setting explicit `name` and/or `description` attributes on the plugin subclass still works and may be used if the defaults are not suitable.

### Misc

A few typing issues and checks for `None` have been fixed.  (I'm sure there are still plenty to fix.)

### Issue(s) Resolved

Fixes #

### Description of Changes

[importlib.metadata.entry_points]: https://docs.python.org/3.10/library/importlib.metadata.html#entry-points

### DIFF
--- a/lektor/pluginsystem.py
+++ b/lektor/pluginsystem.py
@@ -91,17 +91,14 @@ class Plugin:
         return None
 
     @property
-    def path(self) -> str:
+    def path(self) -> str | None:
         mod = sys.modules[self.__class__.__module__.split(".", maxsplit=1)[0]]
+        if mod.__file__ is None:
+            return None  # pragma: no cover
         path = Path(mod.__file__).resolve().parent
-        package_cache = self.env.project.get_package_cache_path()
-        try:
-            # We could use Path.is_relative_to(), except that's py39+ only
-            path.relative_to(package_cache)
+        if path.is_relative_to(self.env.project.get_package_cache_path()):
             # We're only interested in local, editable packages. This is not one.
             return None
-        except ValueError:
-            pass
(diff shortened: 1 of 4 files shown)


yagebu (approved):
Nice one :)

PR #1254, opened 2026-02-21
build: use npm install --no-save to avoid dirty git source

### LINKED ISSUE #1250: Should we be building the packaged frontend JS from package-lock.json?
Currently, when we build a new distribution (e.g. for release), in order to build the frontend javascript, etc, we do, effectively:

```sh
cd frontend
# Install dependencies
npm install
# Run esbuild to build the stuff in lektor/admin/static/
npm run build
```

The `npm install` installs requirements based on the requirements listed in `frontend/package.json`, which, in general, pins dependencies to a semver-compatible range, but does not pin specific versions.

Should we replace `npm install` with `npm ci`?  This would use the exact version of dependencies, as specified `package-lock.json`, rather than  the latest semver-compatible version.

Using `package-lock.json` in this way, would:
- Make the (Python) package build process more reproducible.  At present, if I build the package tomorrow (from the same git commit) I may get newer versions of npm packages.
- Make explicit what npm packages were used to build the frontend resources included in the (Python) package.  The `package-lock.json` would serve as an inventory of exactly what went into `lektor/admin/static/app.js`, et. al.

----

Here's the code that actually builds the frontend code when building a wheel and/or sdist:

https://github.com/lektor/lektor/blob/8d420ec1c28967758d5c264c2e8308f0a8e8827e/build_frontend.py#L70-L86

yagebu:
### Issue(s) Resolved

Fixes #1250 

### Description of Changes

### DIFF
--- a/build_frontend.py
+++ b/build_frontend.py
@@ -79,8 +79,8 @@ class FrontendBuildHook(BuildHookInterface):
         if not frontend.is_dir():
             app.abort("frontend source is missing. can not build frontend")
 
-        app.display_info("npm install")
-        subprocess.run("npm install", cwd=frontend, shell=True, check=True)
+        app.display_info("npm install --no-save")
+        subprocess.run("npm install --no-save", cwd=frontend, shell=True, check=True)
         app.display_info("npm run build")
         subprocess.run("npm run build", cwd=frontend, shell=True, check=True)
         app.display_success("built frontend static files")
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ build-js: frontend/node_modules
 
 frontend/node_modules: frontend/package-lock.json
 	@echo "---> installing npm dependencies"
-	@cd frontend; npm install
+	@cd frontend; npm install --no-save
 	@touch -m frontend/node_modules
 
 # Run tests on Python files.


build_frontend.py
```diff
@@ -80,7 +80,7 @@ def initialize(self, version, build_data):
             app.abort("frontend source is missing. can not build frontend")
 
         app.display_info("npm install")
```
dairiki:
```suggestion
        app.display_info("npm install --no-save")
```

dairiki:
I wish I understood `npm` better.

E.g. Why does `npm install` sometimes change the peer dependency info in the lock file?  (Is it just that differing versions of `npm` write slightly different lock files?)

Are we losing anything, or stepping into any traps by always using `--no-save`?

If there's a good reason that `npm` wants to update the lock file, maybe we shouldn't ignore it?


Also note that we don't have a pressing issue with this in the 3.4 branch, since, at least in CI, the build hook that runs `npm install` does so after the release version is determined.

----

That said, I'm fine merging this, if you want.

yagebu:
I think trying to find meaning in how npm misbehaves in these cases is just a waste of time... I've even seen the same version of npm produce different lockfiles with npm install and npm update.

Since most of these dependencies are development dependencies anyway, I'll merge this to avoid breaking flows, I think there's negligible impact of any mis-installations anyway.

PR #1291, opened 2026-09-16
Fix HTML escaping in project quickstart templates

### LINKED ISSUE #622: Lektor quickstart does not escape title and name
When running the quick start and entering "foo <bar>" as a title or as a name, the "<bar>" disappears from the final site in some places since it is not escaped before being hardcoded into the site.
Maybe a related question, should this be hardcoded into the template?

xuu33030:
### Issue(s) Resolved

Fixes #622

### Related Issues / Links

Follows the suggested approach in the issue: keep the values hard-coded in the generated template, but escape them for HTML.

### Description of Changes

The quickstart generator renders several file formats without autoescaping. In `layout.html.in`, project and author names were inserted directly into HTML, so names containing angle brackets or ampersands were interpreted as markup/entities.

Apply Jinja's `e` filter to the project name in the title and heading, and to the author name in the footer. Leave project configuration and content data unchanged so they are not double-escaped when rendered later.

- Screenshot/documentation update: not included; this changes escaping rather than layout or public APIs.

The regression test generates and builds real quickstart sites with and without a blog, checks the resulting HTML, and confirms the original values remain intact in project/content data. The angle-bracket project-name case is skipped on Windows because the project name is also used as a filename; the ampersand case and author-name escaping still run there.

Validation on macOS, Python 3.13.5:

- Before the fix: both original regression cases failed on the generated HTML assertion (existing quickstart tests passed).
- `python -m pytest tests/test_quickstart.py tests/test_builder.py tests/test_environment.py tests/test_project.py tests/test_themes.py -q`: **88 passed, 1 xfailed**.
- Ruff 0.15.1 check and format check: passed.
- Pylint 4.0.4 on `tests/test_quickstart.py`: passed.
- `git diff --check`: passed.

The full multi-version/OS test matrix and frontend build/tests were not run locally. The documented `HATCH_BUILD_NO_HOOKS=true` option was used for the Python development install.

AI assistance was used for investigation, implementation, and test execution.


dairiki:
@xuu33030 Thank you!  Merged.

